PR #18, opened 2012-01-04
Support jQuery 1.6+

rrotter:
Changed .attr to .prop where appropriate to support jQuery 1.6+. This
breaks support for any jQuery version prior to 1.6.


nikolas:
So, should this get merged? If so, the PR I've opened for jQuery 1.5.2 here: #41 can be closed.

rrotter:
it's dead jim

PR #21, opened 2012-02-22
RTLing the nav slider

yankl:
The nav slider would go left to right even in
right-to-left books. This makes the UI more
intuitive for RTL books.


rchrd2:
@yankl Thank you for your pull request. I recently took over this project, and as you could imagine, there's a large backlog of issues to address. Though not immediately, I will review this merge it in if it passes our QA.

PR #38, opened 2015-03-08
Bug fix: only initialize BR and startPolling once

garyrh:
When I'm reading a book that I've checked out on OL, I usually minimize the toolbars as they're a bit distracting. Unfortunately, I noticed they kept popping up every 5 minutes along with a message saying that I was disconnected from OL (even though I definitely wasn't.)

So I dug through the source code, found what I believed to be the problem, and wrote [a quick userscript](http://pastebin.com/th48MwHv) for it. The script has worked well for me so far, so I've put the changes into BookReaderJSIA.php.


garyrh:
I've reset the fix to something that wasn't just a [bandage](http://web.archive.org/web/20150705122406/https://github.com/openlibrary/bookreader/pull/38).

PR #49, opened 2017-04-26
Fix undeclared  var "br" in Search bindings and function callbacks

DiegoPino:
There are a number of places where  variable `br` is called (a local context `this` alias) without being declared. This is wrong and breaks the binding between Search form and their callbacks. The only other workaround is to copy the affected methods and reimplement them in your own extension. 

Seems like this is because of some left overs/partial update from an upstream merge.

### DIFF
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -2932,6 +2932,7 @@ BookReader.prototype.getPageWidth2UP = function(index) {
 //______________________________________________________________________________
 BookReader.prototype.search = function(term, options) {
     options = options !== undefined ? options : {};
+    var br = this;
     var defaultOptions = {
         // {bool} (default=false) goToFirstResult - jump to the first result
         goToFirstResult: false,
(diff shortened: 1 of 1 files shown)


DiegoPino:
@rchrd2 any reason why you are closing this?

rchrd2:
Sorry. I think I fat fingered. This looks like a reasonable pull request. Though I have a couple concerns.

1) I think it would be better to change the `br.` to just `this.`
2) I am worried that `this` might not be correctly bound in some of these functions. Some of them appear to be callbacks, and might not have the right `this`. 

That said, it's definitely not okay to be assuming a global `br` exists.

Also, it looks like since I deleted the `v2` branch, this PR needs to be changed to point to master.

DiegoPino:
@rchrd2 cool. Thanks. Will pull against the main branch and address your concerns. (Have to check on 2. not sure what changed since last time if checked, but "this" was correctly bound when I tested)

PR #44, opened 2015-12-03
Cleanup code with PhpStorm

stweil:

PR #56, opened 2017-08-06
Cleanup code with PhpStorm

stweil:
This is a rebase of pull request #44.

### DIFF
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -627,7 +627,7 @@ BookReader.prototype.drawLeafsOnePage = function() {
         var topInView    = (leafTop >= scrollTop) && (leafTop <= scrollBottom);
         var bottomInView = (leafBottom >= scrollTop) && (leafBottom <= scrollBottom);
         var middleInView = (leafTop <=scrollTop) && (leafBottom>=scrollBottom);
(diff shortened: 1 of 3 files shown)


BookReader/BookReader.js
```diff
@@ -627,7 +627,7 @@ BookReader.prototype.drawLeafsOnePage = function() {
         var topInView    = (leafTop >= scrollTop) && (leafTop <= scrollBottom);
         var bottomInView = (leafBottom >= scrollTop) && (leafBottom <= scrollBottom);
         var middleInView = (leafTop <=scrollTop) && (leafBottom>=scrollBottom);
-        if (topInView | bottomInView | middleInView) {
```
rchrd2: I need to review this. There's a chance the original developers could have intended to use `|`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Bitwise_OR

Maybe it has the same effect though. Will get back to you tomorrow.

stweil: In this case the situation is simple: all three variables contain boolean values (see lines 627 ff),
so using `||` is fine.

rchrd2: Okay. That makes sense. I will merge this in. Thank you.

PR #59, opened 2017-10-25
Add options for loading tiles via AJAX and custom AJAX request headers

jonathangreen:
## Overview

This PR is based in spirit on https://github.com/openseadragon/openseadragon/pull/1055. It adds the option to load book pages using AJAX and sending custom request headers. This feature is very useful when using token-based authentication for your pages.

When the option not set then tiles should load as they did before.

### property: `loadWithAjax` (Boolean)
When this option is set on the `BookReader` all of the images are loaded in with AJAX.

### function: `getAjaxHeaders(index)`
When this function is provided on the `BookReader` it will be called with the index of the page and returns a  `String -> String` mapping of headers to be sent with the book page. This allows us to send different headers for every page if desired.

## Potential Issues

Setting the `loadWithAjax` option will limit browser support to IE 10+ due to the use of responseType = 'arraybuffer' and [createObjectURL](http://caniuse.com/#feat=bloburls). However if this option isn't used, then the behaviour is the same as it was before. 

## Feedback

Any feedback on the PR is welcome. I'd love to see it get merged.

### DIFF
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -663,6 +663,11 @@ BookReader.prototype.setupTooltips = function() {
     ;
 }
 
+// loadImage()
+//______________________________________________________________________________
+BookReader.prototype.loadImage = function(src, index, img) {
+    img.src = src;
+}
 
 // drawLeafs()
 //______________________________________________________________________________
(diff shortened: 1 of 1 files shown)


rchrd2:
Hi I am on vacation now, but will look into this when I get back next week. 
Thank you.

jonathangreen:
Hi @rchrd2 just a ping to see if you've had any chance to take a look at this.

whikloj:
+1 - subscribing

jonathangreen:
Bumping this issue to say I've resolved the merge conflict from upstream, and would still love to see this get merged.

rchrd2:
Hello. At the moment, we at the Internet Archive don't have a use case for this, and I am hesitant to add this as a core feature into BookReader, because it'll be something we have to maintain.

However, I am open to adding the `loadImage` function, which can be used as a hook for customizing the image fetching strategy. I think you could do everything you need if we just had `loadImage` as a separate function.

And then in the examples directory, we can show an example of using ajax to load images.

jonathangreen:
@rchrd2 I have modified this pull request to just create a `loadImage` stub function as you mentioned. Any review would be appreciated, I'd love to get this PR closed.

jonathangreen:
Closing since maintainer doesn't seem interested in merging.

PR #60, opened 2018-01-22
Fix regex issue

appdev312:
WEBDEV-1292

### DIFF
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -2948,12 +2948,23 @@ BookReader.prototype.search = function(term, options) {
     term = term.replace(/['"]+/g, '');
     term = '"' + term + '"';
 
-    var url = 'https://'+this.server.replace(/:.+/, ''); //remove the port and userdir
-    //url    += '/fulltext/inside.php?item_id='+this.bookId;
-    url    += '/fulltext/new_inside.php?item_id='+this.bookId;
-    url    += '&doc='+this.subPrefix;   //TODO: test with subitem
-    url    += '&path=' + (this.bookPath + '$').replace('/' + this.subPrefix + '$', '').replace('$', ''); //remove subPrefix from end of path
-    url    += '&q='+escape(term);
+    // Remove the port and userdir
+    var url = 'https://' + this.server.replace(/:.+/, '') + '/fulltext/new_inside.php?';
+
+    // Remove subPrefix from end of path
+    var path = this.bookPath;
+    if (this.bookPath.length - this.subPrefix.lastIndexOf(this.subPrefix) == this.subPrefix.length) {
+      path = this.bookPath.substr(0, this.bookPath.length - this.subPrefix.length);
+    }
+
+    var urlParams = {
+      'item_id': this.bookId,
+      'doc': this.subPrefix,
+      'path': path,
+      'q': term,
+    };
+
+    url += $.param(urlParams);
 
     if (!options.disablePopup) {
         this.showProgressPopup('<img id="searchmarker" src="'+this.imagesBaseURL + 'marker_srch-on.png'+'"> Search results will appear below...');
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -170,6 +170,10 @@ function BookReader() {
     // Experimental Controls (eg b/w)
     this.enableExperimentalControls = false;
 
+    // Base server used by some api calls
+    this.server = '';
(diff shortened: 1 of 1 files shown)


BookReader/BookReader.js
```diff
@@ -2952,7 +2952,7 @@ BookReader.prototype.search = function(term, options) {
     //url    += '/fulltext/inside.php?item_id='+this.bookId;
     url    += '/fulltext/new_inside.php?item_id='+this.bookId;
     url    += '&doc='+this.subPrefix;   //TODO: test with subitem
-    url    += '&path='+this.bookPath.replace(new RegExp('/'+this.subPrefix+'$'), ''); //remove subPrefix from end of path
+    url    += '&path=' + this.bookPath.replace('/' + this.subPrefix, ''); //remove subPrefix from end of path
```
rchrd2: This is a good step forward, but I think you need replace the last instance. For example:

```
FooBarFoo
```

Should become

```
FooBar
```

And it should only replace at the end of the string.

So 

```
FooBarFoo2
```
Should not get replaced:
```
FooBarFoo2
```

It should be this way, because the original regex used `$` at the end.

rchrd2:
Hello. This still won't. What if the string has a `$` in it. I took a closer look at this, and ended up finding the solution. I also noticed that the URL was being built in a faulty way. Can you please just change the code to this?

```
    // Remove the port and userdir
    var url = 'https://' + this.server.replace(/:.+/, '') + '/fulltext/new_inside.php?';

    // Remove subPrefix from end of path
    var path = this.bookPath;
    if (this.bookPath.length - this.subPrefix.lastIndexOf(this.subPrefix) == this.subPrefix.length) {
      path = this.bookPath.substr(0, this.bookPath.length - this.subPrefix.length);
    }

    var urlParams = {
      'item_id': this.bookId,
      'doc': this.subPrefix,
      'path': path,
      'q': term,
    };

    url += $.param(urlParams);

```

appdev312:
@rchrd2 , just pushed the change.

BookReader/BookReader.js
```diff
@@ -170,6 +170,10 @@ function BookReader() {
     // Experimental Controls (eg b/w)
     this.enableExperimentalControls = false;
 
+    // Base server used by some api calls
+    this.server = '';
+    this.searchInsideUrl = '/fulltext/new_inside.php';
```
rchrd2: Note, I made this endpoint configurable because we are moving to "inside.php" soon, and having it configurable will make it easy to roll this out.

BookReader/BookReader.js
```diff
@@ -2948,12 +2952,23 @@ BookReader.prototype.search = function(term, options) {
...
+    // Remove the port and userdir
+    var url = 'https://' + this.server.replace(/:.+/, '') + this.searchInsideUrl + '?';
+
+    // Remove subPrefix from end of path
+    var path = this.bookPath;
+    if (this.bookPath.length - this.bookPath.lastIndexOf(this.subPrefix) == this.subPrefix.length) {
+      path = this.bookPath.substr(0, this.bookPath.length - this.subPrefix.length);
+    }
+
+    var urlParams = {
```
evanminto: Suggestion: Might want to move the `var` up to avoid any hoisting problems.

evanminto:
Suggestion: Move the URL generation into its own method. Otherwise, LGTM

BookReader/BookReader.js
rchrd2: I'm not sure I understand. Wouldn't just using `urlParams` without `var` be syntactically incorrect?

evanminto: You’re familiar with variable hoisting in JavaScript, right?

Basically, right now `urlParams` is actually being declared at the top of your function scope, and on this line you’re only assigning its value, due to how JS works. I’m suggesting the common JS best practice of moving the var declaration (not necessarily the value assignment) to the top of the function, or at least closer to the top, to prevent any bugs related to this.

rchrd2: Yes I'm familiar with hoisting (you actually explained it to me previously). I don't think it will cause an issue here, and most of the rest of the code is not accounting for it either, so I think we should leave it as is.

rchrd2: I also misread that earlier. I thought you said "might want to *remove the var*". heh

evanminto: Ah, gotcha. To make my general position clear: Vars should ALWAYS be at the top, not only when context demands it. Not saying that has to happen here though. I understand not wanting to move too many things around.

PR #61, opened 2018-02-17
Add auto mode to 1up autosize (in addition to height and width).

rchrd2:
Remove the old responsiveAutofit (which is not needed anymore).

### DIFF
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -109,8 +109,7 @@ function BookReader() {
 
     // Object to hold parameters related to 1up mode
     this.onePage = {
-        autofit: 'height',       // valid values are height, width, none
-        responsiveAutofit: true, // selects the value of autofit at init
+        autofit: 'auto',       // valid values are height, width, auto, none
     };
 
     // Object to hold parameters related to 2up mode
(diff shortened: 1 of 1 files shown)


BookReader/BookReader.js
```diff
@@ -1430,13 +1418,26 @@ BookReader.prototype.nextReduce = function( currentReduce, direction, reductionF
...
-    }
-
-    // Asked for specific autofit mode
-    for (var i = 0; i < reductionFactors.length; i++) {
-        if (reductionFactors[i].autofit == direction) {
-            return reductionFactors[i];
-        }
+    } else if (direction === 'auto') {
+      // Auto mode chooses the least reduction
+      var choice = null;
```
evanminto: Suggestion: Move this declaration (not the assignment, obviously) up to the top of the function to prevent hoisting confusion.

BookReader/BookReader.js
```diff
@@ -1430,13 +1418,26 @@ BookReader.prototype.nextReduce = function( currentReduce, direction, reductionF
...
+      for (var i = 0; i < reductionFactors.length; i++) {
```
evanminto: Suggestion: Same hoisting issue as above. Also see the similar `for` loop below.

BookReader/BookReader.js
```diff
@@ -1430,13 +1418,26 @@ BookReader.prototype.nextReduce = function( currentReduce, direction, reductionF
...
+              }
+          }
+      }
+      if (choice) {
+          return choice;
+      }
+    } else if (direction === 'height' || direction === 'width') {
+      // Asked for specific autofit mode
+      for (var i = 0; i < reductionFactors.length; i++) {
+          if (reductionFactors[i].autofit == direction) {
```
evanminto: Suggestion: Probably want `===` here and elsewhere.

BookReader/BookReader.js
rchrd2: Thus far, the BookReader.js file does not deal with this, so I will leave it as is.

PR #68, opened 2018-03-30
Add additional tracking event when view changes.

bfalling:
When the view changes, a bookreader event fires which tracks download counts for the book. We'd also like to track view changes as a separate analytics "event". This PR adds the additional call.

### DIFF
--- a/BookReader/plugins/plugin.url.js
+++ b/BookReader/plugins/plugin.url.js
@@ -113,7 +113,7 @@ BookReader.prototype.updateLocationHash = function(skipAnalytics) {
                 kind: 'event',
                 ec: 'BookReader',
                 ea: 'UserChangedView',
-                el: location.pathname,
+                el: window.location.pathname,
                 cache_bust: Math.random()
             });
         }
--- a/BookReader/plugins/plugin.url.js
+++ b/BookReader/plugins/plugin.url.js
@@ -88,7 +88,7 @@ BookReader.prototype.updateLocationHash = function(skipAnalytics) {
     // Send analytics events if the location hash is changed (page flip or mode change),
     // which indicates that the user is actively reading the book. This will cause the
     // archive.org download count for this book to increment (via setting the `bookreader` value),
-    // and also send a tracking event (via settig the `kind` attribute to `event`).
+    // and also send a tracking event (via setting the `kind` attribute to `event`).
     // Note that users with Adblock Plus will not send data to analytics.archive.org
     if (!skipAnalytics && typeof(archive_analytics) != 'undefined') {
         if (this.oldLocationHash != newHash) {
--- a/BookReader/plugins/plugin.url.js
+++ b/BookReader/plugins/plugin.url.js
@@ -85,9 +85,10 @@ BookReader.prototype.updateLocationHash = function(skipAnalytics) {
         window.location.replace(newHash);
     }
 
-    // Send an analytics event if the location hash is changed (page flip or mode change),
+    // Send analytics events if the location hash is changed (page flip or mode change),
     // which indicates that the user is actively reading the book. This will cause the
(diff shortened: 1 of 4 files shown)


BookReader/plugins/plugin.url.js
```diff
@@ -85,9 +85,10 @@ BookReader.prototype.updateLocationHash = function(skipAnalytics) {
         window.location.replace(newHash);
     }
 
-    // Send an analytics event if the location hash is changed (page flip or mode change),
+    // Send analytics events if the location hash is changed (page flip or mode change),
     // which indicates that the user is actively reading the book. This will cause the
-    // archive.org download count for this book to increment.
+    // archive.org download count for this book to increment (via setting the `bookreader` value),
+    // and also send a tracking event (via settig the `kind` attribute to `event`).
```
evanminto: Blocker: Typo, "setting"

BookReader/plugins/plugin.url.js
```diff
@@ -104,7 +105,17 @@ BookReader.prototype.updateLocationHash = function(skipAnalytics) {
...
 
+            // Send bookreader ping
             archive_analytics.send_ping(values, null, 'augment_for_ao_site');
+
+            // Also send tracking event ping
+            archive_analytics.send_ping({
+                kind: 'event',
+                ec: 'BookReader',
+                ea: 'UserChangedView',
+                el: location.pathname,
```
rchrd2: Question: Do you want to be sending the path here? You could for example, send something like `this.pageIndex` (iirc).

Also, shouldn't it be `window.location.pathname` to be safe?

CHANGELOG.md
```diff
@@ -36,32 +39,32 @@ Version 3.0.0 is an effort to make BookReader more modular, extensible, and easi
...
 - Make BookReader easier to use, by adding `options` to the constructor, and adding new `options.data` option. The old way of overriding properties should still work, but it is deprecated. With `options.data`, all BookReader needs is the image URLs and dimensions. To have dynamic image URLs (eg for scaling), omit the URL from `options.data`, and include `options.getPageURI`.
-- Factor out extra features into plugins. See `plugins` directory. Example plugins include:  
-    - plugins.chapters.js - render chapter markers  
-    - plugins.search.js - add search ui, and callbacks  
-    - plugins.tts.js - add tts (read aloud) ui, sound library, and callbacks  
-    - plugins.url.js - automatically updates the browser url  
-    - plugins.resume.js - uses cookies to remember the current page  
-    - plugins.mobile_nav.js - adds responsive mobile nav to BookReader  
+- Factor out extra features into plugins. See `plugins` directory. Example plugins include:
+    - plugins.chapters.js - render chapter markers
```
rchrd2: It looks like your editor removed the trailing spaces. I believe the extra spaces are necessary for Markdown, but maybe not for github-flavored Markdown.

Given that this would probably keep happening to unsuspecting developers in the future, I guess we should just see what happens.

BookReader/plugins/plugin.url.js
bfalling: Fixed.

BookReader/plugins/plugin.url.js
bfalling: Sure. Fixed.

CHANGELOG.md
bfalling: Wow, TIL about Markdown's trailing 2 spaces indicating a <br> break. I can't help thinking what a dangerous, non-intuitive, and invisible bit of formatting that is!

In this case, they're being used after list items, so removing them doesn't seem to cause any problems.

rchrd2: I'm pretty sure it doesn't break because this is Github, but maybe I am mistaken. Either way, I am fine with this change.

PR #70, opened 2018-04-25
Add two-way iframe communication via postMessage to URL Plugin

evanminto:
Also adds some QUnit tests and updates documentation.

Jira: WEBDEV-1528


rchrd2:
In package.json there's a script for "node-qunit-phantomjs", but it's not included as a dep. Can you please add `node-qunit-phantomjs` to package.json or remove the script if it's not used.

evanminto:
My thinking on the iframe communication being its own plugin: I considered it, but it’s kind of tied to the idea of the params path and the URL. I don't see a way that it could be decoupled from that, but maybe I'm missing something.

evanminto:
I should say, it's not IMPOSSIBLE to decouple it, but it would require a bigger rethink of the interface that plugins and other code use to interact with BookReader (i.e. more based on events and less on prototype modification).

rchrd2:
But what if it was a plugin that depends on the url plugin? In the comment it could say "depends on plugin.url.js".

evanminto:
I'm not comfortable with having plugin dependencies like that, at least not right now while the plugins are all tightly coupled to the internal workings of BookReader. You'd have one plugin which makes prototype modifications and relies on prototype modifications made by another plugin. Gets pretty gnarly if you ever want to, say, change the URL plugin.

rchrd2:
The unfortunately fact is, plugins already have some dependencies. Each is
not 100% isolated.

evanminto:
This is ready for re-review, @rchrd2.

PR #78, opened 2018-09-27
WIP V4

rchrd2:


rchrd2:
Update: Nov 30, 2018. This was merged into master as v4-beta.

The rational was that this code has stabilized, and we want v4 to be discoverable and the starting point for contributions.

PR #86, opened 2018-10-23
Feature/add option not track index 0

rchrd2:


BookReader/plugins/plugin.url.js
```diff
@@ -104,46 +110,66 @@ BookReader.prototype.urlStartLocationPolling = function() {
...
-      }
-    } else {
-      window.location.replace('#' + newFragment);
+  if (currFragment === newFragment) {
+    return;
+  }
+
+  if (this.options.urlMode === 'history') {
+    if (window.history && window.history.replaceState) {
+      var baseWithoutSlash = this.options.urlHistoryBasePath = this.options.urlHistoryBasePath.replace(/\/+$/, '')
```
rchrd2: Typo. double assignment. Remove `this.options.urlHistoryBasePath = `

rchrd2:
@mekarpeles I refactored this after you reviewed. 

It turns out that filtering in `BookReader.prototype.paramsFromCurrent` was too general. We only should filter in the context of updating the URL. This led to a bug where the resume cookie was not being set, because `params.index` was pre-filtered from that function.

The new change-set is simpler in scope and cleaner.

PR #89, opened 2018-11-27
set image as background-image instead of src

psyntium:
this prevents user to easily right-click > save image as..


rchrd2:
Thanks for the PR, but this is not something that should generally be part of bookreader. BR already has this option `protected`, to prevent right click. In some cases, like with public-domain books, we actually want it to be easy to download the images.

PR #96, opened 2019-03-24
Fix touch enabled windows device issues.

hilltopmediadesign:
Chrome running on windows touch enabled devices (2 in 1's etc.) was not registering touch events: 1) In 2 page view, tapping page to go up or back a page was not working. 2) In thumbnail view, tapping thumbnail did nothing. Update fixes both issues.


hilltopmediadesign:
Tested on Asus Q525UA.

PR #97, opened 2019-04-13
Implement true fullscreen

arthuredelstein:
When the user presses the "toggle fullscreen" button, the BookReader
expands to fill the whole screen, using the browser's built-in
Fullscreen API (as used by YouTube, for example). The user gets a
"true" fullscreen experience, with no distractions.

When the user presses the ESC button, or the "toggle fullscreen"
button again, then the fullscreen mode exits.

This patch also ensures that the "share", "info" and "download"
popup modals are properly displayed in fullscreen mode.

If fullscreen mode is not supported for a given device or document,
then gracefully falls back to the legacy "full window" behavior.


arthuredelstein:
This patch works with the demos in this repository. I am hoping this functionality makes it into openlibrary.org. Unfortunately the fullscreen toggle button is hidden in the Internet Archive bookreader instance at present. The toggle button can be restored with very minor patches in https://archive.org/bookreader/BookReader-ia.css and in one other piece of inline code:
`// Wire up full-screen button, to bring back to details page
          BookReader.prototype.toggleFullscreen = function() {
            window.location.href = window.location.href.replace('/stream/', '/details/');
          };`
I'm happy to provide these patches, but I haven't found the repository that contains these files. Any advice would be appreciated!

arthuredelstein:
A demo of the patch is here: just click the "Toggle Fullscreen" button:
https://arthuredelstein.net/bookreader/BookReaderDemo/demo-simple.html

rchrd2:
On the Archive.org website, it was actually preferred to have the fullscreen that is there. If a user wants to enter "true fullscreen", they can use the browser's fullscreen button to make it full. 

I think it's nice to have the option in bookreader to have "true fullscreen", but I don't think we should remove the other fullscreen. Perhaps you can re-package this into a plugin, called `truefullscreen.plugin.js` and add new methods that coincide with the existing fullscreen methods. This way, each website can decide what they want.

rpattabi:
In case of the mobile browser, there is no explicit full screen button / option (chrome & firefox). Instead they take scrolling as a way to dismiss address bar, etc. In case of BookReader (that is part of openlibrary.org), the scrolling works but the browser keeps address bar, etc. So it is better not to assume the existence of full screen button / option in the browser.

rpattabi:
@arthuredelstein I don't know what is the better place to ask this question. Could you help me to get to take BookReader full screen in mobile browser? OpenLibrary.org's BookReader does not provide a full screen button, neither the browser like chrome or firefox.

rpattabi:
For now, I am using this [workaround](https://android.stackexchange.com/a/216675/21092).

PR #99, opened 2019-04-24
hotfix: hamburger icon invisible on mobile

### LINKED ISSUE #98: Cannot zoom in on mobile
The hamburger menu icon isn't visible, so it's impossible to zoom in on mobile

cdrini:
Closes #98 

The root issue is that the SVG is not displaying (for some reason), but for now we show the PNG by default instead.

**NOTE** I don't have a local copy of this running, so didn't try building/compiling it or anything. Just made the changes directly. Here's the before/after from applying the changes in devtools:

| Before | After |
|---|---|
| image | image |

### DIFF
--- a/src/css/_MobileNav.scss
+++ b/src/css/_MobileNav.scss
@@ -47,7 +47,8 @@ html.mm-background .BookReader {
 .BRmobileHamburger {
 	background: center center no-repeat transparent;
   background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABsAAAATCAYAAABhh3Y4AAAAAXNSR0IArs4c6QAAAFZJREFUSA1jrKmpafv//38xELMx0AgwMjL+AuJeFqAlubS0COR+qPm5TED2FJDNNPIU2FiozybT0o5Rs4dBCDCO5jNKYnE0n1ESeiNI72g+oyiyYfkMAKCpRQemMsj0AAAAAElFTkSuQmCC');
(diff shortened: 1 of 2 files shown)


src/css/_MobileNav.scss
```diff
@@ -47,7 +47,7 @@ html.mm-background .BookReader {
...
-  background: url('data:image/svg+xml;utf8,<svg width="15px" height="12px" viewBox="0 0 15 12" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"> <g id="hamburger-svg" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" stroke-linecap="square"> <path d="M0,1 L14,1" id="Line" stroke="#000000"></path> <path d="M2.39808173e-14,6 L14,6" id="Line-Copy" stroke="#000000"></path> <path d="M1.15463195e-14,11 L14,11" id="Line-Copy-2" stroke="#000000"></path> </g></svg>') center / 50% 50% no-repeat;
+  /*background: url('data:image/svg+xml;utf8,<svg width="15px" height="12px" viewBox="0 0 15 12" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"> <g id="hamburger-svg" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" stroke-linecap="square"> <path d="M0,1 L14,1" id="Line" stroke="#000000"></path> <path d="M2.39808173e-14,6 L14,6" id="Line-Copy" stroke="#000000"></path> <path d="M1.15463195e-14,11 L14,11" id="Line-Copy-2" stroke="#000000"></path> </g></svg>') center / 50% 50% no-repeat;*/
```
iisa: do we need to keep this commented out?

cdrini: Ideally we should be using the SVG line, but there's a bug in it (somewhere). That line should be fixed and used, but in the interest of progress, it's easier to just use the PNG fallback. Added as a comment, @mekarpeles created an issue 👍

rchrd2: So, this removes the SVG? Can we keep the SVG? and fallback to the PNG? SVG's are preferred.

cdrini: This comments out the SVG because it's not working right now. We should fix the SVG in a future issue, but I think it's more important to get it to display for now (why this is a hotfix).

cdrini: This is the issue mek created: https://github.com/internetarchive/bookreader/issues/100

iisa:
Created a new issue #100 - to actually fix the broken SVG

bfalling:
LGTM as well

PR #102, opened 2019-05-16
WEBDEV-1992: Fix Rapid Reload Bug in archive.org `/stream` View

jbuckner:
### Background
There is a bug in the archive.org `/stream` view where after a user selects a page from the thumbnail view, the page starts rapidly switching between the thumbnail view and the full page view, causing the browser to freeze.

### Demo
Click on a thumbnail to reproduce
Bug: https://archive.org/stream/goodytwoshoes00newyiala#mode/thumb
Fix: https://ia-petabox-review-jasonb-boo-u4y32j.archive.org/stream/goodytwoshoes00newyiala#page/n8/mode/thumb

### Cause
There were two calls being made `trigger(BookReader.eventNames.fragmentChange)`, one in `updateFirstIndex()` and another in `switchMode()`. This was triggering other listeners that kept calling back into `switchMode()` and we'd get caught in the loop.

### Solution
This adds a parameter to suppress the `trigger` calls so we can make both of the calls first, then call `trigger` when they're done.

### DIFF
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -1654,7 +1654,7 @@ BookReader.prototype.switchMode = function(mode, suppressFragmentChange) {
     }
 
     if (!suppressFragmentChange) {
-        this.trigger(BookReader.eventNames.fragmentChange);
+      this.trigger(BookReader.eventNames.fragmentChange);
     }
 };
 
@@ -2250,7 +2250,7 @@ BookReader.prototype.currentIndex = function() {
 BookReader.prototype.updateFirstIndex = function(index, suppressFragmentChange) {
     this.firstIndex = index;
     if (!suppressFragmentChange) {
-        this.trigger(BookReader.eventNames.fragmentChange);
+      this.trigger(BookReader.eventNames.fragmentChange);
     }
     this.updateNavIndexThrottled(index);
 };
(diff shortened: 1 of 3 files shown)


tests/test-bookreader.js
```diff
@@ -0,0 +1,78 @@
...
+                // we need to reset `triggerCalled` to false right before calling `updateFirstIndex` because there are
+                // other calls that trigger `fragmentChange` before we get to this point in the test
+                triggerCalled = false;
+                br.updateFirstIndex(1, true);
+                assert.notOk(triggerCalled);
+            }
+        );
+    });
+
+})
```
mekarpeles: do y'all have an policy around end of file newlines? non-blocker, just so I know and can follow w/ my commits

jbuckner: Good catch. I like adding the new line at the end. I probably didn't have VSCode set up to auto-add it.

BookReader/BookReader.js
```diff
@@ -1000,13 +1000,14 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
                 link = document.createElement("a");
                 $(link).data('leaf', leaf);
                 link.addEventListener('mouseup', function(event) {
-                  self.updateFirstIndex($(this).data('leaf'));
+                  self.updateFirstIndex($(this).data('leaf'), true);
```
mekarpeles: could be useful to have a variable for `suppressFragmentChange = true` so it's more immediately clear why the `true` is necessary in this case.

approach lgtm!

PR #105, opened 2019-06-04
Adds autoplay demo and default values for flipSpeed and flipDelay

shaneriley:
**Description**

Adds a demo page for the autoplay plugin, adds default values for `flipSpeed` and `flipDelay` rather than use hard-coded values in the autoplay plugin, and adds the ability to pass in override values to the `autoToggle` method. Added QUnit tests.

**Technical**

The version of QUnit being used pointed to a non-existent version of fsevents. Upgraded to the latest version of QUnit, which is now no longer under the package name "qunitjs" but simply "qunit".

**Testing**

Run a local web server and view the demo page at http://localhost:8000/BookReaderDemo/demo-autoplay.html

**Evidence**

bookreader_autoplay

### DIFF
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -72,8 +72,6 @@ BookReader.defaultOptions = {
     thumbPadding: 10,
     // speed for flip animation
     flipSpeed: 'fast',
-    // delay between page flips while autoplaying
-    flipDelay: 5000,
 
     showToolbar: true,
     showNavbar: true,
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -34,7 +34,7 @@ function BookReader(options) {
     this.setup(options);
 }
 
-BookReader.version = "4.2.1";
+BookReader.version = "4.2.2";
 
 // Mode constants
 BookReader.constMode1up = 1;
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -72,6 +72,8 @@ BookReader.defaultOptions = {
     thumbPadding: 10,
     // speed for flip animation
     flipSpeed: 'fast',
+    // delay between page flips while autoplaying
+    flipDelay: 5000,
 
     showToolbar: true,
     showNavbar: true,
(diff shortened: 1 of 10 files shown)


BookReaderDemo/BookReaderJSAutoplay.js
```diff
@@ -0,0 +1,57 @@
...
+    imagesBaseURL: '../BookReader/images/',
+
+    ui: 'full', // embed, full (responsive)
+
+    el: selector,
+  };
+  Object.assign(options, extraOptions);
+  var br = new BookReader(options);
+  br.init();
+  br.autoToggle(options);
```
iisa: 👍

BookReader/plugins/plugin.autoplay.js
```diff
@@ -56,9 +56,11 @@ BookReader.prototype.bindNavigationHandlers = (function(super_) {
  * Starts autoplay mode
  * TODO move to a plugin
```
rchrd2: Just noticed this PR. Looks good for the most part. You can remove the TODO comment, since it's in a plugin now.

Also, one suggestion would be to take explicit params in the autoToggle function, rather than a generic options. This locks down the methods signature. If you do want to take options, consider having a set of defaultOptions within the method, so that it's clear what type of options are supported.

BookReader/BookReader.js
```diff
@@ -223,6 +225,7 @@ BookReader.prototype.setup = function(options) {
 
     this.animating = false;
     this.flipSpeed = options.flipSpeed;
+    this.flipDelay = options.flipDelay;
```
rchrd2: If flipDelay is not used anywhere else in book reader, you may want to move this property into the autoPlay plugin.

shaneriley: That's true. I'll move it over.

BookReader/plugins/plugin.autoplay.js
shaneriley: I initially had them broken out as separate parameters, but I didn't want to force users to pass a value for the first one if they only wanted to override the second. I'll put in some defaults and extend the options on them.

PR #113, opened 2019-06-13
Fix BookReader Analytics Not Firing

jbuckner:
This fixes the initialization of the `archive_analytics` plugin. The `enableArchiveAnalytics` flag was not present during initialization (possibly a race condition?), but is available during `setup()`. Most of the other plugins use the `setup()` method for their initialization.

**NOTE** I tried writing tests for this, but there is a whole slew of initialization issues with default values not being set and in order to get the tests slightly working, I had to modify a bunch of files that change the initialization behavior. I think this can be handled independently.

### DIFF
--- a/BookReader/plugins/plugin.archive_analytics.js
+++ b/BookReader/plugins/plugin.archive_analytics.js
@@ -5,11 +5,11 @@ jQuery.extend(BookReader.defaultOptions, {
   enableArchiveAnalytics: true
 });
 
-BookReader.prototype.init = (function(super_) {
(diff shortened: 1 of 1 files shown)


rchrd2:
I believe `setup()` is supposed to take options and assign them to properties. But `init()` does actual work (rendering, binding, and so on).

jbuckner:
I was digging into this bug some more and I don't think this is actually an issue. I see the analytics firing and see the events in Google Analytics. Going to close this PR.

PR #114, opened 2019-06-13
Update js-yaml to Current Version

jbuckner:
We got a notification from Github that our version of `js-yaml` has a security vulnerability. Thisbumps it up to the current (fixed) version.


jbuckner:
@iisa I missed your comment but this is a dependency of `qunit` so we weren't managing it directly in our `package.json`

PR #128, opened 2019-08-15
adding more tests

imskr:
Am I testing DOM elements the right way?

### DIFF
--- a/tests/BookReader.test.js
+++ b/tests/BookReader.test.js
@@ -1,59 +1,118 @@
-const { br } = require('../BookReader/BookReader.js');
-const _ = require('lodash');
-const sinon = require('sinon');
+const { br } = require("../BookReader/BookReader.js");
+const $ = require("jquery");
+const _ = require("lodash");
+const sinon = require("sinon");
 
+describe("Testing BookReader.util", () => {
+  test("clamp function returns Math.min(Math.max(value, min), max)", () => {
+    expect(BookReader.util.clamp(2, 1, 3)).toEqual(2);
+  });
 
-test('clamp function returns Math.min(Math.max(value, min), max)', () => {
-    expect(BookReader.util.clamp(2,1,3)).toEqual(2);
-});
+  test("calculate a percentage suitable for CSS", () => {
+    expect(BookReader.util.cssPercentage(2, 1)).toEqual(200 + "%");
+  });
 
-test('calculate a percentage suitable for CSS', () => {
-    expect(BookReader.util.cssPercentage(2,1)).toEqual(200+'%');
-});
+  test("notInArray function which return false if value present", () => {
+    expect(BookReader.util.notInArray(4, [4, "Pete", 8, "Shubham"])).toEqual(false);
+    expect(BookReader.util.notInArray(10, ["8", "9", "10", 10 + ""])).toEqual(true);
+  });
 
-test('escapeHTML function which replaces the string', () => {
-    expect(BookReader.util.escapeHTML('Me & You')).toEqual('Me &amp; You');
-    expect(BookReader.util.escapeHTML('Me > You')).toEqual('Me &gt; You');
-    expect(BookReader.util.escapeHTML('Me < You')).toEqual('Me &lt; You');
-    expect(BookReader.util.escapeHTML('Me " You')).toEqual('Me &quot; You');
-});
+  test("escapeHTML function which replaces the string", () => {
+    expect(BookReader.util.escapeHTML("Me & You")).toEqual("Me &amp; You");
+    expect(BookReader.util.escapeHTML("Me > You")).toEqual("Me &gt; You");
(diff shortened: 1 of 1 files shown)


tests/BookReader.test.js
```diff
@@ -1,59 +1,154 @@
...
+describe("Testing setUpKeyListeners function", () => {
+  function key() {
+    alert( "Handler for .keydown() called." );
+  };
+  beforeEach(() => {
+    BookReader.prototype.setupKeyListeners();
+  });
+  test("jquery keydown", () => {
+    $(document).keydown(key);
+  });
```
jbuckner: I think for this test, you want to check to see if something has happened after the keypress has occurred. For instance, if you look inside the `setupKeyListeners` method, you can see when the user presses the left or right keys, we call `self.left()` or `self.right()`.

What I would do here is create a book reader object, override the `left` and `right` methods to do something that you can check against. ie:

```
var br = new BookReader();
br.left = jest.fn();  // mock out the `left` function
br.setupKeyListeners(); // set up the book reader keylisteners
$(document).keydown(key); // I'm not exactly sure how to simulate the "left" key getting pressed
expect(br.left).toHaveBeenCalledTimes(1); // check to make sure `left` got called
```
---
Also since you only have a single test in here, you don't necessarily need `beforeEach`. It doesn't hurt, but it's more useful if you have a bunch of tests that need the same setup and since you just have a single test, it doesn't really help that much.

tests/BookReader.test.js
```diff
@@ -1,59 +1,154 @@
...
+  function fox() {
+    alert( "The quick brown fox jumps over the lazy dog." );
+  };
+  beforeEach(() => {
+    BookReader.prototype.unbind('SKR', fox);
+  });
+  test("jquery unbind", () => {
+    $(document).bind('BookReader:' + 'SKR', fox);
+  });
 });
```
jbuckner: What are you testing in these `bind` and `unbind` tests? Are you testing to see whether the jQuery `bind` and `unbind` functionality works? If so, I wouldn't worry about that. That is functionality that is "internal" to jQuery so I don't think we need to make sure that it works.

tests/BookReader.test.js
```diff
@@ -1,59 +1,154 @@
...
-  clock.restore();
+describe('Testing Throttle function', () => {
+  let clock;
+
+  beforeEach(() => {
+    clock = sinon.useFakeTimers();
+  });
+  afterEach(() => {
+    clock.restore();
+  });
```
jbuckner: You don't really need the `beforeEach` and `afterEach` if you only have a single test in here. These methods are most useful if you have a bunch of tests that all need the same setup.

tests/BookReader.test.js
```diff
@@ -1,59 +1,154 @@
...
+    const callMethod = jest.fn();
+    const throttled = BookReader.util.throttle(() => callMethod(), 100);
+
+    const t = setInterval(() => {
+       throttled()
+    }, 50);
+
+    clock.tick(400);
+    expect(callMethod).toHaveBeenCalledTimes(4);
+    clearInterval(t);
```
jbuckner: This test looks good 👍

tests/BookReader.test.js
```diff
@@ -1,59 +1,154 @@
...
+    prop1: "SK",
+    prop2: 20
+  };
+  const obj2 = {
+    prop2: 19,
+    prop3: "SKR"
+  };
+  var modifiedNewParams = $.extend({}, obj2);
+  beforeEach(() => {
+    BookReader.prototype.extendParams(obj1, obj2);
```
jbuckner: Instead of calling the `.prototype.extendParams()`, I would instantiate a BookReader object and call the method on it. ie:
```
var br = BookReader();
br.extendParams(obj1, obj2);
```
Also, it doesn't look like you're running and `expect` calls in this test. You probably want to create an expected dictionary and compare that to the result after calling `br.extendParams()`, ie:
```
var expectedResult = { foo: "bar", baz: "boop" };
var testObject = { foo: "bar" };
var extendedObject = { baz: "boop" };
br.extendParams(testObject, extendedObject);
expect(testObject).toEqual(expectedResult);
```

tests/BookReader.test.js
imskr: Okay

tests/BookReader.test.js
imskr: Thank you for detailed review, this is very helpful

tests/BookReader.test.js
imskr: I added it to cover the coverage!

tests/BookReader.test.js
imskr: noted, I will not add beforeEach and afterEach for single tests

imskr:
@jbuckner The tests successfully simulates the keys but it shows error for `home` and `end` Keys. Also, Coverage not covering the if statement under `case KEY_UP`

PR #134, opened 2019-08-23
Removes mode switch to 2up when exiting fullscreen in < 800px width browsers

### LINKED ISSUE #103: Mobile View: coming back from full screen should keep its original view state
Problem: 
Start book reader in 1 page view in mobile size.
Click Full Screen.
Close Full Screen.
Book reader shows view in 2 page view

shaneriley:
Fixes #103

fullscreen_toggle

### DIFF
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -1720,11 +1720,6 @@ BookReader.prototype.exitFullScreen = function() {
 
     $(document).unbind('keyup', this._fullscreenCloseHandler);
 
-    var windowWidth = $(window).width();
-    if (windowWidth <= this.onePageMinBreakpoint) {
-        this.switchMode(this.constMode2up);
-    }
-
     this.isFullscreenActive = false;
     this.updateBrClasses()
 


BookReader/BookReader.js
```diff
@@ -1720,11 +1720,6 @@ BookReader.prototype.exitFullScreen = function() {
 
     $(document).unbind('keyup', this._fullscreenCloseHandler);
 
-    var windowWidth = $(window).width();
-    if (windowWidth <= this.onePageMinBreakpoint) {
-        this.switchMode(this.constMode2up);
-    }
-
```
bfalling: QUESTION: This is such a clear, specific switch. I wonder what context led to its inclusion. CC: @rchrd2

rchrd2: Hi @bfalling. Yes there was a reason for this. It is this way, so that on mobile devices, on the details pages, and not in full screen, we show the little 2 page book. And then when you go to full screen you view the 1 page mode. -- The scrolling 1 page mode on the details page is not great, because you are scrolling within a div, and it's not very usable. The idea was to show a small overview of the book, but then require the the user to go to full screen to actually read the book.

You can see the original designs here: https://docs.google.com/presentation/d/1b72JUiW2ICc0ehwIUfDTdwCRaMhsUKsz35zyYHDShrk/edit#slide=id.p

PR #135, opened 2019-08-23
Creates vendor fullscreen plugin

shaneriley:
This creates a vendor native fullscreen mode by way of a plugin, ported from https://github.com/internetarchive/bookreader/pull/97

### DIFF
--- a/BookReader/plugins/plugin.vendor-fullscreen.js
+++ b/BookReader/plugins/plugin.vendor-fullscreen.js
@@ -1,0 +1,189 @@
+/**
+ * Toggles browser's native fullscreen mode if available
+ */
+
+
+(function(BR) {
+  var event_namespace = ".bookreader_vendor-fullscreen";
+
+  jQuery.extend(BR.defaultOptions, {
+    enableVendorFullscreenPlugin: true
+  });
+
+  BR.prototype.setup = (function(super_) {
+    return function(options) {
+      super_.call(this, options);
+
+      this.isVendorFullscreenActive = false;
+    };
+  })(BR.prototype.setup);
+
+  BR.prototype.getInitialMode = (function(super_) {
+    return function(params) {
+      var nextMode = super_.call(this, params);
+      if (this.isVendorFullscreenActive) {
+        nextMode = this.constMode1up;
+      }
+      return nextMode;
+    };
+  })(BR.prototype.getInitialMode);
+
+  BR.prototype.init = (function(super_) {
+    return function() {
+      super_.call(this);
+
+      if (BR.util.fullscreenAllowed()) {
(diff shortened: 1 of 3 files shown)


BookReader/plugins/plugin.vendor-fullscreen.js
```diff
@@ -0,0 +1,189 @@
...
+  BR.util.requestFullscreen = function(element) {
+      if (element.requestFullscreen) {
+        element.requestFullscreen();
+      } else if (element.webkitRequestFullscreen) {
+        element.webkitRequestFullscreen();
+      } else if (element.mozRequestFullScreen) {
+        element.mozRequestFullScreen();
+      } else if (element.msRequestFullscreen) {
+        element.msRequestFullscreen();
+      }
```
jbuckner: There's some spacing inconsistencies.. like 4/2 spaces here and up above.

BookReader/plugins/plugin.vendor-fullscreen.js
```diff
@@ -0,0 +1,189 @@
...
+      return nextMode;
+    };
+  })(BR.prototype.getInitialMode);
+
+  BR.prototype.init = (function(super_) {
+    return function() {
+      super_.call(this);
+
+      if (BR.util.fullscreenAllowed()) {
```
jbuckner: You can return early since the entire rest of this method is inside this `if` statement.

BookReader/plugins/plugin.vendor-fullscreen.js
```diff
@@ -0,0 +1,189 @@
...
+    }
+  })(BR.prototype.init);
+
+  BR.prototype.enterFullWindow = function() {
+    this.refs.$brContainer.css('opacity', 0);
+
+    var windowWidth = $(window).width();
+    if (windowWidth <= this.onePageMinBreakpoint) {
+      this.switchMode(this.constMode1up);
+    }
```
jbuckner: Is this the same "bug" from #134 ?

BookReader/plugins/plugin.vendor-fullscreen.js
shaneriley: Yeah, there's likely some cleanup that can be done throughout. Things that stuck out while I was scanning it were addressed, but there's likely more to do with it.

BookReader/plugins/plugin.vendor-fullscreen.js
shaneriley: I think this was the original author's way of changing the behavior in the same way that my fix did, where I just removed the check entirely.

rpattabi:
Any pointers on how to make use of this? I don't see full screen option when reading (openlibrary). Or is it just a matter of time openlibrary incorporates this?

PR #137, opened 2019-09-02
Skips vendor native setup

### LINKED ISSUE #136: Skips vendor-native fullscreen setup if on mobile device
**Description**

Nearly all browsers on mobile devices prevent fullscreen mode. To retain existing fullscreen-like functionality, the setup should be skipped if on a mobile device.

**Expectation**

When on a mobile device, tapping the fullscreen icon should use the existing fullscreen mode rather than the vendor native fullscreen.

**Context**

Any mobile device.

**Proposal & Constraints**

Restore prototype methods and exit early from init method if device is mobile.

shaneriley:
Fixes #136

### DIFF
--- a/BookReader/plugins/plugin.vendor-fullscreen.js
+++ b/BookReader/plugins/plugin.vendor-fullscreen.js
@@ -2,15 +2,19 @@
  * Toggles browser's native fullscreen mode if available
  */
 
-
 (function(BR) {
-  var event_namespace = ".bookreader_vendor-fullscreen";
+  var event_namespace = '.bookreader_vendor-fullscreen';
+  var super_methods = {};
+  function isMobile() {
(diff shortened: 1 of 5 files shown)


BookReader/plugins/plugin.vendor-fullscreen.js
```diff
@@ -29,6 +34,12 @@
   })(BR.prototype.getInitialMode);
 
   BR.prototype.init = (function(super_) {
+    if (isMobile()) {
+      for (var method in super_props) {
```
bfalling: QUESTION: Where is `super_props` introduced?

shaneriley: Whoops! Should be super_methods

Issue #138, opened 2019-09-03
Refactor plugin.tts.js to not expose SpeechEngine methods on BookReader


cdrini:
Because we will be adding a lot of options to AudioReader, we don't want to pollute the `BookReader` prototype namespace with a bunch of facade methods for the `SpeechSynthesisEngine`. So let's instead define `BookReader.tts` to reference a speech engine, and call methods on that (e.g. Before: `BookReader.ttsStart()`, we'll have `BookReader.tts.start()`).

### Summary of Requirements
- [ ] `BookReader` should no longer have `tts`-related methods
- [ ] `BookReader` should have a `tts` object of type `SpeechSynthesisEngine` which performs all the methods previously performed by the `tts` methods
- [ ] Law of refactor: there should be no observable changes.

### Relationships
- Part of #94
- blocked by #129

cdrini:
Largely done by #1376

PR #140, opened 2019-09-13
Refactor (rewrite?) plugin.tts.js + add es6 support to plugin.tts.js

### LINKED ISSUE #129: Refactor plugin.tts.js to include class SpeechSynthesisEngine
Because we will be introducing another way for speech synthesis to occur (i.e. using the browser APIs), we will lay the ground-work for that change by abstracting the current implementation into a `SpeechSynthesisEngine` interface/abstract class. In a future issue, we will be adding another concrete class that implements this interface to synthesize using the native Browser apis.

**Note** that JS doesn't natively support things like interfaces or abstract classes (or even classes in ES5), so the exact implementation details are left to the developer so long as the abstraction requirements are met.

### Summary of Requirements
- [ ] There should be some interface/abstract class/whatever called `SpeechSynthesisEngine`
- [ ]  There should be a class, `FestivalSpeechEngine` which implements this interface/class
- [ ] `plugin.tts.js` will construct an instance of the speech engine and call _only_ methods define on the interface/abstract class.
- [ ] Law of refactor: there should be no observable changes.

### Relationships
- Part of #94

cdrini:
Refactor plugin.tts in preparation for Web Speech API support and Festival/WebSpeech interoperability. Turns out trying to do this without classes/Promises was really difficult, so added es6 support there as well.

Fixes #129

## Summary of Requirements
- [X] There should be some interface/abstract class/whatever called `SpeechSynthesisEngine`
    - named it `AbstractTTSEngine`
- [X]  There should be a class, `FestivalSpeechEngine` which implements this interface/class
    - named `FestivalTTSEngine`
- [X] `plugin.tts.js` will construct an instance of the speech engine and call _only_ methods define on the interface/abstract class.
- [~] Law of refactor: there should be no observable changes.
    - there were some small observable changes (see Notes)

## Notes

- Imported a `Promise`s library polyfill; there's a decent amount of async code inherent to playing audio files, so using Promises to avoid "callback hell"
- Observable changes:
    - This ~introduces~ exchanges a bug. Before: In 2up mode, clicking next page would _often_ stop ReadAloud. After: Never stops ReadAloud. I've opted to for the new bug because the stop button will take you out of the undesired state while still making continuous ReadAloud possible. The old bug prevented continuous ReadAloud in 2up with no work-around.
    - ~Because we now buffer 2 pages and 2 chunks of audio, there is never a long pause between paragraphs or pages; they will start playing immediately.~ This doesn't work on mobile, so scrapped.
    - Probably other things :( There are a number of changes, so testing would be appreciated.
- After this, adding Web Speech API support will be a max 2 commit change :)

## Open Questions
- Should I import the es6 promises polyfill into the plugin.tts.js? I don't want to change how these files are used
- Having the compiled JS file in the repo will cause merge conflicts super easily. Maybe we can `gitignore` the file, and only have it be updated when a release occurs?
- The only library that I could find that does something similar to `AsyncStream` is highland.js, but that's like 100KB :/ If anyone has any recommendations, would love to replace that with a 3rd party module

## Testing
- Tested with https://archive.org/details/tinypicturebook00chariala which contains all the cases
- Tested on Windows 10: Firefox 69.0 (64-bit); Chrome 76 (64-bit); Edge 44.18362.329.0; IE11
  - ✔️✔️✔️✔️ Empty pages are shown (in 2up) but do not affect ReadAloud
  - ✔️✔️✔️✔️ ReadAloud waits for page flip animation
  - ✔️✔️✔️✔️ Works in 2UP mode across pages
  - ✔️✔️✔️✔️ Works in 1UP mode across pages
  - ✔️✔️✔️✔️ Behaves normally when reaching book end
  - ✔️✔️✔️✔️ Highlights appear while text is being spoken

## Stakeholders
@mekarpeles @jbuckner @shaneriley


BookReader/es6-promise-polyfill/promise.min.js
```diff
@@ -0,0 +1,6 @@
+(function(t){function z(){for(var a=0;a<g.length;a++)g[a][0](g[a][1]);g=[];m=!1}function n(a,b){g.push([a,b]);m||(m=!0,A(z,0))}function B(a,b){function c(a){p(b,a)}function h(a){k(b,a)}try{a(c,h)}catch(d){h(d)}}function u(a){var b=a.owner,c=b.state_,b=b.data_,h=a[c];a=a.then;if("function"===typeof h){c=l;try{b=h(b)}catch(d){k(a,d)}}v(a,b)||(c===l&&p(a,b),c===q&&k(a,b))}function v(a,b){var c;try{if(a===b)throw new TypeError("A promises callback cannot return that same promise.");if(b&&("function"===
```
jbuckner: Is it possible for us to import this via npm instead of including the file in the repo?

BookReader/plugins/plugin.tts.js
```diff
@@ -1,527 +1,2 @@
...
...afTop = "+n+" topOfFirstChunk = "+i+" botOfLastChunk = "+s);var u=this.refs.$brContainer.prop("scrollTop"),a=u+this.refs.$brContainer.height();soundManager.debugMode&&console.log("containerTop = "+u+" containerBot = "+a),(i<u||s>a)&&this.refs.$brContainer.stop(!0).animate({scrollTop:i},"fast")}},BookReader.prototype.ttsHilite1UP=function(t){for(var e=0;e<t.lineRects.length;e++){var n=t.lineRects[e][0],o=t.lineRects[e][1],r=t.lineRects[e][2],i=t.lineRects[e][3],s=document.createElement("div");this.ttsHilites.push(s),$(s).prop("className","BookReaderSearchHilite").appendTo(this.$(".pagediv"+t.leafIndex)),$(s).css({width:(r-n)/this.reduce+"px",height:(o-i)/this.reduce+"px",left:n/this.reduce+"px",top:i/this.reduce+"px"})}},BookReader.prototype.ttsHilite2UP=function(t){for(var e=0;e<t.lineRects.length;e++){var n=t.lineRects[e][0],o=t.lineRects[e][1],r=t.lineRects[e][2],i=t.lineRects[e][3],s=document.createElement("div");this.ttsHilites.push(s),$(s).prop("className","BookReaderSearchHilite").css("zIndex",3).appendTo(this.refs.$brTwoPageView),this.setHilightCss2UP(s,t.leafIndex,n,r,i,o)}},BookReader.prototype.ttsRemoveHilites=function(){$(this.ttsHilites).remove(),this.ttsHilites=[]}}]);
```
jbuckner: How come this file is minified?

BookReader/plugins/plugin.tts.js.map
```diff
@@ -0,0 +1 @@
...return {AbstractStream<PageChunk & TSound>}\n     */\n    getPlayStream() { return null; }\n\n    /**\n     * @abstract\n     * @param {PageChunk & TSound} chunk\n     * @return {PromiseLike} promise called once playing finished\n     */\n    playChunk(chunk) { return null; }\n\n    /**\n     * @private\n     * Gets the text on a page with given index\n     * @param {Number} leafIndex\n     * @return {PromiseLike<Array<PageChunk>>}\n     */\n    fetchPageChunks(leafIndex) {\n        return $.ajax({\n            type: 'GET',\n            url: 'https://'+this.opts.server+'/BookReader/BookReaderGetTextWrapper.php',\n            dataType:'jsonp',\n            data: {\n                path: this.opts.bookPath+'_djvu.xml',\n                page: leafIndex\n            }\n        })\n        .then(\n            /** @param {Array<[String, ...Array<DJVURect>]>} chunks */\n            function (chunks) {\n                return chunks.map(c => {\n                    return {\n                        leafIndex,\n                        text: c[0],\n                        lineRects: c.slice(1)\n                    };\n                });\n            }\n        );\n    }\n}"],"sourceRoot":""}
```
jbuckner: Can we generate the map file at build/deploy time instead of checking it into the repo?

src/js/plugins/tts/AbstractTTSEngine.js
```diff
@@ -0,0 +1,136 @@
+import AsyncStream from './AsyncStream.js';
+
+/**
+ * @export
+ * @typedef {Object} PageChunk
+ * @property {Number} leafIndex
+ * @property {String} text
+ * @property {DJVURect[]} lineRects
```
jbuckner: I'm kind of curious about these three `typedef`s. What is the goal of these? Are they intended to define objects for external use? I'm inclined to define these as actual classes instead of defining them in a comment block like this, but I may not be understanding their use case. ie:
```
class PageChunk {
  constructor(leafIndex, text, lineRects) {
    this.leafIndex = leafIndex;
    this.text = text;
    this.lineRects = lineRects;
  }
}

class TTSEngineOptions {
  constructor(server, bookPath, onLoadingStart, ....) {
    this.server = server;
    this.bookPath = bookPath;
    ....
  }
}
```

src/js/plugins/tts/plugin.tts.js
```diff
@@ -0,0 +1,253 @@
+/**
+ * Plugin for Text to Speech in BookReader
+ */
+
+import FestivalTTSEngine from './FestivalTTSEngine.js';
+
+// Default options for TTS
+jQuery.extend(BookReader.defaultOptions, {
+    server: 'ia600609.us.archive.org',
```
jbuckner: Should we be hardcoding this address?

src/js/plugins/tts/plugin.tts.js
```diff
@@ -0,0 +1,253 @@
...
+        if (this.options.enableTtsPlugin && this.ttsEngine.isSupported) {
+            $("<button class='BRicon read js-tooltip'></button>").insertAfter($el.find('.BRpage .BRicon.thumb'));
+        }
+        return $el;
+    };
+})(BookReader.prototype.initNavbar);
+
+// ttsToggle()
+//______________________________________________________________________________
+BookReader.prototype.ttsToggle = function () {
```
jbuckner: Since you've got several methods in here that are tts-specific, what do you think of namespacing them instead of putting them all at the top-level? ie `BookReader.tts.toggle()`, `BookReader.tts.start()`, etc.

src/js/plugins/tts/AsyncStream.js
```diff
@@ -0,0 +1,212 @@
...
+ * Class that makes it easier to deal with a series of async events
+ * (represented as Promises). All the function let you interact with
+ * the promise's value instead of the promise itself, letting you 
+ * treat it as a stream of synchronuous values. The notable exception
+ * being the `pull` method, which should be used to consume the stream,
+ * and hence must return the promise directly.
+ * @template T
+ */
+export default class AsyncStream {
+    /** @typedef {TStreamItem<T>} TItem */
```
jbuckner: This looks like it'd be a good use case for TypeScript generics 😉  _hides_ 🙈

BookReader/es6-promise-polyfill/promise.min.js
cdrini: :+1: Will do

src/js/plugins/tts/plugin.tts.js
cdrini: This is the way it was before (for better or worse) :P

cdrini:
Yaaay! Thanks for taking the time to review this, @jbuckner ! I know it's a doozy 😅

src/js/plugins/tts/AsyncStream.js
cdrini: Hahaha; vs code does autocompletion for these, too! I broke it with the `flatten` method, but everything else here typechecks. If you add `//@ts-check` to the top of a JS file, it'll **highlight all the type violations** 😍

src/js/plugins/tts/plugin.tts.js
cdrini: https://github.com/internetarchive/bookreader/issues/138 :) In the interest of keeping the scope of this small (lol), I wanted to avoid doing that as part of this PR. It is planned, but this first refactor took longer than anticipated, so I may have to punt that for later :/

BookReader/plugins/plugin.tts.js
cdrini: Mostly because that's the webpack default. I kind of want it to be minified, because then it's less likely someone will edit it by accident and not edit the original. Thoughts?

BookReader/plugins/plugin.tts.js.map
cdrini: This is one of the bigger questions I have about the approach I'm using. I'm going to post a response to the full PR for better visibility.

src/js/plugins/tts/AbstractTTSEngine.js
cdrini: They're there to let us better document types, mainly (in these files or others). i.e. "What parameter does this function take? It takes a POJO with these fields". Other languages have features like `data class`es which would be appropriate for these, but I feel full `class`es might be a little too much, since these don't have any init logic or associated functionality (ie methods). They also cause a decent amount of duplication with the parameter name/`this.* = *`.

I think converting these to classes sets a precedent that _all_ POJOs (e.g. `{ x: 'foo' }`) should be converted to es6 classes, which I kind of disagree with.

cdrini:
@jbuckner (wrt https://github.com/internetarchive/bookreader/pull/140#discussion_r324813663 )

Basically the main requirement I've place on myself for this PR is: There should be no deploy changes; the repo should just be checked out, and everything should work. (I.e. no HTML `<script>` import updates).

Assumption: Deploying is essentially just a `git checkout`.

To maintain that, I've
1) Included the compiled JS file in the repo; that's not best practice, but it avoids having the build step be run on a server/etc
2) I've included the source-map so that debugging will still work as before (again, in the repo). This is also not best practice

Possible avenues to avoid including the build step:
1. Remove compiled JS/sourcemap, and attach the built files to the release (see https://help.github.com/en/articles/creating-releases ). This might require changes to our deploy process. This would also allow for future automation with `GitHub Actions`
2. gitignore the build files, except for releases. Reduces merge conflict headaches, but not much else

Are my assumptions correct? Any other possible solutions I'm missing? (Also cc @shaneriley who most recently deployed this code).

cdrini:
(I like Option 1 _a lot_ actually now that I phrase it)

BookReader/es6-promise-polyfill/promise.min.js
cdrini: done

BookReader/plugins/plugin.tts.js.map
cdrini: As discussed; map file is now vanquished \o/

BookReader/plugins/plugin.tts.js
cdrini: These files are now no longer in the repo, so should be good.

cdrini:
Discussed with @jbuckner in person, and it looks like we're installing bookreader via npm. So we remove the build files and replaced them with `prepare` step of `npm` (see https://docs.npmjs.com/misc/scripts ), and gitignore the build files (to make dev's lives easier). I like this solution; the only caveat is that now someone _can't_ just download the repo and link to the `BookReader` files (since some of those files don't exist :P).

Issue #132, opened 2019-08-19
BrowserSpeechEngine should choose voice based on book language


cdrini:
Edit `BrowserSpeechEngine` so that it chooses the synthesizer's voice based on the language of the book.

### Summary of Requirements
- [ ] Should choose most appropriate voice from those available
- [ ] Should default to the first available voice if preferred voice not available

### Relationships
- Part of #94 
- Depends on #131 , #130

PR #146, opened 2019-09-24
Create/use WebTTSEngine in tts plugin

### LINKED ISSUE #130: Create/use BrowserSpeechEngine class in plugin.tts.js
Create a new subclass of `SpeechSynthesisEngine` that uses the native browser apis.

### Summary of Requirements
- [ ] New class `BrowserSpeechEngine` that implements `SpeechSynthesisEngine`
- [ ] Browsers with speech synthesis support should use `BrowserSpeechEngine`
- [ ] Browsers without speech synthesis support should use `FestivalSpeechEngine`

### Relationships
- Part of #94
- Depends on #129

cdrini:
Create a new class, `WebTTSEngine`, that performs TTS using the browser's Web Speech APIs

Closes #130

## Summary of Requirements
- [X] New class `BrowserSpeechEngine` that implements `SpeechSynthesisEngine`
   - Called it `WebTTSEngine`
- [X] Browsers with speech synthesis support should use `BrowserSpeechEngine`
- [X] Browsers without speech synthesis support should use `FestivalSpeechEngine`

## Notes
- ~Implemented a `find` poly fill; don't like this cause there are existing libraries, but adding one doubled the size of the webpack bundle :/~
- Used browserlist + babel's core-js to automatically provide polyfills. This did see an increase in the filesize (~20kb before gzip), but considering a single page's JPG ~130kb, I think that's worth the reduction of code complexity, increase of development speed, and security of using 3rd party polyfills (better tested).
- Observable changes: New browsers will now use the default system's voice to read the book. This doesn't mean it'll use the Google voices; that will happen as part of #132 . Older browsers should be unaffected.

## Open Questions
- ~Can we _not_ manually polyfill `Array.prototype.find`?~ Yes!

## Testing
Tested with https://archive.org/details/farmfolkapicture00bren . The failure also occurs on prod, so unrelated to these changes.

| IE11 | Edge | FF69 | Chrome 77 | FF-N@And6 | OpM@And6 | Chrom@And6 | Saf?@iPad | Test |
| --- | --- | --- | --- | --- |--- | --- | --- | --- |
| ✔ | ✔ | ✔ | ✔ | ✔ | ✔ | ✔ | ✔ | Moves btwn paragraphs |
| ✔ | ✔ | ✔ | ✔ | ✔ | ✔ | ✔ | ✔ | Moves btwn pages |
| ✔ | ✔ | ✔ | ✔ | ✔ | ✔ | ✔ | ✔ | Works in 1up mode |
| - | - | - | - | ✔ | ❌ | ✔ | ❌ | Works with screen off |
| - | ✔ | ✔ | ✔ | ✔ | - | ✔ | ✔ | Doesn't use festival |
| ✔ | - | - | - | - | ✔ | - | - | Uses festival |

## Stakeholders
@mekarpeles @jbuckner


cdrini:
This is blocked by #149

src/js/plugins/tts/WebTTSEngine.js
```diff
@@ -0,0 +1,93 @@
...
+     * @param {PromiseLike} endPromise promise that will file once the utterance is done
+     * @return {PromiseLike}
+     */
+    chromePausingBugFix(endPromise) {
+        const sleepPromise = sleep(14000).then(() => 'timedout');
+        return Promise.race([sleepPromise, endPromise])
+        .then(result => {
+            if (result == 'timedout') {
+                window.speechSynthesis.pause();
+                return sleep(25)
```
mekarpeles: is 25ms synonymous with instant?

src/js/plugins/tts/polyfills.js
```diff
@@ -0,0 +1,36 @@
...
+ * - https://github.com/paulmillr/Array.prototype.find/blob/master/implementation.js
+ * Notable differences: we don't throw errors in as many cases. We don't use the npm
+ * module because it ends up being very large (for some reason); ~23kB!
+ * @template T
+ * @this {T[]}
+ * @param {(element: T, index: number, array: T[]) => boolean} predicate 
+ * @param {any} [thisArg] argument used as "this" for predicate
+ * @return {T | undefined}
+ */
+export function find(predicate, thisArg=undefined) {
```
mekarpeles: Is the tts module the right place for this utility?

src/js/plugins/tts/utils.js
```diff
@@ -0,0 +1,21 @@
...
+}
+
+/**
+ * Checks whether the current browser is a Chrome/Chromium browser
+ * Code from https://stackoverflow.com/a/4565120/2317712
+ * @param {string} [userAgent]
+ * @param {string} [vendor]
+ * @return {boolean}
+ */
+export function isChrome(userAgent=navigator.userAgent, vendor=navigator.vendor) {
```
mekarpeles: Is this tts specific? Or is there a more central location where shared utilities should go? @jbuckner @iisa @cdrini any thoughts?

cdrini: Functions like these are generally considered bad practice; instead of browser sniffing, feature detection is the recommended practice. But I couldn't think of a way to feature detect this bug :/ keeping this scoped here makes it less likely to be reused.

src/js/plugins/tts/polyfills.js
cdrini: Killed in favour of automatic polyfilling.

src/js/plugins/tts/WebTTSEngine.js
cdrini: Yep :p

cdrini:
Known issues: iOS reads English in an Arabic accent; will be fixed by #132 .

PR #155, opened 2019-10-05
Add eslint rule comma-dangle

nirmalks:
* Add comma-dangle rule only-multiline


cdrini:
Thanks @nirmalks ! Could you also run an eslint fix step to fix any existing violations of this rule?

nirmalks:
> Thanks @nirmalks ! Could you also run an eslint fix step to fix any existing violations of this rule?

Yeah, I will look into it

iisa:
Hi @nirmalks - are you still interested in resolving this issue?

nirmalks:
> Hi @nirmalks - are you still interested in resolving this issue?

Yes, I'm looking into this now. I will update in case if I face any issues. Sorry for the delay

nirmalks:
> Thanks @nirmalks ! Could you also run an eslint fix step to fix any existing violations of this rule?

1. One doubt,"only-multiline" - it is mentioned that it does not require trailing commas. Running eslint fix does not fix any code. 

>  "only-multiline" allows (but does not require) trailing commas when the last element or property is in a different line than the closing ] or } and disallows trailing commas when the last element or property is on the same line as the closing ] or }

2. "always-multiline" - mandates trailing commas and eslint fix fixes the code. 
"always-multiline" requires trailing commas when the last element or property is in a different line than the closing ] or } and disallows trailing commas when the last element or property is on the same line as the closing ] or }
Can i add always-multiline or do you want to persist with "only-multiline"(its mentioned explicitly in the issue) and fix the existing code manually.

iisa:
@nirmalks , thanks again for taking this on, can you rebase? i think that will allow us to merge it in. @cdrini

nirmalks:
> @nirmalks , thanks again for taking this on, can you rebase? i think that will allow us to merge it in. @cdrini

@iisa So can I stick with "only-multiline" or change the rule to "always-multiline" ? Will rebase accordingly

cdrini:
Sorry for the delay on this! "only-multiline" is good :+1:

PR #157, opened 2019-10-09
Include build files in repo (for now)

### LINKED ISSUE #156: Plugins not transpiled correctly on archive.org
**Description**
When loaded into petabox, the compiled files still include es6 keywords (like class).

**Expectation**
They should be identical to the files generated when running webpack locally.

cdrini:
Closes #156. To avoid the headaches of the build step having to run on petabox, we run the build step locally and include the files in the repo. This means downstream users of this package don't need to run a build step. Long term solution would probably be just to publish this on npm, but this is blocking a whole host of changes from happening.

## Notes
- add the build files into the repo
- create a release flow; this ensure that the build files are always up-to-date at least during a release.
- test release flow by version bumping to v4.3.0

## Testing
- ✅ petabox no longer runs a postinstall step, and has es5-compatible JS

## Stakeholders
@iisa @mitra42 @mekarpeles @jbuckner 

*Release Link:* [Create Release](https://github.com/internetarchive/bookreader/releases/new?tag=v4.3.0&title=v4.3.0&body=-%20New%20plugin%2C%20%60fullscreen_menu_toggle%60%2C%20allowing%20center-tap%20to%20hide%20UI%0A-%20TTS%20Plugin%20logs%20analytics%20events%20(if%20enabled)%0A-%20ES6%20compilation%20step%20no%20longer%20performed%20on%20client%3B%20build%20files%20are%20included%20in%20repo%20so%20clients%20have%20no%20build%20step%0A) (To do after merged)


scripts/version.js
```diff
@@ -0,0 +1,28 @@
+const fs = require('fs');
+const { execSync } = require('child_process');
+const { version: NEW_VERSION } = require('../package.json');
+
+async function main() {
```
jbuckner: Oh this is cool!

.gitattributes
```diff
@@ -0,0 +1,5 @@
+# Build files
+BookReader/plugins/*.js.map binary
+BookReader/plugins/plugin.archive_analytics.js binary
+BookReader/plugins/plugin.fullscreen_menu_toggle.js binary
+BookReader/plugins/plugin.tts.js binary
```
jbuckner: How come you mark these as `binary`?

scripts/version.js
cdrini: 😊 Haha, one of the benefits of not having to worry about client environments is that I can use latest node features :P

mitra42:
Thanks - and incidentally this should fix two problems I have on dweb-mirror: a) the missing plugins that this was intended to fix and b) the really slow `sass` build our partner was complaining about.

cdrini:
@mitra42 Awesome! Let me know if you hit any new issues :)

PR #163, opened 2019-10-15
menu toggle - provide escape hatches for nav traversal returning exception

iisa:
### Problem:
Access Restricted View does not have a navbar. (who would've thunk?)

### Solution:
For menu toggle - check if navbar is present.  IF so, then run plugin, if not return immediately.

#### Access Restricted Book: 
https://www-isa.archive.org/details/alopeciaareata0000levy
vs.
https://archive.org/details/alopeciaareata0000levy

Page successfully renders.
Image in Theatre shows.

#### Borrowable book:
https://www-isa.archive.org/details/citadel00cron_1/page/n423
vs.
https://archive.org/details/citadel00cron_1/page/n423

Page successfully renders.
Theatre shows bookreader.
`www-isa` shows navbar without triangle
`www-isa` black bg can toggle nav
`www-isa` 2page view, clicking center toggles nav


src/js/plugins/menu_toggle/plugin.menu_toggle.js
```diff
@@ -97,6 +103,18 @@
 
     BookReader.prototype.initMenuToggle = function brInitMenuToggle(e) {
       var br = this;
+      let hasNav = false;
+
+      try {
+        hasNav = br.navigationIsVisible();
+      } catch {
+        hasNav = false;
```
shaneriley: This statement is redundant, since it's already set to false on instantiation.

src/js/plugins/menu_toggle/plugin.menu_toggle.js
```diff
@@ -97,6 +103,18 @@
 
     BookReader.prototype.initMenuToggle = function brInitMenuToggle(e) {
       var br = this;
+      let hasNav = false;
```
bfalling: QUESTION: Does this need to be ES5 compatible, i.e. use var instead of let?

cdrini: No; that's what the build step is for.

src/js/plugins/menu_toggle/plugin.menu_toggle.js
```diff
@@ -13,23 +13,29 @@
...
+      hideArrow(br);
       removeToggleFromNav();
       br.showNavigation();
     }
   
     var removeToggleFromNav = function removeToggleFromNav(br) {
+      if (!br.refs || !br.refs.$BRnav) {
+        return;
+      }
       var $menuTab = br.refs.$BRnav.children('.js-tooltip');
```
mc2: And, this looks like a complex object... 
**Question**: should this be documented in the header? Where to find the object definition?

iisa: good notion, i just added a comment at the top of the file to note this.

bfalling:
LGTM

PR #171, opened 2019-10-25
Updating Show/Hide Navigation functionality

iisa:
Adding the new `menu_toggle` plugin surfaced a few places that need optimization in the core code around:
- perceived slowness in navigation animation at certain modes
- in 1up mode, content viewport does not re-adjust to make use of the new real estate

Fixing the above consisted of:
- a refactor in the Show/Hide navigation functionality
  - animation config now has:
    - duration
    - callback that handles book window resizing for a smoother UX experience in 1up mode
    - new custom event: `navToggled` - that fires every time the nav toggles

Plugin code changes handles the following:
- throttling event handlers to make Nav animations more performant
- better variable naming
- adding JSDocs


BookReader/BookReader.js
```diff
@@ -3873,16 +3890,50 @@ BookReader.prototype.navigationIsVisible = function() {
...
  * Hide navigation elements, if visible
  */
 BookReader.prototype.hideNavigation = function() {
     // Check if navigation is showing
     if (this.navigationIsVisible()) {
-        var toolbarHeight = this.getToolBarHeight();
-        var navbarHeight = this.getNavHeight();
-        this.refs.$BRtoolbar.animate({top: toolbarHeight * -1});
-        this.refs.$BRnav.animate({bottom: navbarHeight * -1});
+        var hide = true;
```
jbuckner: Does the value of `hide` ever change or can you just pass `true` to `toggleNavigation`?

BookReader/BookReader.js
```diff
@@ -1240,12 +1246,23 @@ BookReader.prototype.zoom1up = function(direction) {
...
-    top: this.getToolBarHeight(),
-    bottom: this.getNavHeight(),
-  });
+BookReader.prototype.resizeBRcontainer = function(fillMainContainer) {
+    if (!fillMainContainer) {
+        return this.refs.$brContainer.css({
+            top: this.getToolBarHeight(),
+            bottom: this.getNavHeight()
+          });
+    }
```
jbuckner: Not a blocker, but since these two are mutually exclusive options (and you're not actually returning anything), I might do `if / else` instead of `if / return`.

src/js/plugins/menu_toggle/plugin.menu_toggle.js
```diff
@@ -1,113 +1,194 @@
...
       removeEventHandlers(br);
+      br.showNavigation();
     }
   
+    /**
+     * Removes click handlers on elements that house the book pages
+     *
+     * @param { object } br - BookReader instance
+     */
     var removeEventHandlers = function removeEventHandlers(br) {
```
jbuckner: Is this method the opposite of the `registerClickHandlers` function that was renamed? If so I might rename this to `removeClickHandlers` for parallel naming structure.

src/js/plugins/menu_toggle/plugin.menu_toggle.js
```diff
@@ -1,113 +1,194 @@
...
+      return isCenterClick;
     }
 
-    function registerEventHandlers(br) {
+    /**
+     * Confirms whether or not the click happened in the background
+     *
+     * @param { object } event - JS event object
+     */
+    var isBackground = function isBackground(event) {
```
jbuckner: I might have this method take the `target` directly instead of the event since it's only checking for `event.target` inside of it.

src/js/plugins/menu_toggle/plugin.menu_toggle.js
```diff
@@ -1,113 +1,194 @@
...
+      if (book) {
+        book.addEventListener('click', onBookClick.bind(null, br), true);
+      }
+    }
+
+    /**
+     * Install menu toggle
+     * attaches event handlers, sets up DOM on load
+     */
+    BookReader.prototype.install = function installMenuToggle(e) {
```
jbuckner: I might name this method more specifically.. `install` is kind of generic and makes me think it's installing the entire book reader, but it's just installing the menu toggle right?

BookReader/BookReader.js
iisa: no, value doesn't change at this point, I just wanted to be very explicit.

src/js/plugins/menu_toggle/plugin.menu_toggle.js
iisa: thanks for catching

src/js/plugins/menu_toggle/plugin.menu_toggle.js
iisa: good call 👍

src/js/plugins/menu_toggle/plugin.menu_toggle.js
iisa: yes, +1 to more explicit

BookReader/BookReader.js
iisa: ✅

src/js/plugins/menu_toggle/plugin.menu_toggle.js
```diff
@@ -1,113 +1,194 @@
...
+      if (book) {
+        book.addEventListener('click', onBookClick.bind(null, br), true);
+      }
+    }
+
+    /**
+     * Install menu toggle
+     * attaches event handlers, sets up DOM on load
+     */
+    BookReader.prototype.installMenuToggle = function installMenuToggle(e) {
```
jbuckner: Does this method need to be public on the `BookReader.prototype` or can it be "internal" like `registerClickHandlers` is?

BookReader/BookReader.js
```diff
@@ -3873,16 +3891,50 @@ BookReader.prototype.navigationIsVisible = function() {
     return tooltop == 0;
 };
 
+/**
+ * Main controller that toggles navigation.
+ * Defaults to SHOW the navigation chrome
+ */
+BookReader.prototype.toggleNavigation = function brToggleNavigation(hide) {
```
jbuckner: Since we have the two public methods `hideNavigation` and `showNavigation`, does this method need to be public?

Also, this method name is kind of confusing when it's called with the intention of showing the menu. ie `br.toggleNavigation()` makes it sound like it's going to either show or hide the navigation (to the opposite of its current state) whereas that is going to cause the nav to show.

BookReader/BookReader.js
```diff
@@ -3873,16 +3891,50 @@ BookReader.prototype.navigationIsVisible = function() {
...
+    var animationLength = this.constNavAnimationDuration;
+    var animationType = 'linear';
+    var resizePageContainer = function resizePageContainer () {
+        /* main page container fills whole container */
+        if (this.constMode2up !== this.mode) {
+            var fillContainerWithBook = true;
```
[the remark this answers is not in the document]
iisa: thanks for catching

src/js/plugins/menu_toggle/plugin.menu_toggle.js
iisa: I was patterning it with the other plugin code, where the initializer is bound to the object. 
Should we start changing that behavior?

BookReader/BookReader.js
iisa: Seems like all core functionality is currently bound to the `BookReader` prototype object.  I'd like to keep to the convention set. Also, since it is based on its instance, it makes sense for it to be part of the prototypal core, so we can call `this.toggleNavigation`

I do feel it can be better named. Will update name.

src/js/plugins/menu_toggle/plugin.menu_toggle.js
jbuckner: Ah hmm.. I guess in my mind these types of methods don't really need to be used by the consumer and attached to the "public/prototype" interface. They're typically used internally by the initializer so I tend to not expose anything publicly if they don't need to access it.

iisa: ya, for sure. let me update.

jbuckner:
Up to 38% CodeCov! 🔥 👏

PR #172, opened 2019-10-28
Zoomed In Page Click

iisa:
Problem:
When zoomed in, user can not click on leaves of book to change the page.

Solution:
- Set the click handlers that change the page, on zoom.
- remove the `move` pointer
- deprecate certain core functions that are used around cursor changes.

PR #174, opened 2019-10-30
Plugin, search: hide search results pins when menu toggles

iisa:
Add custom event listener to new ambient event `BookReader:navToggled`
where the pins hide when the menu hides

Please note that this compounds on the latest menu toggle #171 & zoom click #172 fixes

https://webarchive.jira.com/browse/WEBDEV-2736

Evidence:
br-search-pins-toggle-with-menu

### DIFF
--- a/BookReader/plugins/plugin.search.js
+++ b/BookReader/plugins/plugin.search.js
@@ -41,6 +41,12 @@ BookReader.prototype.init = (function (super_) {
                 goToFirstResult: true
             });
         }
+
+        var br = this;
+        $(document).on('BookReader:navToggled', function() {
+            var nextVisibleState = br.navigationIsVisible() ? 'visible' : 'hidden';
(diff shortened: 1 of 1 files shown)


BookReader/plugins/plugin.search.js
```diff
@@ -41,6 +41,12 @@ BookReader.prototype.init = (function (super_) {
                 goToFirstResult: true
             });
         }
+
+        var br = this;
+        $(document).on('BookReader:navToggled', function() {
+            var nextVisibleState = br.navigationIsVisible() ? 'visible' : 'hidden';
```
bfalling: QUESTION: Wait, wouldn't this be the opposite? If this is a toggle and the navigation is currently visible, then wouldn't the next visible state be "hidden"?

iisa: thanks for catching, naming has been updated.

PR #176, opened 2019-10-30
plugin, menu toggle: adding check for user drag

iisa:
event handlers for `mouseup` and `mousedown` to figure out if user is dragging
if so, do not toggle menu.

this function can be candidate for formal util if we find the need.

drag-handler
`https://www-isa.archive.org/details/enchantedland00deve/page/n3`


src/js/plugins/menu_toggle/plugin.menu_toggle.js
```diff
@@ -188,6 +192,35 @@
...
+    var holdOffOnToggle = false;
+    /**
+     * attaches mouseup & mousedown event handlers to assess if user is dragging
+     * sets `initialX`, `initialY`, and `holdOffOnToggle`
+     *
+     * `holdOffOnToggle` is used in fn `toggleRouter`
+     * to determine if menu toggle should happen
+     */
+    var registerDragHandlers = function registerDragHandlers() {
+      var background = document.querySelector('.BookReader') || {};
```
bfalling: QUESTION: If there's no `.BookReader` element and `background` becomes `{}`, what happens with `addEventListener()`? Anything bad?

src/js/plugins/menu_toggle/plugin.menu_toggle.js
```diff
@@ -188,6 +192,35 @@
       toggleRouter(br, e, atBookCenter);
     }
 
+    var initialX;
+    var initialY;
+    var holdOffOnToggle = false;
```
bfalling: SUGGESTION: Since `toggleRouter()` above relies on `holdOffOnToggle` being in its closure variables, you might consider moving the `holdOffOnToggle` declaration up above to make it explicit, rather than relying on the reader recognizing the variable hoist that happens from it being declared down here.

src/js/plugins/menu_toggle/plugin.menu_toggle.js
iisa: good question. main thing about the plugin is nothing instantiates unless bookreader is actually onscreen.
`|| {}` is a devving artifact. removed.
+ added early escapes

PR #180, opened 2019-11-04
2up Zoomed In - do not turn page while dragging

iisa:
core: add drag check in 2up mode to see if page should be turned
`setMouseHandlers2UP` now has drag check
if dragging, do not change page


iisa:
closing, no longer needed

PR #181, opened 2019-11-12
Updates to Drag Motion Detection

iisa:
In the core code, we detect desktop mouse movement and determine when a user drags.
In there, we have a notion of "swipe", which is essentially a shorter `drag`.

Current implementation muddies up next phase in standardizing book tap targets in 2up mode.

Solution, is to remove page flip when swiping.  This renders the notion of `swipe` obsolete.

This PR reconfigures the internal reference dictionary that we build from handling mouse events.  Updates include:
- renaming top level reference dictionary from `_swipe` => `_drag`
- renaming initializer function from `initSwipeData` => `initDragCheck`
- remove page flip connected to `swipe` notion

### DIFF
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -3793,8 +3793,6 @@ BookReader.prototype.initDragCheck = function(clientX, clientY) {
      * Based on the really quite awesome "Today's Guardian" at http://guardian.gyford.com/
      */
     this._drag = {
-        mightBeFlicking: false,
-        didFlick: false,
         mightBeDraggin: false,
         didDrag: false,
         startTime: (new Date).getTime(),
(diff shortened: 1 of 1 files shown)


BookReader/BookReader.js
```diff
@@ -3839,48 +3836,26 @@ BookReader.prototype.swipeMousedownHandler = function(event) {
 };
 
 BookReader.prototype.swipeMousemoveHandler = function(event) {
```
jbuckner: Want to update this method name to `dragMousemoveHandler`?

BookReader/BookReader.js
```diff
@@ -3889,14 +3864,13 @@ BookReader.prototype.swipeMousemoveHandler = function(event) {
 };
 
 BookReader.prototype.swipeMouseupHandler = function(event) {
```
jbuckner: Same here with the `swipe`/`drag` naming.

iisa: thanks for tipping me over that fence! :done:

iisa:
closing for now. will reopen if needed

PR #185, opened 2019-12-02
ReadAloud Jump back

### LINKED ISSUE #183: ReadAloud option to jump back
Ability to jump back

### Summary of Requirements
- [ ] `AbstractTTSEngine.jumpBack` method
- [ ] Jump back icon displayed in read aloud toolbar
- [ ] Replace `AsyncStream` with not streams, since streams don't easily support backward navigation
- [ ] Should work for both TTS Engines

### Relationships
- Part of #94

cdrini:
Closes #183 

readaloud jump back

### Summary of Requirements
- [X] `AbstractTTSEngine.jumpBack` method
- [X] Jump back icon displayed in read aloud toolbar
- [X] Replace `AsyncStream` with not streams, since streams don't easily support backward navigation
- [X] Should work for both Engines

### Notes
- Added regenerator runtime so now we can use `async/await` in the test! It will also _automatically_ be added to the bundles if those keywords are used in the src JS.
- Replaced `AsyncStream` with `PageChunkIterator`; this is a more tradition iterator, keeping track of/incrementing page number/chunk index number. Less clean and more code, but allows for simpler backward navigation.
- `PageChunk` was getting a little bloated, so I pulled it out into its own class/file

### Open Questions/ToDos
- [X] Double check there hasn't been a substantial increase in file size are result of webpack config changes
  - archive_analytics: 9.76 -> 9.82
  - menu_toggle: 10.6 -> 10.7
  - tts: 89 -> 90.3
- [x] Fix up the failing browser tests
  - some of the "rare"-er bugs are still present (like spamming), but I think that's ok

### Testing
image


src/js/plugins/tts/PageChunkIterator.js
```diff
@@ -0,0 +1,153 @@
...
+     * Sends the cursor back 1
+     * @return {Promise}
+     **/
+    decrement() {
+        return this._cursorLock = this._cursorLock
+        .then(() => this._decrementUncontrolled());
+    }
+
+    /**
+     * Gets without ensuring synchronization
```
iisa: i am unsure what this means...

src/js/plugins/tts/PageChunkIterator.js
```diff
@@ -0,0 +1,153 @@
...
+            return chunks[this._cursor.chunk++];
+        });
+    }
+
+    /**
+     * Decrements without ensuring synchronization
+     * @return {Promise}
+     */
+    _decrementUncontrolled() {
+        if (this._cursor.page == 0 && this._cursor.chunk == 0) {
```
iisa: suggestion: use `===`
Also, can you use falsey values to make these comparator blocks work?

src/js/plugins/tts/PageChunkIterator.js
```diff
@@ -0,0 +1,153 @@
...
+            return this._fetchPageChunks(this._cursor.page);
+        } else if (this._cursor.page > 0 && this._cursor.chunk == 0) {
+            this._cursor.page--;
```
iisa: suggestion: space between `--`

iisa: taking a look at this whole block, there may be opportunity to refactor this succinctly as
`return this._fetchPageChunks(this._cursor.page);` is used, with decrement & callback as optional behavior.

src/js/plugins/tts/utils.js
```diff
@@ -29,7 +29,7 @@ export function approximateWordCount(text) {
 /**
  * Waits the provided number of ms and then resolves with a promise
  * @param {number} ms
- * @return {PromiseLike}
+ * @return {Promise}
```
iisa: 🗝

tests/plugins/tts/PageChunkIterator.test.js
```diff
@@ -0,0 +1,180 @@
+import sinon from 'sinon';
```
iisa: 👏 awesome test suite!

tests/plugins/tts/WebTTSEngine.test.js
```diff
@@ -32,24 +32,22 @@ describe('WebTTSSound', () => {
             expect(sound.rate).toBe(2);
         });
 
-        test('reloading does not resolve the original promise', () => {
+        test('reloading does not resolve the original promise', async () => {
```
iisa: really digging the use of async/await here 👏

BookReader/images/icon_skip-back.svg
```diff
@@ -0,0 +1,13 @@
...
+    <desc>Created with Sketch.</desc>
+    <g id="icon/replay" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="Replay-button">
+            <polyline id="Path" stroke="#333" stroke-width="2"
+                points="14.4444444 16.6666667 20 16.6666667 20 3.33333333 5.55555556 3.33333333" />
+            <polygon id="Path-2" fill="#333" points="5.55555556 0 5.55555556 6.66666667 1.11111111 3.33333333" /><text
+                id="10" font-family="HelveticaNeue, Helvetica Neue" font-size="10" font-weight="normal" fill="#333">
+            </text></g>
+    </g>
+</svg>
```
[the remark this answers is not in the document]
cdrini: ✅

src/js/plugins/tts/AbstractTTSEngine.js
```diff
@@ -91,28 +79,20 @@ export default class AbstractTTSEngine {
...
-        .map(chunk => {
-            this.opts.onLoadingStart();
-            chunk.sound = this.createSound(chunk);
-            chunk.sound.rate = this.playbackRate;
-            chunk.sound.voice = this.voice;
-            chunk.sound.load(() => this.opts.onLoadingComplete());
-            return chunk;
+        this._chunkIterator = new PageChunkIterator(numLeafs, leafIndex, {
+            server: this.opts.server,
+            bookPath: this.opts.bookPath,
```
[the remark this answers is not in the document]
cdrini: Discussed; Going to opt-out since `pageBufferSize` is also there :+1:

src/js/plugins/tts/AbstractTTSEngine.js
```diff
@@ -91,28 +79,20 @@ export default class AbstractTTSEngine {
...
+            pageBufferSize: 5,
```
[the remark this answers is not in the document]
cdrini: Defined on the `PageChunkIterator` options (p.s. if you're using vs code, hovering shows it :) )

image

src/js/plugins/tts/PageChunk.js
```diff
@@ -0,0 +1,94 @@
+export default class PageChunk {
```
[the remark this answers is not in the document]
cdrini: ✅

src/js/plugins/tts/PageChunk.js
```diff
@@ -0,0 +1,94 @@
...
+                error: rej,
+            })
+            .then(chunks => {
+                res(PageChunk._fromTextWrapperResponse(leafIndex, chunks));
+            });
+        });
+    }
+
+    /**
+     * @param {number} leafIndex
```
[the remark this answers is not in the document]
cdrini: Thanks for keeping me honest :)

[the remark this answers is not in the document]
cdrini: ✅

src/js/plugins/tts/PageChunk.js
```diff
@@ -0,0 +1,94 @@
...
+            // And the top to be the same height
+            firstRect[3] = bottom - secondHeight;
+        }
+
+        return rects;
+    }
+}
+
+/**
+ * @typedef {[number, number, number, number]} DJVURect
```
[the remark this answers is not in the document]
cdrini: I know right! Now instead of being a cryptic array, anywhere you see `DJVURect` you can see _exactly_ what these numbers mean :)

src/js/plugins/tts/PageChunkIterator.js
cdrini: Expanded; it's not pretty 😅

src/js/plugins/tts/PageChunkIterator.js
cdrini: I think I'd rather opt-out of this one; since these are explicitly numbers, I find it easier to read with `0`. Also I'd have to change `==` throughout a lot of places in this file if I were to follow that convention :P

src/js/plugins/tts/PageChunkIterator.js
```diff
@@ -0,0 +1,153 @@
...
+     */
+    _fetchPageChunksDirect(index) {
+        return PageChunk.fetch(this.opts.server, this.opts.bookPath, index);
+    }
+}
+
+PageChunkIterator.AT_END = "__PageChunkIterator.AT_END__";
+
+/** @type {PageChunkIteratorOptions} */
+const DEFAULT_OPTS = {
```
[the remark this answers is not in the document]
cdrini: In terms of ordering, this is more of a "helper" to this class, so I find having the file's main class at the top a little easier to read.

tests/plugins/tts/WebTTSEngine.test.js
cdrini: This code is kind of awkward, but it's slightly _less_ awkward because of async/await :P

src/js/plugins/tts/PageChunkIterator.js
cdrini: I'm not sure I follow the refactor; could you expand?

src/js/plugins/tts/AbstractTTSEngine.js
```diff
@@ -91,28 +79,20 @@ export default class AbstractTTSEngine {
...
         });
 
         this.step();
         this.events.trigger('start');
     }
 
     stop() {
-        this.activeSound.stop();
+        if (this.activeSound) this.activeSound.stop();
```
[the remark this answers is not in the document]
cdrini: They should be redundant, but I'd like them to happen anyways; there are a lot of async things happening, so it's very possible for the system to fall out of sync. Setting these variables ensures that the state variables reflect the true state of the system.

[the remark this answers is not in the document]
iisa: hmmm, to me. `true` state would only fire the event if it actually stopped.

>  there are a lot of async things happening, so it's very possible for the system to fall out of sync.

This statement makes me think that it should only fire if we stopped it.  But, also, I am here as a soundboard for you. :) carry on, if you think the system needs this echo at all times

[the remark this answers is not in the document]
cdrini: I actually agree with you 100% :P this is an example of slightly "paranoid" code (which is usually a bad thing). In this case, because the underlying WebSpeech API is _really_ unreliable/finicky is why this code is a little ham-fisted :P

src/js/plugins/tts/PageChunkIterator.js
cdrini: Fixed! I did something more promise-y though :P Also opting out of the `--` change; it's inconsistent with the code style of the repo.

iisa: look so good!

cdrini:
Thank you for the wonderful feedback and for making the code better :)

PR #186, opened 2019-12-04
Remove commented background images

maxocub:
Hi, I know it's not something that will be useful to a lot of people, but when BookReader is used in a Django app, the collectstatic command craches because of those commented backgroung images.


cdrini:
Thanks @maxocub ! Could you also delete the file `BookReader/images/left_edges.png`? It doesn't look like it's referenced anywhere else in the codebase.

maxocub:
Great, thank you @cdrini!

PR #188, opened 2019-12-11
add keyboard hooks for zoom in/out

cdrini:
Pressing -/= or numpad -/+ will now zoom out/in. We already support left/right/pgup/pgdown/home/end, so this seems fine.

### Testing
- [x] Tested works in 1up/2up/thumb
- [x] Testing typing these chars in search box doesn't do anything weird

### DIFF
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -711,6 +711,11 @@ BookReader.prototype.setupKeyListeners = function() {
     var KEY_RIGHT = 39;
     var KEY_DOWN = 40;
 
+    var KEY_MINUS = 173;
+    var KEY_NUMPAD_SUBTRACT = 109;
+    var KEY_EQUAL = 61;
+    var KEY_NUMPAD_ADD = 107;
+
     // We use document here instead of window to avoid a bug in jQuery on IE7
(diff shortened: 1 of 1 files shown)


BookReader/BookReader.js
```diff
@@ -752,6 +757,16 @@ BookReader.prototype.setupKeyListeners = function() {
...
                     break;
+                case KEY_MINUS:
+                case KEY_NUMPAD_SUBTRACT:
+                    e.preventDefault();
+                    self.zoom(-1);
+                    break;
+                case KEY_EQUAL:
+                case KEY_NUMPAD_ADD:
+                    e.preventDefault();
+                    self.zoom(+1);
```
rchrd2: The `+1` is not necessary here. Could just be `1`.

cdrini: Opting to leave this as is; I think the plus makes this easier to read. Seeing `zoom(+` makes me think that it's zooming in. Seeing `zoom(1)` doesn't necessarily do that; it looks like it might be resetting the zoom to 100%. (longer term we might create some wrapper function called `zoomIn` and `zoomOut` that hide the abstraction).

PR #189, opened 2019-12-16
re-adding click prevention on full screen

iisa:
- this will change when we work on page flipping at full screen


iisa:
@cdrini this will need to go in before you cut a new release

PR #190, opened 2019-12-17
Make ReadAloud ignore hyphens on line-breaks

### LINKED ISSUE #179: Read Aloud breaks up hyphenated words
First, thanks for such an awesome online library. I use Text-to-Speech extensively when reading books from the archive, it helps me concentrate better when reading a book and reduces my eye strain a lot.

**Description**
On digitized books, the hyphenated text has been broken into two words when speaking. For example, the word **negative** has been broken into 2 lines:

An excerpt from the book:

> We can also transform any we consider nega-
> tive

When read aloud, the word "negative" is pronounced as "nega-tive". The "tive" is pronounced like "five" which is incorrect.

It happens on this book a lot (I am loaning it right now, but might also happen with other books). https://archive.org/details/satirmodelfamily00sati

**Expectation**
I expected that hyphenated words are pronounced correctly.

**Context**
macOS Catalina (10.15)
Voice: Samantha
Google Chrome 78.0.3904.87

**Proposal & Constraints**
Automatically delete `- ` (a hyphen followed by a space) when uttering speak. I solved this problem on my side by running this userscript to monkeypatch the browser’s speech synthesis engine:

```js
speechSynthesis.speak = (speak => function(utterance) {
  utterance.text = utterance.text.replace(/- /g, '')
  return speak.call(this, utterance)
})(speechSynthesis.speak)
```

This should not harm the cases where hyphens are expected. For example, the word **life-preserving** has been wrapped into `life- preserving`. Removing the hyphens result in `lifepreserving`, but the TTS engine can pronounce it fine.

**Stakeholders**

> @-tag any stakeholders of this bug

cdrini:
Fixes #179 ; When hyphens appear at the end of a line, they are in the OCR as `citi- zens`. This causes the TTS to stutter. Remove any hyphens followed by whitespace, so that the TTS voice does not stutter.

## Testing
- [x] Unit tests
- [x] Test it does not stutter on this page http://localhost:8000/BookReaderDemo/demo-ia-plato.html#page/30/mode/2up


src/js/plugins/tts/PageChunk.js
```diff
@@ -90,6 +91,15 @@ export default class PageChunk {
...
         return rects;
     }
+
+    /**
+     * Remove "dangling" hyphens from read aloud text to avoid TTS stuttering
+     * @param {string} text 
+     * @return {string}
+     */
+    static _removeDanglingHyphens(text) {
+        return text.replace(/-\s+/g, '');
```
iisa: i ran this through regexr, it looks like it will only split out the last hyphen, even though there can/may be multiple hyphens

https://regexr.com/4qsgv

cdrini: As discussed, leaving as is; if we notice it occurring in the wild, we can fix it :+1:

PR #197, opened 2020-01-17
Added demo page for search plugin

nsharma123:
This will allow you to see search plugin functionality here: You can visit http://localhost:8000/BookReaderDemo/demo-plugin-search.html to see search plugin working.

### DIFF
--- a/BookReaderDemo/index.html
+++ b/BookReaderDemo/index.html
@@ -19,6 +19,7 @@
       <li><a href="demo-iiif.html">IIIF</a></li>
       <li><a href="demo-autoplay.html">Autoplay (kiosk mode)</a></li>
       <li><a href="demo-plugin-menu-toggle.html">Plugin: Full screen menu toggle</a></li>
+      <li><a href="demo-plugin-search.html">Plugin: Search field</a></li>
     </ul>
   </body>
 </html>
--- a/BookReaderDemo/demo-plugin-search.html
+++ b/BookReaderDemo/demo-plugin-search.html
@@ -1,0 +1,39 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>bookreader search plugin demo</title>
+
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
+    <meta name="apple-mobile-web-app-capable" content="yes">
+
+    <!-- JS dependencies -->
+    <script src="../BookReader/jquery-1.10.1.js"></script>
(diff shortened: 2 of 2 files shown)


BookReaderDemo/index.html
```diff
@@ -19,6 +19,7 @@
       <li><a href="demo-iiif.html">IIIF</a></li>
       <li><a href="demo-autoplay.html">Autoplay (kiosk mode)</a></li>
       <li><a href="demo-plugin-menu-toggle.html">Plugin: Full screen menu toggle</a></li>
+      <li><a href="demo-plugin-search.html">Plugin: Search field</a></li>
```
cdrini: As far as I can tell, the search plugin only works for IA books, so the "simple" demo won't work (it throws an error when you try to search). Let's not add a new html file, and link to the "From Internet Archive" page since it includes the search plugin:

```suggestion
      <li><a href="demo-ia-plato.html">Plugin: Search field (only works for Internet Archive books)</a></li>
```

nsharma123: Right, It return error when you search but appending the search term in URL when you click on search icon. something like `/search/word-you-want-to-search-in-book`.
So I guess the search is working at bookreader repository when you pull code and run through python server. BUT at archive.org end, the search term is not appending in existing URL when you search anything in a book. Do you think something is wrong at petabox or book-reader repo?

cdrini: It doesn't work from `/details`, but seems to work in `/stream` (e.g. https://archive.org/stream/tinypicturebook00chariala).

This is the code in bookreader that deals with changing the url: https://github.com/internetarchive/bookreader/blob/master/BookReader/plugins/plugin.url.js

In that file, it looks like there's an option for what properties get reflected in the url:
https://github.com/internetarchive/bookreader/blob/8b7e6ac85cdbb3590a47ee6d28b8269838638e18/BookReader/plugins/plugin.url.js#L22

Checking on archive.org (run `br.options.urlTrackedParams` in the console), we get:
```js
["page"]
```

This suggests that in the petabox side of code, someone decided that the details page should only reflect the page (for some reason). I'd recommend seraching petabox for `urlTrackedParams` to see where it's set, and see if there are any comments there, or check the git blame to see if there's any useful history.

Hope that helps!

nsharma123: Yupe, that is make sense as we are passing some required things (`BR objects`, `urlTrackedParams` etc..) form petabax codebase in-order to proceed with search functionality.
 
So, I am closing this PR. Thanks @cdrini!

PR #198, opened 2020-01-30
BookReader: zoomed-in 2 page view, clicking page does not change the page

nsharma123:
This PR is based on https://webarchive.jira.com/browse/WEBDEV-2737

- When you have zoomed-in 2 page view, should be be able to drag image to see around.
- When you have zoomed-in 2 page view, you should be able to click on book image to change page.

### DIFF
--- a/BookReader/BookReader.js
+++ b/BookReader/BookReader.js
@@ -2798,27 +2798,31 @@ BookReader.prototype.flipRightToLeft = function(newIndexL, newIndexR) {
 BookReader.prototype.setMouseHandlers2UP = function() {
     var self = this;
     ['L', 'R'].forEach(function(dir) {
-      var currentIndexEl = self.prefetchedImgs[self.twoPage['currentIndex' + dir]];
-      self.setClickHandler2UP(currentIndexEl,
-          { self: self },
-          function(e) {
-              if (e.which == 3) {
-                  // right click
-                  return !e.data.self.protected;
-              }
+        var currentIndexEl = self.prefetchedImgs[self.twoPage['currentIndex' + dir]];
+        self.setClickHandler2UP(currentIndexEl,
+            { self: self },
+            function(e) {
+                if (e.which == 3) {
+                    // right click
+                    return !e.data.self.protected;
+                }
 
-              // Changes per WEBDEV-2737
-              // BookReader: zoomed-in 2 page view, clicking page should change the page
-              $(currentIndexEl)
-              .mousemove(function() {
-                  e.preventDefault();
-              })
-              .mouseup(function() {
-                  e.data.self.trigger(BookReader.eventNames.stop);
-                  e.data.self.left();
-              });
-          }
-      );
+                // Changes per WEBDEV-2737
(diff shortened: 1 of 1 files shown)


BookReader/BookReader.js
```diff
@@ -2796,7 +2796,8 @@ BookReader.prototype.flipRightToLeft = function(newIndexL, newIndexR) {
 };
 
 BookReader.prototype.setMouseHandlers2UP = function() {
-    this.setClickHandler2UP( this.prefetchedImgs[this.twoPage.currentIndexL],
+    var currentIndexL = this.prefetchedImgs[this.twoPage.currentIndexL];
```
shaneriley: I feel like the changes made in this function could be rewritten to avoid a lot of duplicate code. I don't know if I can easily describe the approach, but I can add a commit on this branch demonstrating it.

BookReader/BookReader.js
```diff
@@ -2798,27 +2798,31 @@ BookReader.prototype.flipRightToLeft = function(newIndexL, newIndexR) {
...
-      );
+                // Changes per WEBDEV-2737
+                // BookReader: zoomed-in 2 page view, clicking page should change the page
+                $(currentIndexEl)
+                .mousemove(function() {
+                    e.preventDefault();
+                })
+                .mouseup(function() {
+                    e.data.self.trigger(BookReader.eventNames.stop);
+                    if (dir === 'L') {
```
shaneriley: That's odd. I must not have copied that change over to the committed code. It's in the file on the testing vm-home instance. What I did was a one-liner like
`e.data.self[dir === 'L' ? 'left' : 'right']()`

nsharma123: Changed per one-liner suggestion. Thanks @shaneriley!

BookReader/BookReader.js
```diff
@@ -2796,40 +2796,30 @@ BookReader.prototype.flipRightToLeft = function(newIndexL, newIndexR) {
...
 BookReader.prototype.setMouseHandlers2UP = function() {
-    this.setClickHandler2UP( this.prefetchedImgs[this.twoPage.currentIndexL],
-        { self: this },
-        function(e) {
-            if (e.which == 3) {
-                // right click
-                if (e.data.self.protected) {
-                    return false;
+    var self = this;
+    ['L', 'R'].forEach(function(dir) {
```
jbuckner: Curious, instead of binding these two individual functions separately like this, could you abstract the handler and pass in a direction with the event? ie:
```
var handler = function(e, direction) { 
  if (e.which == 3) {
    ...
  }
  ...
};

this.setClickHandler2UP(this.prefetchedImgs[self.twoPage['currentIndexL']]), { self: self, direction: 'L' }, handler);
this.setClickHandler2UP(this.prefetchedImgs[self.twoPage['currentIndexR']]), { self: self, direction: 'R' }, handler);
```

cdrini: I'm not sure if that is specifically the best solution, but I think we should avoid concatenating properties (e.g. `self.twoPage['currentIndex' + dir]`); that makes it a lot harder to grep to find where a given property is used.

PR #106, opened 2019-06-07
Replaces hamburger image icon with CSS

shaneriley:
Fixes #100 

With PNG hamburger:

Screen Shot 2019-06-07 at 3 29 30 PM

With CSS hamburger:

Screen Shot 2019-06-07 at 2 31 30 PM

PR #200, opened 2020-02-12
Switch back to using PNG for hamburger icon

### LINKED ISSUE #144: Hamburger icon only clickable on lines
**Description**
You can only activate the hamburger icon by clicking on the lines; if you click the whitespace between or near the lines, nothing happens.

bookreader hamburger icon

**Expectation**
Clicking anywhere on the containing box should trigger the action.

**Context**
- Windows 10, Firefox 69

**Proposal & Constraints**
- The icon should be in an `<a>` tag with padding; or a button.

**Stakeholders**

@mekarpeles

cdrini:
Closes #144 

Reverts internetarchive/bookreader#106 ; it was causing an odd issue where the button would only work if clicked on the hamburger buns or patty :P Choosing a revert here since it seems like the least likely change to cause more bugs (the fact that it requires very little work, is also appealing :P ). We can find a longer-term solution in the future.

✅ Tested: FF73; Ch 80 on Win10
Steps:
1. Visit http://localhost:8000/BookReaderDemo/demo-fullscreen.html#mode/2up
2. Resize until in mobile view
3. Click hamburger icon

### DIFF
--- a/src/css/_MobileNav.scss
+++ b/src/css/_MobileNav.scss
@@ -45,32 +45,19 @@ html.mm-background .BookReader {
 }
 
 .BRmobileHamburger {
-  display: block;
-  position: absolute;
-  top: 0;
-  left: 17px;
-  width: 26px;
-  height: 3px;
-  margin: 18px 0;
-  outline: none;
+	background: center center no-repeat transparent;
+  background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABsAAAATCAYAAABhh3Y4AAAAAXNSR0IArs4c6QAAAFZJREFUSA1jrKmpafv//38xELMx0AgwMjL+AuJeFqAlubS0COR+qPm5TED2FJDNNPIU2FiozybT0o5Rs4dBCDCO5jNKYnE0n1ESeiNI72g+oyiyYfkMAKCpRQemMsj0AAAAAElFTkSuQmCC');
+  /* Should use SVG (below). See https://github.com/internetarchive/bookreader/issues/100 */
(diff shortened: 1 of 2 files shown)


src/css/_MobileNav.scss
```diff
@@ -45,32 +45,19 @@ html.mm-background .BookReader {
...
-  top: 0;
-  left: 17px;
-  width: 26px;
-  height: 3px;
-  margin: 18px 0;
-  outline: none;
+	background: center center no-repeat transparent;
+  background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABsAAAATCAYAAABhh3Y4AAAAAXNSR0IArs4c6QAAAFZJREFUSA1jrKmpafv//38xELMx0AgwMjL+AuJeFqAlubS0COR+qPm5TED2FJDNNPIU2FiozybT0o5Rs4dBCDCO5jNKYnE0n1ESeiNI72g+oyiyYfkMAKCpRQemMsj0AAAAAElFTkSuQmCC');
+  /* Should use SVG (below). See https://github.com/internetarchive/bookreader/issues/100 */
+  /*background: url('data:image/svg+xml;utf8,<svg width="15px" height="12px" viewBox="0 0 15 12" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"> <g id="hamburger-svg" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" stroke-linecap="square"> <path d="M0,1 L14,1" id="Line" stroke="#000000"></path> <path d="M2.39808173e-14,6 L14,6" id="Line-Copy" stroke="#000000"></path> <path d="M1.15463195e-14,11 L14,11" id="Line-Copy-2" stroke="#000000"></path> </g></svg>') center / 50% 50% no-repeat;*/
```
jbuckner: Did you mean to keep this comment in here?

cdrini: Yep; I just did a full revert. (I also think that comment is useful, cause I think it should be re-instated at some point :P )

shaneriley: We could dig through the commit history to locate it if you're comfortable with that.

cdrini: Haha, I was the one to originally add that comment :P The inline SVG wasn't properly encoded, which was causing it to not render. I fixed it by using the PNG, but left the SVG in case someone wanted to fix/re-instate it at some point. https://github.com/internetarchive/bookreader/pull/99/files

PR #201, opened 2020-02-13
Adds TestCafe dev dependency, instructions, example test

shaneriley:
Please see the README in tests/e2e for instructions.


cdrini:
Oh, it looks like `jest` is trying to run those tests :P Let me see if I can find a quick fix...

PR #208, opened 2020-02-28
Run testcafe in Travis

cdrini:
Followed instructions on https://devexpress.github.io/testcafe/documentation/continuous-integration/travis.html

- Added an `npm run serve` command that launches the app on localhost:8000
- `npm run testcafe:e2e` now starts the app itself; might conflict?

I'm a little nervous they might be flaky; but lets see.


.testcaferc.json
```diff
@@ -1,4 +1,5 @@
 {
+  "appCommand": "npm run serve",
```
iisa: Q: how is this command run? I did not see it explicitly called in the `test:e2e` script

cdrini: `test:e2e` calls `testcafe`, which looks at this config file (and runs this command) 🕸

iisa: ah ok, thanks for clarifying. i was looking at the command line docs, but finally found this `https://devexpress.github.io/testcafe/documentation/using-testcafe/configuration-file.html#appcommand`

iisa:
🍰

PR #209, opened 2020-02-28
Update webpack to dynamically compile files from a list

iisa:
Takes a file from the file list and outputs compiled file in the `BookReader` directory.
It keeps the same flattened directory structure as the file mapper it replaces.

Note: I purposely kept the file list, and didn't fully automate because I wanted to make the entrypoints explicit.

This is part of internal ticket: https://webarchive.jira.com/browse/WEBDEV-3183

Regex example: https://regexr.com/4v9m4

### How to test:
- pull down branch
- at top of BookReader directory, run the following:
  - `npm install` to update your modules
  - `npx webpack` to build the files
    - if you didn't delete any built files or update an es6 file listed, no files should have updated
    - if you update a file and run `npx webpack` - a change should be picked up by git for you to commit/review

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,10 +1,21 @@
 const path = require('path');
 
-module.exports = [
-    buildJsFromTo({from: 'plugins/plugin.archive_analytics.js', to: 'BookReader/plugins/plugin.archive_analytics.js'}),
-    buildJsFromTo({from: 'plugins/menu_toggle/plugin.menu_toggle.js', to: 'BookReader/plugins/plugin.menu_toggle.js'}),
-    buildJsFromTo({from: 'plugins/tts/plugin.tts.js', to: 'BookReader/plugins/plugin.tts.js'}),
+const listOfFiles = [
+    'plugins/plugin.archive_analytics.js',
+    'plugins/menu_toggle/plugin.menu_toggle.js',
+    'plugins/tts/plugin.tts.js',
 ];
+
+const compileFiles = async () => {
+    const nestedDirs = new RegExp('/(.*)/');
+    return await listOfFiles.map((filePath) => {
+        const flattenedFilePath = filePath.replace(nestedDirs, '/');
+        return buildJsFromTo({
+            from: filePath,
+            to: `BookReader/${flattenedFilePath}`
+        });
+    });
+};
 
 /**
  * @param {Object} opts
@@ -32,3 +43,5 @@ function buildJsFromTo({ from: srcEntryFile, to: outputFile }) {
         devtool: 'source-map'
     };
 }
+
+module.exports = compileFiles();
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -2,6 +2,9 @@ const path = require('path');
 
 module.exports = buildJSFiles();
 
+/**
+ * Applies bundling to the listed files.
+ */
 function buildJSFiles () {
     const listOfFiles = [
         'plugins/plugin.archive_analytics.js',
@@ -19,6 +22,8 @@ function buildJSFiles () {
 }
 
 /**
+ * Applies webpack config to files that it is bundling.
+ *
  * @param {Object} opts
  * @param {String} opts.from
  * @param {String} opts.to
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,21 +1,22 @@
 const path = require('path');
 
-const listOfFiles = [
-    'plugins/plugin.archive_analytics.js',
-    'plugins/menu_toggle/plugin.menu_toggle.js',
-    'plugins/tts/plugin.tts.js',
-];
+module.exports = buildJSFiles();
 
-const compileFiles = async () => {
+function buildJSFiles () {
+    const listOfFiles = [
+        'plugins/plugin.archive_analytics.js',
+        'plugins/menu_toggle/plugin.menu_toggle.js',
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -1,11 +1,22 @@
...
-    buildJsFromTo({from: 'plugins/plugin.archive_analytics.js', to: 'BookReader/plugins/plugin.archive_analytics.js'}),
-    buildJsFromTo({from: 'plugins/menu_toggle/plugin.menu_toggle.js', to: 'BookReader/plugins/plugin.menu_toggle.js'}),
-    buildJsFromTo({from: 'plugins/tts/plugin.tts.js', to: 'BookReader/plugins/plugin.tts.js'}),
+const listOfFiles = [
+    'plugins/plugin.archive_analytics.js',
+    'plugins/menu_toggle/plugin.menu_toggle.js',
+    'plugins/tts/plugin.tts.js',
 ];
 
+const compileFiles = async () => {
```
cdrini: Let's not make this rely on closure to get access to the files it's compiling; makes it a little easier to understand what it's doing when looking at it in isolation :)

Oh, also I don't think think this needs to be `async`; no promises

Also, let's call this `buildJsFiles`, so that it's consistent with the 

```suggestion
/**
 * @param {String[]} files
 */
function buildJsFiles(files) {
```

webpack.config.js
```diff
@@ -1,11 +1,22 @@
...
 
-module.exports = [
-    buildJsFromTo({from: 'plugins/plugin.archive_analytics.js', to: 'BookReader/plugins/plugin.archive_analytics.js'}),
-    buildJsFromTo({from: 'plugins/menu_toggle/plugin.menu_toggle.js', to: 'BookReader/plugins/plugin.menu_toggle.js'}),
-    buildJsFromTo({from: 'plugins/tts/plugin.tts.js', to: 'BookReader/plugins/plugin.tts.js'}),
+const listOfFiles = [
+    'plugins/plugin.archive_analytics.js',
+    'plugins/menu_toggle/plugin.menu_toggle.js',
+    'plugins/tts/plugin.tts.js',
 ];
```
cdrini: Let's keep the exports at the top here; otherwise it's a lot of jumping around to see what's happening or where the important stuff is happening.

Also, saving the list of files in an array doesn't really add anything in this case; it's useful if we were going to be re-using it, or if the variable name helped make the code clearer, but placing the function here makes this self-documenting, since the name of the function explains what the list is.

```suggestion
module.exports = buildJsFiles([
    'plugins/plugin.archive_analytics.js',
    'plugins/menu_toggle/plugin.menu_toggle.js',
    'plugins/tts/plugin.tts.js',
]);
```

webpack.config.js
```diff
@@ -1,11 +1,22 @@
...
 
+const compileFiles = async () => {
+    const nestedDirs = new RegExp('/(.*)/');
+    return await listOfFiles.map((filePath) => {
+        const flattenedFilePath = filePath.replace(nestedDirs, '/');
```
cdrini: ++; this is a really perfect example of a good variable name. Although we could inline this since it's only used once, I think that would result in the code being harder to understand.

webpack.config.js
```diff
@@ -1,11 +1,22 @@
...
+        return buildJsFromTo({
+            from: filePath,
+            to: `BookReader/${flattenedFilePath}`
+        });
+    });
```
cdrini: - Remove await
- Inline the regex; the variable name self-documents what the regex is.

```suggestion
    return files.map(filePath => {
        const flattenedFilePath = filePath.replace(/\/(.*)\//, '/');
        return buildJsFromTo({from: filePath, to: path.join('BookReader', flattenedFilePath)});
    });
```

webpack.config.js
```diff
@@ -32,3 +43,5 @@ function buildJsFromTo({ from: srcEntryFile, to: outputFile }) {
         devtool: 'source-map'
     };
 }
+
+module.exports = compileFiles();
```
cdrini:
```suggestion
```

webpack.config.js
iisa: that's fine.  I am ok with changing the name of the top level function but I want to keep the webpack config separate from the files.

webpack.config.js
iisa: `the variable name self-documents what the regex is.` - yes it speaks to the output but I think it's important to declare what the regex itself is matching to, even if the regex is simple.

`filePath.replace(nestedDirs, '/');` reads better to me than `filePath.replace(/\/(.*)\//, '/');`

webpack.config.js
iisa: yes totally 👍  - I was using es6 arrows. will hoist ⬆️ to `function`

cdrini: Ahhh, gotcha; I forgot they don't hoist! I was wondering why that was :P

iisa: done

webpack.config.js
cdrini: :+1: Sounds good to me! Could we change its name to `nestedDirRegex`? Regexes tend to be difficult to name, because they always sound like something else (e.g. is it the name of the directory that's nested? Is it the file's directory?).

iisa: Yep. I can see that naming complexity. Done.

PR #212, opened 2020-03-03
WIP - Es6 plugin autoplay

iisa:


src/js/plugins/plugin.autoplay.js
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Plugin which adds an autoplay feature. Useful for kiosk situations.
+ */
+jQuery.extend(BookReader.defaultOptions, {
+  enableAutoPlayPlugin: true
+});
+
+/**
+ * @augments BookReader.setup
```
cdrini: I think `@override` is the best for these; `augments` looks like it's only for `class`es/inheritance. (ditto elsewhere)

```suggestion
 * @override
```

src/js/plugins/plugin.autoplay.js
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * @augments BookReader.init
+ */
+BookReader.prototype.init = (function(super_) {
+  return function (options) {
+    super_.call(this, options);
+
+    if (!this.options.enableAutoPlayPlugin) return;
+    this.bind(BookReader.eventNames.stop, (e, br) => br.autoStop());
```
cdrini: No longer need to pass `br` at all :) 

```suggestion
    this.bind(BookReader.eventNames.stop, () => this.autoStop());
```

src/js/plugins/plugin.autoplay.js
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * @augments BookReader.bindNavigationHandlers
+ */
+BookReader.prototype.bindNavigationHandlers = (function(super_) {
+  return function() {
+    super_.call(this);
+
+    if (!this.options.enableAutoPlayPlugin) return;
+
+    var self = this;
```
cdrini: No need anymore 🎉 

```suggestion
```

src/js/plugins/plugin.autoplay.js
```diff
@@ -0,0 +1,130 @@
...
+
+    // Note the mobile plugin attaches itself to body, so we need to select outside
+    var jIcons = this.$('.BRicon').add('.BRmobileMenu .BRicon');
```
cdrini: Oh, let's also use const/let; I think that should be an easy one to do as well
```suggestion
    const jIcons = this.$('.BRicon').add('.BRmobileMenu .BRicon');
```

src/js/plugins/plugin.autoplay.js
```diff
@@ -0,0 +1,130 @@
...
+
+    jIcons.filter('.play').click(function() {
+      self.autoToggle();
+      return false;
+    });
```
cdrini:
```suggestion
    jIcons.filter('.play').click(() => {
      this.autoToggle();
      return false;
    });
```

src/js/plugins/plugin.autoplay.js
```diff
@@ -0,0 +1,130 @@
...
+
+    jIcons.filter('.pause').click(function() {
+      self.autoToggle();
+      return false;
+    });
```
cdrini:
```suggestion
    jIcons.filter('.pause').click(() => {
      this.autoToggle();
      return false;
    });
```

src/js/plugins/plugin.autoplay.js
```diff
@@ -0,0 +1,130 @@
...
+    bComingFrom1up = true;
+    this.switchMode(this.constMode2up);
+  }
+
+  // Change to autofit if book is too large
+  if (this.reduce < this.twoPageGetAutofitReduce()) {
+    this.zoom2up('auto');
+  }
+
+  var self = this;
```
cdrini:
```suggestion
```

src/js/plugins/plugin.autoplay.js
```diff
@@ -0,0 +1,130 @@
...
+    this.$('.pause').show();
+    this.autoTimer = setInterval(function(){
+      if (self.animating) return;
+
+      if (Math.max(self.twoPage.currentIndexL, self.twoPage.currentIndexR) >= self.lastDisplayableIndex()) {
+        self.flipBackToIndex(1); // $$$ really what we want?
+      } else {
+        self.flipFwdToIndex();
+      }
+    }, this.flipDelay);
```
cdrini:
```suggestion
    this.autoTimer = setInterval(() => {
      if (this.animating) return;

      if (Math.max(this.twoPage.currentIndexL, this.twoPage.currentIndexR) >= this.lastDisplayableIndex()) {
        this.flipBackToIndex(1); // $$$ really what we want?
      } else {
        this.flipFwdToIndex();
      }
    }, this.flipDelay);
```

tests/plugins/plugin.autoplay.test.js
```diff
@@ -0,0 +1,46 @@
+/* global BookReader */
+require('../../BookReader/jquery-1.10.1.js');
```
cdrini: Can we use `import` here instead?

tests/plugins/plugin.autoplay.test.js
```diff
@@ -0,0 +1,46 @@
...
+    expect(br.auto).toEqual(false);
+  });
+  test('has added BR property: autoTimer', () => {
+    expect(br).toHaveProperty('autoTimer');
+    expect(br.autoTimer).toEqual(null);
+  });
+  test('has added BR property: flipDelay', () => {
+    expect(br).toHaveProperty('flipDelay');
+    expect(br.flipDelay).toBeTruthy();
+    expect(br.flipDelay).toBeGreaterThan(1);
```
cdrini: I don't think these will be too helpful in catching bugs; unit tests that test the functionality of the methods are probably better; e.g. "calling autoStop clears interval"; essentially you want to assert that functions fulfill their contract; so when certain pre-conditions are present, a test should assert that the post-conditons are correct.

webpack.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = buildJSFiles([
     'plugins/plugin.archive_analytics.js',
     'plugins/menu_toggle/plugin.menu_toggle.js',
     'plugins/tts/plugin.tts.js',
+    'plugins/plugin.autoplay.js'
```
cdrini: Also need to add this to `.gitattributes`

cdrini: Let's keep these sorted; that way we'll have fewer merge conflicts than if we always added to the end :P

tests/plugins/plugin.autoplay.test.js
iisa: done

src/js/plugins/plugin.autoplay.js
```diff
@@ -0,0 +1,127 @@
...
+BookReader.prototype.setup = (function(super_) {
+  return function (options) {
+    super_.call(this, options);
+
+    this.auto      = false;
+    this.autoTimer = null;
+    this.flipDelay = 5000;
+  };
+})(BookReader.prototype.setup);
+
```
cdrini: Oooooh! I think I found the transform that will make these look less awful!

This has the benefits of:

- We can export the `class`, so it can more easily be `import`ed in tests
- Get rid of that ugly function scoping and `super_`
- Get rid of awkward `BookReader.prototype` overrides
- Still a pretty mindless transform

```suggestion
export default class BookReaderWithAutoplayPlugin extends BookReader {
    /** @override */
    setup(options) {
        super.setup(options);

        this.auto = false;
        this.autoTimer = null;
        this.flipDelay = 5000;
    }
}

if (window && window.BookReader) {
    window.BookReader = BookReaderWithAutoplayPlugin;
}
```

Thoughts?

jbuckner: Interesting approach! So is the idea then for each plugin, we just keep replacing `window.BookReader` with the class that is getting exported, which should be the most up-to-date version of BookReader (in that it has all other plugins loaded in it)?

Do we need that `window.BookReader` check? I think we can set it regardless.. ie:
```
if (window) {
  window.BookReader = BookReaderWithAutoplayPlugin;
}
```

cdrini: Yeah! That's essentially what it's doing already using es5 syntax. I think you're right, the `window.BookReader` check is unnecessary; unless, do es6 modules imported with `type="module"` have access to `window`? That's the only case where it might matter.

jbuckner: So I just thought of something.. since `BookReaderWithAutoplayPlugin` is extending `BookReader`, I think setting `window.BookReader` to that is going to get you just the `BookReaderWithAutoplayPlugin` without any of the other plugins. You're basically overwriting it each time.

cdrini: It would be sort of recursive; since the plugins are loaded in order, each one overwrites the `BookReader` the previous one modified (which is also the way it currently works :P )

e.g.
```js
window.BookReader = class BookReader {}
window.BookReader = class BookReaderWithAutoPlay extends BookReader {}
window.BookReader = class BookReaderWithIframe extends BookReader {}
```

jbuckner: Is it extending `window.BookReader` though? I would expect it to be extending the one imported.. ie:
```
import BookReader from './bookreader.js';

class BookReaderWithAutoPlay extends BookReader {}
```

cdrini: In a browser environment it would be extending `window.BookReader`; in a node environment it would be extending the `import`. Not sure what happens in browser modules 🤔

jbuckner: Ohhh interesting!

So just thinking out loud (and not a blocker to this)... is there any way we could extend this idea if we wanted to test multiple plugins independent of `window.BookReader`?

cdrini: Hahaha, my thoughts exactly! I've just been trying to see if there's a way to convert a list of subclasses into an inheritance chain...

cdrini: But yeah, I don't think it's blocking. `jest` offers a `window` object, so importing will work as it does in the browser.

cdrini: Actually, just tested doing this for iframe and it requires _a lot_ of polyfills :/ The file went from 2KB min -> 23 Kb :/ We'd need to have some sort of breaking change to include a onetime polyfill otherwise this will be duplicated for every plugin. So let's not do this for now :(

src/js/plugins/plugin.autoplay.js
```diff
@@ -0,0 +1,127 @@
...
+    jIcons.filter('.pause').click(() => {
+      this.autoToggle();
+      return false;
+    });
+  };
+})(BookReader.prototype.bindNavigationHandlers);
+
+/**
+ * Starts autoplay mode
+ * @param {object} overrides - accepts numeric properties flipSpeed and flipDelay
```
cdrini: Jsdoc also supports:

```suggestion
 * @param {object} overrides
 * @param {number} overrides.flipSpeed
 * @param {number} overrides.flipDelay
```

tests/plugins/plugin.autoplay.test.js
```diff
@@ -0,0 +1,60 @@
...
+  });
+  test('autoplay does not start when BookReaderInitializes', () => {
+    br.autoToggle = jest.fn();
+    br.init();
+    expect(br.autoToggle).toHaveBeenCalledTimes(0);
+  });
+  test('autoplay will run without `flipSpeed` parameters', () => {
```
cdrini: Nice! These tests look awesome :)

src/js/plugins/plugin.autoplay.js
jbuckner: Oy! Bummer! Good idea, though!

PR #214, opened 2020-03-03
Es6 plugin iframe

cdrini:
https://webarchive.jira.com/browse/WEBDEV-3196

Testing:
Visit http://localhost:8000/BookReaderDemo/demo-embed.html locally, and ensure `window.addEventListener('message', e => console.log(e))` fires on page flip


nsharma123:
@cdrini Please create a JIRA in respect of of the PR related to es6 for bookreader plugins. Thanks!

src/js/plugins/plugin.iframe.js
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * @private
+ * Using window.postMessage() and event listeners, the plugin notifies the
+ * parent window when pages change, and the parent window can also
+ * explicitly request a page change by sending its own message.
+ *
+ * @param {BookReader} br
+ * @param {Window?} [parent]
+ */
+export function _attachEventListeners(br, parent=window.parent) {
```
jbuckner: How come you prefixed this with an underscore?

cdrini: I'm only exporting this because I want to test it :P It actually shouldn't be exported. I'm opting to use the underscore to better communicate that it's private. Not crazy about this, but I'm not sure if there's a better way.

nsharma123:
Lets make consistency indentation spaces consistency. sometimes we use 4 spaces and sometimes 2 spaces.

PR #222, opened 2020-03-04
[es6-bookreader-core] move BookReader.js to src/js

cdrini:
Part one: this just copies BookReader over, updates any references, and fixes indentation.

Note:
- Opted to skip the `let`/`const` in this first step; will do in future PRs


scripts/version.js
```diff
@@ -15,11 +15,6 @@ async function main() {
 
     // build
     execSync('npm run build', { stdio: "inherit" });
-    
-    // Update BookReader.version
-    const old_js = fs.readFileSync('BookReader/BookReader.js').toString();
-    const new_js = old_js.replace(/^BookReader\.version = .*/m, `BookReader.version = '${NEW_VERSION}';`);
-    fs.writeFileSync('BookReader/BookReader.js', new_js);
```
iisa: Q: why do we no longer need this?

cdrini: This was doing a find and replace of the old version in BookReader.js and replacing it with the new version :P Now that we have this behind a compiler, we can just `import` the version from `package.json` directly :)

cdrini: https://github.com/internetarchive/bookreader/blob/es6-bookreader-core/src/js/BookReader.js#L22

PR #223, opened 2020-03-04
[es6-bookreader-core] extract BookReader.util

cdrini:
Extract a chunk of BookReader.js: static util methods. Turns out most of the jest tests were for this anyways :P


src/js/BookReader.js
```diff
@@ -20,6 +20,7 @@ This file is part of BookReader.
 
 */
 import { version as VERSION } from '../../package.json';
+import * as utils from './BookReader/utils.js';
```
iisa: 👏

tests/BookReader/utils.test.js
```diff
@@ -0,0 +1,63 @@
+import sinon from 'sinon';
+import {
+    clamp,
+    cssPercentage,
+    debounce,
+    decodeURIComponentPlus,
+    encodeURIComponentPlus,
+    escapeHTML,
+} from '../../src/js/BookReader/utils.js';
```
iisa: Q: is this every util?

tests/BookReader/utils.test.js
```diff
@@ -0,0 +1,63 @@
...
+
+test('Encodes a URI component and converts emptyStr to +', () => {
+    expect(encodeURIComponentPlus("?x=test ")).toEqual("%3Fx%3Dtest+");
+    expect(encodeURIComponentPlus("ABC abc 123")).toEqual("ABC+abc+123");
+});
+
+let clock;
+
+beforeEach(() => {
+  clock = sinon.useFakeTimers();
```
iisa: cool

tests/BookReader/utils.test.js
cdrini: All but 1 I think :P

tests/BookReader/utils.test.js
cdrini: This was here before me, so can't take full credit :P

cdrini:
> I saw some string concatenations that were touched, but I am assuming that we will do that update for that in another round of fixes

Exactly; as I split things out, I'll es6-ify them :+1:

PR #225, opened 2020-03-05
ES6 Plugin: Mobile Nav

iisa:
Plugin now uses:
- `() => {}` where possible without breaking scopes & closures
- string templates for DOM
- tests

### To Test:
- pull down branch
- update your node modules, cmd: `npm install`
- run your local server, cmd: `npm run serve'
- go to page with a browser width lower than 800px: `http://127.0.0.1:8080/BookReaderDemo/demo-ia-olivertwist.html`
- Mobile menu should appear
- Then click Menu 🍔 at top right -> menu slides in from left
- click 'Settings' -> Settings menu slides in
- Click on every Option in Settings -> they all work as expected
- click 'About This Book' -> About menu slides in 
- Book Info is displayed, link in menu changes URL to relative path `~/details/<bookID>`
- click 'Share This Book' -> share menu slides in
- Click on every Option in Share -> the allowed share options pop up their info, note URL is a relative path.

Internal ticket: https://webarchive.jira.com/browse/WEBDEV-3201


src/js/plugins/plugin.mobile_nav.js
```diff
@@ -133,50 +134,74 @@ BookReader.prototype.buildToolbarElement = (function (super_) {
...
-      +"        Share This Book"
-      +"      </span>"
-      +"      <div class=\"BRmobileShare\"></div>"
-      +"    </li>"
-      +"  </ul>"
-      +"</nav>"
-  );
+  const settingsSection = `
+    <span>
+        <span class="DrawerIconWrapper">
```
cdrini: This would be a good candidate for a helper function; something like `drawerIconWrapper(this.imagesBaseURL, 'gear', 'settings-icon')`

src/js/plugins/plugin.mobile_nav.js
```diff
@@ -133,50 +134,74 @@ BookReader.prototype.buildToolbarElement = (function (super_) {
 BookReader.prototype.buildMobileDrawerElement = function() {
   let experimentalHtml = '';
   if (this.enableExperimentalControls) {
-    experimentalHtml += "<div class=\"DrawerSettingsTitle\">Experimental (may not work)</div>"
-        +"        <button class='BRaction default high-contrast-button'>Toggle high contrast</button>";
+    experimentalHtml = `
+        <p class="DrawerSettingsTitle">Experimental (may not work)</p>
```
cdrini: I'm a little wary of changing the HTML itself; that might have subtle regressions in how things look which would be kind of annoying to test/deal with.

```suggestion
        <div class="DrawerSettingsTitle">Experimental (may not work)</div>
```

src/js/plugins/plugin.mobile_nav.js
```diff
@@ -133,50 +134,74 @@ BookReader.prototype.buildToolbarElement = (function (super_) {
...
+    <span>
+        <span class="DrawerIconWrapper">
+            <img class="DrawerIcon" src="${`${this.imagesBaseURL}icon_gear.svg`}" alt="settings-icon"/>
+        </span>
+        Settings
+    </span>
+    <div class=DrawerSettingsWrapper>
+        <div class="DrawerSettingsLayoutWrapper">
+            <button class="DrawerLayoutButton one_page_mode">
+                <img src="${`${this.imagesBaseURL}icon_one_page.svg`}" alt="Single Page"/>
```
cdrini: No need for nested `${}`s (ditto elsewhere)

```suggestion
                <img src="${this.imagesBaseURL}icon_one_page.svg" alt="Single Page"/>
```

src/js/plugins/plugin.mobile_nav.js
```diff
@@ -133,50 +134,74 @@ BookReader.prototype.buildToolbarElement = (function (super_) {
...
+    <div class="BRmobileShare"></div>
+  `;
+  const navMenu = `
+    <nav id="BRmobileMenu" class="BRmobileMenu">
+        <ul>
+            <li>${settingsSection}</li>
+            <li class="BRmobileMenu__moreInfoRow">${moreInfo}</li>
+            <li>${share}</li>
+        </ul>
+    </nav>
```
cdrini: This is so much nicer 😊

src/js/plugins/plugin.mobile_nav.js
```diff
@@ -9,122 +9,122 @@
...
+        extensions: [ "panelshadow" ],
+      }, {
+        offCanvas: {
+          wrapPageIfNeeded: false,
+          zposition: 'next',
+          pageSelector: this.el,
         }
+      });
+
+      let $BRpageviewField = $mmenuEl.find('.BRpageviewValue');
```
cdrini: Oh drats! Looks like `estlint` replaced all the `var`s with `let` -\_-`; some of these can/should be `const`; I'll see if I can add the `prefer-const` rule to the eslint config.

src/js/plugins/plugin.mobile_nav.js
```diff
@@ -0,0 +1,220 @@
...
+
+    if (this.enableMobileNav) {
+      let $drawerEl = this.buildMobileDrawerElement();
+      this.refs.$br.append($drawerEl);
+
+      // Render info into mobile info before mmenu
+      this.buildInfoDiv(this.$('.BRmobileInfo'));
+      this.buildShareDiv(this.$('.BRmobileShare'));
+
+      $mmenuEl = $drawerEl;
```
cdrini: Why not define it here?
```suggestion
      const $mmenuEl = $drawerEl;
```

src/js/plugins/plugin.mobile_nav.js
```diff
@@ -0,0 +1,220 @@
...
+        Share This Book
+    </span>
+    <div class="BRmobileShare"></div>
+  `;
+  const navMenu = `
+    <nav id="BRmobileMenu" class="BRmobileMenu">
+        <ul>
+            <li class="br-mmenu settings">${settingsSection}</li>
+            <li class="br-mmenu more-info BRmobileMenu__moreInfoRow">${moreInfo}</li>
+            <li class="br-mmenu share">${share}</li>
```
cdrini: These names are too general; they're likely to conflict with something else (especially when in archive.org). Let's try this

```suggestion
            <li class="BRmobileMenu__settings">${settingsSection}</li>
            <li class="BRmobileMenu__moreInfoRow">${moreInfo}</li>
            <li class="BRmobileMenu__share">${share}</li>
```

src/js/plugins/plugin.mobile_nav.js
iisa: quite interesting find. on line 99 it is referenced and when we define inside of the if block, it doesn't get hoisted up to the top like a `var` does

src/js/plugins/plugin.mobile_nav.js
iisa: hmmm, even after I checked it? p offers the same `block` display + it's semantic. It displays as expected

src/js/plugins/plugin.mobile_nav.js
iisa: true.. but I would like to hold for now so I can get through the rest of the plugins.

src/js/plugins/plugin.mobile_nav.js
iisa: ah yes good catch. for some reason, i thought i was in a deeply nested string literal

src/js/plugins/plugin.mobile_nav.js
iisa: ok. will update locally as they connect with unit test

src/js/plugins/plugin.mobile_nav.js
iisa: + it has a class to attach a style to.  I will hold with the update, it returns as the same on prod and demo

src/js/plugins/plugin.mobile_nav.js
iisa: I fixed them :)

src/js/plugins/plugin.mobile_nav.js
```diff
@@ -184,20 +184,20 @@ BookReader.prototype.buildMobileDrawerElement = function() {
   `;
   const share = `
     <span>
-        <span class="DrawerIconWrapper">
-            <img class="DrawerIcon" src="${`${this.imagesBaseURL}icon_share.svg`}" alt="info-share"/>
-        </span>
-        Share This Book
+      <span class="DrawerIconWrapper">
+        <img class="DrawerIcon" src="${`${this.imagesBaseURL}icon_share.svg`}" alt="info-share"/>
```
cdrini:
```suggestion
        <img class="DrawerIcon" src="${this.imagesBaseURL}icon_share.svg" alt="info-share"/>
```

iisa: gahh!

PR #226, opened 2020-03-05
Separate Navbar.js out of BookReader.js

cdrini:
- Best to review this this commit by commit to better see the changes
- Everything is still exposed on BookReader core, but will be deprecated, so I'm ok with that syntax being super ugly :P


src/js/BookReader.js
```diff
@@ -343,7 +344,13 @@ BookReader.prototype.setup = function(options) {
   this.getPageProp = options.getPageProp || BookReader.prototype.getPageProp;
   this.getSpreadIndices = options.getSpreadIndices || BookReader.prototype.getSpreadIndices;
   this.leafNumToIndex = options.leafNumToIndex || BookReader.prototype.leafNumToIndex;
+
+  /** @type {{[name: string]: JQuery}} */
   this.refs = {};
+  
+  this.components = {
```
iisa: Will this be the place to add new components?
What do components consist of?
Do you think we need a lil note above this to instruct future people to register components here?

src/js/BookReader/Navbar/Navbar.js
```diff
@@ -0,0 +1,180 @@
...
+      value: br.currentIndex(),
+      range: "min"
+    });
+
+    $slider.on('slide', (event, ui) => {
+      this.updateNavPageNum(ui.value);
+      return true;
+    });
+
+    $slider.on('slidechange', (event, ui) => {
```
iisa: What do you think about coupling event handlers with the DOM string for BR in general - is this the way to go? or parse them out to another function?

src/js/BookReader/Navbar/Navbar.js
```diff
@@ -0,0 +1,180 @@
...
+
+    this.updateNavPageNum(br.currentIndex());
+
+    return this.$nav;
+  }
+
+  /**
+   * Initialize the navigation bar when embedded
+   */
+  initEmbed() {
```
iisa: Question: if we call `new Navbar()`, and then this gets called, do we draw navbar twice?

src/js/BookReader/Navbar/Navbar.js
```diff
@@ -0,0 +1,180 @@
...
+             <a href="${thisLink}" target="_blank">${br.bookTitle}</a>
+          </span>
+          <span class="BRtoolbarbuttons">
+            <button class="BRicon book_left"></button>
+            <button class="BRicon book_right"></button>
+            <button class="BRicon full"></button>
+          </span>
+      </div>`);
+    this.$root.append(this.$nav);
+    br.refs.$br.append(this.$root);
```
iisa: do both navs get appended at both places?

tests/BookReader/Navbar/Navbar.test.js
```diff
@@ -0,0 +1,16 @@
...
+    });
+
+    test('handle regular page numbers', () => {
+        expect(f(3, 40, '14', '', 40)).toBe('Page 14 of 40');
+    });
+
+    test('handle no max page', () => {
+        expect(f(3, 40, '14', '', null)).toBe('Page 14');
+    })
+});
```
iisa: I am assuming we can catch main issues re: button clicking with the e2e test. 👍 What else other tests do we need here?

src/js/BookReader/Navbar/Navbar.js
cdrini: Good question! The constructor doesn't render anything; `init()` or `initEmbed()` are the one's that actually render 👍

src/js/BookReader/Navbar/Navbar.js
cdrini: Yep! Only one of these `init`s is expected to be run.

PR #229, opened 2020-03-06
[es6-bookreader-core] Separate options.js

cdrini:
Pull out the default options into their own file, and clean up / expand on docstrings


jbuckner:
Nice! 👏

PR #230, opened 2020-03-09
Es6 plugin: Search

iisa:
#### Test Steps
- pull branch
- fresh module install, cmd: `npm install`
- run server: `npm run serve`
- Test desktop
  - go to: http://127.0.0.1:8000/BookReaderDemo/demo-ia-olivertwist.html
  - search text using input: `child` (wait a while, then wait a lil more) -->  results markers successfully pop up
  - hover over results marker. --> text tooltip pops up
    - known: page highlight bleeds over tooltip
  - hover off of results marker --> tooltip disappears
  - click on marker --> takes you to page with results highlighted
  - refresh page that is full of search results (wait for it...) --> results markers show
  - search a new word: `shmeowza` --> no results window pops up & fades out
- Run tests on mobile device (browser dev emulator)
  - go to: http://127.0.0.1:8000/BookReaderDemo/demo-ia-olivertwist.html
  - open 🍔 --> search section is visible
 - search text using input: `child` (wait a while, then wait a lil more) -->  results & count appear below input
  - click on search result -> takes you to page
 - refresh page --> results are shown in the search drawer
  - search a new word: `shmeowza` --> no results are displayed

internal ticket: https://webarchive.jira.com/browse/WEBDEV-3206


src/js/plugins/plugin.search.js
```diff
@@ -0,0 +1,575 @@
...
+  let i, j;
+  for (i = 0; i < results.matches.length; i++) {
+    for (j = 0; j < results.matches[i].par[0].boxes.length; j++) {
+      const box = results.matches[i].par[0].boxes[j];
+      const pageIndex = this.leafNumToIndex(box.page);
+      if (jQuery.inArray(pageIndex, visiblePages) >= 0) {
+        return true;
+      }
+    }
+  }
```
cdrini:
```suggestion
  results.matches.forEach(match => {
    match.par[0].boxes.forEach(box =>
      const pageIndex = this.leafNumToIndex(box.page);
      if (jQuery.inArray(pageIndex, visiblePages) >= 0) {
        return true;
      }
    });
  });
```

src/js/plugins/plugin.search.js
```diff
@@ -0,0 +1,575 @@
...
+  const defaultOptions = {
+    /**
+     * @type { bool } - (default=false) goToFirstResult - jump to the first result
+     */
+    goToFirstResult: false,
+    /**
+     * @type { bool } - (default=false) disablePopup - don't show the modal progress
+     */
+    disablePopup: false
+  };
```
cdrini: Let's not remove the `error` and `success` options; those are still public options, so we don't know if someone who uses BR outside of ia is using them. We need to try to keep things as backwards compatible as possible, because we're not providing notice of breaking changes. This would require a v5 version bump.

src/js/plugins/plugin.search.js
```diff
@@ -0,0 +1,575 @@
...
+  $.ajax({
+    url: url,
+    dataType: 'jsonp'
+  }).then(processSearchResults);
+};
+
+/**
+ * Search Results return handler
+ * @param { object } results
+ * @param { array } results.matches
```
cdrini: Oh, last thing; let's give these some `typedef`s so that we can reference their types, and add info about what fields are present. I think that would help make things much clearer!

src/js/plugins/plugin.search.js
```diff
@@ -0,0 +1,575 @@
...
+  if (firstResultIndex !== null) {
+    this.jumpToIndex(firstResultIndex);
+  }
+}
+
+/**
+ * Main search results error handler
+ * @param { array } results
+ */
+BookReader.prototype.BRSearchCallbackError = function(results) {
```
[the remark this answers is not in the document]
iisa: thanks for catching! I was deep in consolidating..

src/js/plugins/plugin.search.js
```diff
@@ -0,0 +1,556 @@
...
+  const url = `${baseUrl}${paramStr}`;
+
+  if (!options.disablePopup) {
+    this.showProgressPopup(
+      `<img class="searchmarker" src="${this.imagesBaseURL}marker_srch-on.svg" />
+      Search results will appear below...`);
+  }
+
+  const processSearchResults = (data) => {
+    if (data.error || 0 == data.matches.length) {
```
nsharma123: Per es6 standards, you can prevent `Yoda condition` where the literal value coming first and variable is second. So it can be write something like `data.matches.length === 0` (including strict type with ===). Thanks

src/js/plugins/plugin.search.js
```diff
@@ -0,0 +1,556 @@
...
+  this.searchResults = null;
+  this.trigger(BookReader.eventNames.fragmentChange);
+  this.$('.BRnavpos .BRsearch').remove();
+  this.$('.BRmobileSearchResultWrapper').empty(); // Empty mobile results
+};
+
+/**
+ * Returns true if a search highlight is currently being displayed
+ */
+BookReader.prototype.searchHighlightVisible = function() {
```
nsharma123: lets add jsdocs include `@return { ... }`

src/js/plugins/plugin.search.js
iisa: thanks, updated.

src/js/plugins/plugin.search.js
iisa: yes I like. :excellent:

src/js/plugins/plugin.search.js
iisa: thanks for catching. added

src/js/plugins/plugin.search.js
```diff
@@ -0,0 +1,599 @@
...
+})(BookReader.prototype.initToolbar);
+
+/**
+ * Search Options
+ * @typedef {object} searchOptions
+ * @property {boolean} goToFirstResult
+ * @property {boolean} disablePopup
+ * @property {(null|function)} error - @deprecated at v.5.0
+ * @property {(null|function)} success - @deprecated at v.5.0
+ */
```
iisa: new `@typedef`: Search Options
We pass this around to success & error callbacks, so it was helpful to make one for it.
I abstracted it up to the "top level," vscode couldn't back link the references to this when I added it right above the callback inside the function

src/js/plugins/plugin.search.js
```diff
@@ -0,0 +1,599 @@
...
+  }).then(processSearchResults);
+};
+
+/**
+ * Search Inside results shape
+ * @typedef {object} searchInsideResults
+ * @property {string} error
+ * @property {array} matches
+ * @property {boolean} indexed
+ */
```
iisa: new @typedef: `searchInsideResults`
This also gets passed to the varied callbacks we have.

iisa:
Hi @cdrini, I addressed the suggestions you had, thanks!
I will hold here though, because some of the changes we introduced in the for loop updates started breaking the functionality.

This is a great place to leave this file.

src/js/plugins/plugin.search.js
```diff
@@ -0,0 +1,599 @@
...
+  this._BRSearchCallbackError(results, $el);
+};
+
+/**
+ * @private draws search results error
+ * @callback
+ * @param {object} results
+
+ * @param {jQuery} $el
+ * @param {boolean} fade
```
cdrini:
```suggestion
 * @param {SearchInsideResults} results
 * @param {jQuery} $el
 * @param {boolean} fade
```

src/js/plugins/plugin.search.js
cdrini: This should get all of them :+1:

```suggestion
/**
  * @typedef {object} SearchInsideMatchBox
  * @property {number} page
  * @property {number} r
  * @property {number} l
  * @property {number} b
  * @property {number} t
  * @property {HTMLDivElement} [div]
  */

/**
 * @typedef {object} SearchInsideMatch
 * @property {string} text
 * @property {Array<{ page: number, boxes: SearchInsideMatchBox[] }>} par
 */

/**
 * @typedef {object} SearchInsideResults
 * @property {string} error
 * @property {SearchInsideMatch[]} matches
 * @property {boolean} indexed
 */
```

src/js/plugins/plugin.search.js
```diff
@@ -0,0 +1,599 @@
...
+ * @typedef {object} searchInsideResults
+ * @property {string} error
+ * @property {array} matches
+ * @property {boolean} indexed
+ */
+
+/**
+ * Search Results return handler
+ * @callback
+ * @param {searchInsideResults} results
```
cdrini:
```suggestion
 * @param {SearchInsideResults} results
```

src/js/plugins/plugin.search.js
```diff
@@ -0,0 +1,599 @@
...
+/** @deprecated */
+BookReader.prototype.BRSearchCallbackErrorDesktop = function(results, options) {
+  const $el = $(this.popup);
+  this._BRSearchCallbackError(results, $el, true);
+}
+
+/**
+ * Main search results error handler
+ * @callback
+ * @param {searchInsideResults} results
```
cdrini:
```suggestion
 * @param {SearchInsideResults} results
```

src/js/plugins/plugin.search.js
```diff
@@ -0,0 +1,599 @@
...
+    const hasCustomSuccess = typeof options.success === 'function';
+
+    if (responseHasError) {
+      hasCustomError
+        ? options.error(searchInsideResults, options)
+        : this.BRSearchCallbackError(searchInsideResults, options);
+    } else {
+      hasCustomSuccess
+        ? options.success(searchInsideResults, options)
+        : this.BRSearchCallback(searchInsideResults, options);
```
cdrini: For backwards compatibility
```suggestion
        ? options.error.call(this, searchInsideResults, options)
        : this.BRSearchCallbackError(searchInsideResults, options);
    } else {
      hasCustomSuccess
        ? options.success.call(this, searchInsideResults, options)
        : this.BRSearchCallback(searchInsideResults, options);
```

iisa: thanks! that is important 📝

PR #233, opened 2020-03-12
[es6-bookreader-core] Separate Toolbar.js

cdrini:
Oy vey. Gah. So basically realized at some point that because the plugin's _override_ bookreader methods like `initToolbar`, what I was doing wouldn't work, because then the internal e.g. `Toolbar` methods would call their own copy, not the overriden copy. Created some dirty hacks to make this work, which we should kill once we go to v5 and deprecate a lot of these methods.

Review app: https://58-review-br-es6ify-b09jkq.archive.org/details/trees0000rids


iisa:
> Oy vey. Gah. So basically realized at some point that because the plugin's _override_ bookreader methods like `initToolbar`, what I was doing wouldn't work, because then the internal e.g. `Toolbar` methods would call their own copy, not the overriden copy. Created some dirty hacks to make this work, which we should kill once we go to v5 and deprecate a lot of these methods.
> 
> Review app: https://58-review-br-es6ify-b09jkq.archive.org/details/trees0000rids

good catch on the overrides! taking a closer look now...

src/js/BookReader/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,383 @@
...
+      $leftCol.append($(`
+        <div class="BRimageW">
+          <img src="${br.thumbnail}" alt="${escapeHTML(br.bookTitle)}" />
+        </div>`));
+    }
+
+    const $rightCol = $(`<div class="BRinfoRightCol">`);
+
+    // A loop to build fields
+    for (let {extraValueClass='', label, value} of br.metadata) {
```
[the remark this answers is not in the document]
iisa: is there a reference somewhere of `br.metadata`'s values?

src/js/BookReader/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,383 @@
...
+    if ($BRtoolbar && $BRtoolbar.css('display') === 'block') {
+      return ($BRtoolbar.outerHeight() + parseInt($BRtoolbar.css('top')));
+    } else {
+      return 0;
+    }
+  }
+}
+
+export function blankInfoDiv() {
+  // FIXME the <button> el is closed with a </a>
```
iisa: is this comment still relevant?

src/js/BookReader/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,383 @@
...
+    <div class="BRfloat BRshare">
+      <div class="BRfloatHead">
+        Share
+        <button class="floatShut" href="javascript:;" onclick="$.fn.colorbox.close();"><span class="shift">Close</span></a>
+      </div>
+    </div>`);
+}
+
+/**
+ * Helper opens a popup window. On mobile it only opens a new tab. Used for share.
```
iisa: thanks

iisa: is this used outside of this file?

src/js/BookReader/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,383 @@
...
+ * @param {string} href
+ * @param {number} width
+ * @param {number} height
+ * @param {string} name
+ */
+export function createPopup(href, width, height, name) {
```
iisa: do we expect this to be used outside of the file? is it possible to move this and `blankShareDiv` to its own place?

src/js/BookReader/options.js
```diff
@@ -92,7 +92,10 @@ export const DEFAULT_OPTIONS = {
    * https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes ) */
   bookLanguage: null,
   
-  /** Fields used to populate the info window */
+  /**
+   * @type {Array<{label: string, value: *, extraValueClass: string?}>}
+   * Fields used to populate the info window
+   **/
   metadata: [],
```
iisa: Where does this array get populated?

src/js/BookReader/utils/classes.js
```diff
@@ -0,0 +1,36 @@
...
+ * @template TFrom the class we'll be getting the method from
+ * @template TTo the class we'll be adding the method to
+ * @param {new () => TFrom} FromClass the class to get the method from
+ * @param {string} fromMethod the method's name in FromClass
+ * @param {function(TFrom): TTo} fromTransform how to get the TTo `this` to use when setting the method on TFrom
+ * @param {new () => TTo} ToClass the class to add the method to
+ * @param {string} toMethod the name of the method to add to TTo (likely will be equal to fromMethod)
+ * @param {function(TTo): TFrom} toTransform how to get the TFrom this to use when calling the new method
+ */
+export function exposeLinked(FromClass, fromMethod, fromTransform, ToClass, toMethod, toTransform) {
```
iisa: q: how long do we want our args list to be?

src/js/BookReader/utils/classes.js
```diff
@@ -0,0 +1,36 @@
...
+  // Wrapper function needed to "capture" the current version of fromMethod
+  let wrapper = (fn => {
+    return function () {
```
iisa: can we use () => {} here?

src/js/BookReader/utils/classes.js
```diff
@@ -0,0 +1,36 @@
...
+
+  Object.defineProperty(ToClass.prototype, toMethod, {
+    get() { return wrapper; },
+    set(overrideFn) {
+      // overrideFn expects `this` to be ToClass, so ensure as such
+      // But we can also call this method from FromClass; need to ensure
+      // it's always called with a ToClass
+      FromClass.prototype[fromMethod] = function () {
+        const newThis = this instanceof FromClass ? fromTransform(this) : this;
+        return overrideFn.apply(newThis, arguments);
```
iisa: that shiny new this

src/js/BookReader/utils/classes.js
```diff
@@ -0,0 +1,36 @@
...
+ * WARNING: This modifies FromClass' prototype! So FromClasses cannot be shared between
+ * different ToClasses.
+ * @template TFrom the class we'll be getting the method from
+ * @template TTo the class we'll be adding the method to
+ * @param {new () => TFrom} FromClass the class to get the method from
+ * @param {string} fromMethod the method's name in FromClass
+ * @param {function(TFrom): TTo} fromTransform how to get the TTo `this` to use when setting the method on TFrom
+ * @param {new () => TTo} ToClass the class to add the method to
+ * @param {string} toMethod the name of the method to add to TTo (likely will be equal to fromMethod)
+ * @param {function(TTo): TFrom} toTransform how to get the TFrom this to use when calling the new method
```
iisa: This function is sure doing some heavy lifting. 💪 It makes me just love the power of JS even more ⚡️ 

There may be a better way to document this.  Some doc feedback - 
`TTo` and `TFrom` are getting a bit lost.
The function passes around scope, can we make that a bit clearer in this document block?

just brainstorming...
`toMethod the name of the method to add to TTo` => `toMethod - the method that gets updated with new/extended scope`

src/js/BookReader.js
```diff
@@ -4133,17 +3782,7 @@ BookReader.prototype.reloadImages = function() {
...
-BookReader.prototype.getToolBarHeight = function() {
-  if (this.refs.$BRtoolbar && this.refs.$BRtoolbar.css('display') === 'block') {
-    return (this.refs.$BRtoolbar.outerHeight() + parseInt(this.refs.$BRtoolbar.css('top')));
-  } else {
-    return 0;
-  }
-}
-
-/**
+ * @todo Move to Navbar.js
```
iisa: if we won't do this now, can we make an issue for it to do later?

tests/BookReader/utils/classes.test.js
```diff
@@ -0,0 +1,88 @@
...
+            }(BookReader.prototype.createButton);
+        }
+
+        const br = new BookReader();
+        expect(br.createButton()).toBe('plugin 2 plugin 1 root button');
+        expect(br.component.createButton()).toBe('plugin 2 plugin 1 root button');
+        // Calls the overriden method on BookReader!
+        expect(br.component.initToolbar()).toEqual(['plugin 2 plugin 1 root button']);
+    });
+});
```
iisa: 👏 thanks for these tests! 👏

src/js/BookReader.js
```diff
@@ -2978,207 +2974,63 @@ BookReader.prototype.jumpIndexForRightEdgePageX = function(pageX) {
...
+exposeComponentMethod(Toolbar, 'toolbar', 'initToolbar');
+BookReader.prototype.buildShareDiv = Toolbar.prototype.buildShareDiv;
+exposeComponentMethod(Toolbar, 'toolbar', 'buildShareDiv');
+BookReader.prototype.buildInfoDiv = Toolbar.prototype.buildInfoDiv;
+exposeComponentMethod(Toolbar, 'toolbar', 'buildInfoDiv');
+BookReader.prototype.getToolBarHeight = Toolbar.prototype.getToolBarHeight;
+exposeComponentMethod(Toolbar, 'toolbar', 'getToolBarHeight');
+/** @deprecated zoom no longer in toolbar */
+BookReader.prototype.updateToolbarZoom = Toolbar.prototype.updateToolbarZoom;
+exposeComponentMethod(Toolbar, 'toolbar', 'updateToolbarZoom');
```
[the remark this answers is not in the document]
cdrini: I was considering that, but I think that would break the autocompletes; right now we get autocompletion for these methods because of the `BookReader.prototype.Foo = Toolbar.prototype.Foo`; the type checker isn't smart enough to detect that through a for loop (I don't think).

src/js/BookReader.js
```diff
@@ -2978,207 +2974,63 @@ BookReader.prototype.jumpIndexForRightEdgePageX = function(pageX) {
   }
 };
 
+/**
+ * Helper method to expose a component method onto BookReader, in such a way that
+ * if the BookReader method is overriden, so is the component's method.
+ */
+function exposeComponentMethod(ComponentClass, componentKey, method, brMethod=method) {
```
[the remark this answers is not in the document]
cdrini: :+1: Fixed

src/js/BookReader.js
```diff
@@ -2978,207 +2974,63 @@ BookReader.prototype.jumpIndexForRightEdgePageX = function(pageX) {
   }
 };
 
+/**
```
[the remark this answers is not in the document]
cdrini: For now I'm keeping them ~close to where the original methods _used_ to be; I think once the components have been pulled out, we should look at what's left in BookReader.js for a final tidy up :+1:

src/js/BookReader/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,383 @@
+import { escapeHTML } from '../utils.js';
+import { EVENTS } from '../events.js';
+/** @typedef {import("../../BookReader.js").default} BookReader */
+
+export class Toolbar {
+  /**
+   * @param {BookReader} br
+   */
+  constructor(br) {
+    this.br = br;
```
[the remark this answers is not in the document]
cdrini: :D Haha, I feel you; I'm not super crazy over it either (honestly I decided do to this as a result of thinking "this is taking to long :P "), but I figure any more would require a _lot_ of refactoring. So the components are still tightly coupled to BookReader core (which is something we can work towards later), but are _much_ more independent from each other, which I think will make a big difference :+1:

src/js/BookReader/Toolbar/Toolbar.js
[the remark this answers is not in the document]
cdrini: Yep! It's type is in `options.js` :) Added the example off the README to it.
image

src/js/BookReader/Toolbar/Toolbar.js
cdrini: Yep; I added it during the refactor 😅 I didn't want to figure out what the "correct" thing was (not even sure how the browser is rendering this). Will create an issue :+1:

src/js/BookReader/Toolbar/Toolbar.js
cdrini: Nope; it's exported only for testing :+1:

src/js/BookReader/options.js
cdrini: Populated by the developer who's instantiating BR; added an example from the README to the docstring here.

src/js/BookReader/utils/classes.js
cdrini: Unfortunately not; it needs to be able to the `this` from what it was called with (as opposed to `exposeLinked`'s `this`) 😬

src/js/BookReader/utils/classes.js
cdrini: ✨

tests/BookReader/utils/classes.test.js
cdrini: Haha, practically rewrote the function while writing the test, cause I found a bunch of bugs 😅

src/js/BookReader/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,383 @@
...
+      onLoad: () => {
+        br.trigger(EVENTS.stop);
+      }
+    });
+  }
+
+  /**
+   * @todo .BRzoom doesn't exist anywhere, so this is likely dead code
+   * Update the displayed zoom factor based on reduction factor
+   * @param {number} reduce
```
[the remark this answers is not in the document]
cdrini: Likely yes :P Fixed

src/js/BookReader/utils/classes.js
cdrini: :D Haha, why could you be asking :P I think I'll call this a special circumstance; I don't think there's a way to write this function that's not awkward/complex, so will just have to hope future devs use it without digging into it too deeply; we should work toward ripping it out when we find a good alternative way to have the same effect.

src/js/BookReader/utils/classes.js
cdrini: It really is crazy; like this would be near impossible in most other languages (although to be fair it was kind of near impossible in JS, too :P ). The `@template` types are mostly an attempt to make this type check, but I didn't succeed; I'll see if I can get them to behave, and if not I'll drop them :+1:

src/js/BookReader.js
cdrini: Actually, I think I'll just remove the comment; I'm a little on the fence about moving the method; I'll let future devs decide :+1:

src/js/BookReader/utils/classes.js
cdrini: I think I'll rename this to `exposeOverrideable`; that seems a little clearer.

src/js/BookReader/Toolbar/Toolbar.js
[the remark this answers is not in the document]
iisa: nah this is fine. just wanted to color the commentary :)

src/js/BookReader/utils/classes.js
iisa: i figured. just had to ask

src/js/BookReader/utils/classes.js
cdrini: Fixed; and got it fully to type-check, so auto-complete will even provide the _name of the method on FromClass_! I also pushed the `@template` type parameters to the bottom of the doc string; they're not really useful in this case, but needed for the autocompletion.

src/js/BookReader/Toolbar/Toolbar.js
cdrini: I'd rather punt on this one; I think future us can decide if this is generally useful enough to place in a shared util file :+1:

src/js/BookReader.js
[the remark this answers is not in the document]
cdrini: https://github.com/internetarchive/bookreader/issues/240

src/js/BookReader/Toolbar/Toolbar.js
cdrini: https://github.com/internetarchive/bookreader/issues/240

src/js/BookReader.js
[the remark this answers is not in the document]
iisa: ok i can handle that

PR #234, opened 2020-03-13
live server should ignore changes to src files, since it only uses build files

cdrini:


package.json
```diff
@@ -64,7 +64,7 @@
     "lint": "eslint src/js",
     "lint:fix": "eslint --fix src/js",
     "serve": "npx http-server . --port 8000",
-    "serve-live": "npx live-server . --port 8000 --ignore=coverage,tests",
+    "serve-live": "npx live-server . --port 8000 --ignore=coverage,tests,src",
```
iisa: q: what watches the `src` files?

cdrini: webpack/sass; `serve-live` is more of an "internal" command, really; devs should run `serve-dev`. That watches the src js/sass file and recompiles as well as starting the live server.

PR #235, opened 2020-03-13
Make demo JS ie11 comptaible

cdrini:
Object.assign doesn't work in ie11!


iisa:
we love Object.assign

PR #243, opened 2020-03-18
Changed default-param with OR operator

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-3235


src/js/plugins/plugin.resume.js
```diff
@@ -44,10 +44,10 @@ BookReader.prototype.getResumeValue = function() {
  * Can be overriden for different implementation
  *
  * @param {Number} index leaf index
- * @param {String} cookieName
  */
-BookReader.prototype.updateResumeValue = function(index, cookieName='br-resume') {
+BookReader.prototype.updateResumeValue = function(index, cookieName) {
```
nsharma123: Default parameter getting `null` value always which is setting into cookie as well. So we need to apply default with || operator to prevent the case when default parameter value coming `null`. @cdrini !

See reference: https://www.codereadability.com/javascript-default-parameters-with-or-operator/

cdrini:
Good catch @nsharma123 ! Yep, it looks like BookReaderJSIA on petabox is passing in null; so we need to support that old syntax. Although I think the root cause of this issue is that it looks like there was a typo in the code! It was reading `rsumeCookiePath` instead of `resumeCookiePath` from the options :P

src/js/plugins/plugin.resume.js
```diff
@@ -44,10 +44,11 @@ BookReader.prototype.getResumeValue = function() {
...
  * @param {Number} index leaf index
- * @param {String} cookieName
+ * @param {string} [cookieName]
  */
-BookReader.prototype.updateResumeValue = function(index, cookieName='br-resume') {
+BookReader.prototype.updateResumeValue = function(index, cookieName) {
   const ttl = new Date(+new Date + 12096e5); // 2 weeks
-  const path = this.options.rsumeCookiePath || window.location.pathname;
+  const path = this.options.resumeCookiePath || window.location.pathname;
+  cookieName = cookieName || 'br-resume';
```
shaneriley: I'd suggest assigning to a new local variable, as our usual ESLint rules prevent parameter reassignment https://eslint.org/docs/rules/no-param-reassign

cdrini: :+1: fixed

PR #244, opened 2020-03-18
Fix/search highlight

cdrini:
Fix search match not highlighting if match on currently open page in 2up


src/js/plugins/plugin.search.js
```diff
@@ -412,7 +412,7 @@ BookReader.prototype.updateSearchHilites2UP = function() {
       const pageIsInView = jQuery.inArray(pageIndex, this.displayedIndices) >= 0;
 
       if (pageIsInView) {
-        if (box.div === null) {
+        if (!box.div) {
```
cdrini: Because apparently `undefined == null` is true only with double equality comparison 🤷‍♂

cdrini:
Ignoring codecov on this for now to save time; we've tested manually.

PR #247, opened 2020-03-20
[es6-bookreader-core] Separate Mode2Up.js

cdrini:
Still need to add tests, but otherwise, 1.1K lines moved, and 30 methods stated for deprecation!

- Everything that was labelled as deprecated I check both BookReader repo and petabox to see if anything override, so we should be able to safely remove.
- Because this was so big, did _very little_ in terms of es6-ing, and will likely keep it that way.
- Need to unit tests, but this should be good for a review. There's so much code moved I'm not sure if reading is feasible :/ Maybe testing it might be the only to get through it.


src/js/BookReader.js
```diff
@@ -188,11 +189,16 @@ BookReader.prototype.setup = function(options) {
     toolbar: new Toolbar(this),
   };
 
+  this._modes = {
```
iisa: :yas:

src/js/BookReader/Mode2Up.js
```diff
@@ -0,0 +1,1148 @@
...
+    this.br.twoPage.autofit = reductionFactor.autofit;
+    this.br.reduce = reductionFactor.reduce;
+    this.br.pageScale = this.br.reduce; // preserve current reduce
+  
+    // Preserve view center position
+    const oldCenter = this.getViewCenter();
+  
+    // If zooming in, reload imgs.  DOM elements will be removed by prepareTwoPageView
+    // $$$ An improvement would be to use the low res image until the larger one is loaded.
+    if (1 == direction) {
```
iisa: given that last bug from plugin release, leaving this as-is is 👍

Issue #213, opened 2020-03-03
New End to End Test: Autoplay Plugin


iisa:
Create an end-to-end test for the Autoplay plugin

MayankVir:
I am newbie to open source and would like to work upon the issue

cdrini:
Hi @MayankVir ! Oops, it looks like this issue was already completed.

PR #261, opened 2020-03-26
Adds autoplay e2e tests

shaneriley:
Completes #213

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,8 +67,8 @@
     "serve-live": "npx live-server . --port 8000 --ignore=coverage,tests,src",
     "serve-dev": "npx concurrently --kill-others npm:serve-live npm:build-js:watch npm:build-css:watch",
     "test": "jest",
-    "test:e2e": "testcafe",
-    "test:e2e:dev": "testcafe --live --dev",
+    "test:e2e": "BASE_URL=http://127.0.0.1:8000/BookReaderDemo/ testcafe",
+    "test:e2e:dev": "BASE_URL=http://127.0.0.1:8000/BookReaderDemo/ testcafe --live --dev",
     "test:watch": "jest --watch",
     "test:coverage": "jest --coverage --colors",
     "codecov": "codecov"
--- a/tests/e2e/base.test.js
+++ b/tests/e2e/base.test.js
@@ -1,15 +1,15 @@
 /* global fixture */
 import { runBaseTests } from './helpers/base';
 import BookReader from './models/BookReader';
+import { LOCAL_URL } from './constants';
 
-const localURL = 'http://127.0.0.1:8000/BookReaderDemo/';
 const localPages = [
   'demo-ia-plato.html',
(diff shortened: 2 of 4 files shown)


shaneriley:
Personally, I don't think the build files should be under version control. If they are, you'll keep getting file changes like these where there's seemingly no change to their sources. As for why they changed, I have no idea. Probably a combination of a difference of local Node and NPM versions.

iisa:
I agree with not always checking in built files.
I also want a solid snapshot of the code that actually works so we can depend on it.
I think @cdrini has thought about this issue for BookReader more.. any thoughts, updates?

tests/e2e/base.test.js
```diff
@@ -1,15 +1,16 @@
 /* global fixture */
 import { runBaseTests } from './helpers/base';
 import BookReader from './models/BookReader';
+import { LOCAL_URL } from './constants';
 
-const localURL = 'http://127.0.0.1:8000/BookReaderDemo/';
```
cdrini: Let's not pull this out; this would be better as an environment variable or parameter. Having it in a separate file will make that significantly more difficult.

cdrini:
Yep! Agreed. The main issue is that removing the build files would be a breaking change and be a major version bump (making we'd need to provide ample warning, and likely coordinate with known downstream users like Mitra). Whenever we decide to do that major bump, removing the build files will be part of it 👍

Why did you close the pr? The other changes seem fine. A force push should fix things.

tests/e2e/base.test.js
shaneriley: I agree on the env variable (I use an env variable in Petabox's TestCafe tests). Do you want that set in the NPM scripts for testcafe test running, or should we bring in a package like [dotenv](https://www.npmjs.com/package/dotenv) to bring them in from a local file?

shaneriley:
Sorry, must have pushed the wrong comment button this morning.

tests/e2e/base.test.js
cdrini: Your call how much you want to do as part of this PR. I think having a basic environment variable, which defaults to localhost is best in the spirit of yagni. If we end up with a lot of environment variables, then we should consider something more advanced like dotenv.

package.json
```diff
@@ -67,8 +67,8 @@
     "serve-live": "npx live-server . --port 8000 --ignore=coverage,tests,src",
     "serve-dev": "npx concurrently --kill-others npm:serve-live npm:build-js:watch npm:build-css:watch",
     "test": "jest",
-    "test:e2e": "testcafe",
-    "test:e2e:dev": "testcafe --live --dev",
+    "test:e2e": "BASE_URL=http://127.0.0.1:8000/BookReaderDemo/ testcafe",
+    "test:e2e:dev": "BASE_URL=http://127.0.0.1:8000/BookReaderDemo/ testcafe --live --dev",
```
iisa: ok

PR #279, opened 2020-04-08
Add passing of loan id to analytics call for UserChangedView event

bfalling:
We'd like to include a hashed loan identifier when tracking BookReader page views. The loan id is being added to the back-end code. When present in the BookReader options, we'd like to include in the analytics call.


src/js/plugins/plugin.archive_analytics.js
```diff
@@ -53,7 +53,10 @@ BookReader.prototype.archiveAnalyticsSendFragmentChange = function() {
     archive_analytics.send_ping(values, null, "augment_for_ao_site");
 
     // Also send tracking event ping
-    archive_analytics.send_event('BookReader', 'UserChangedView', null);
+    var additionalEventParams = this.options.lendingInfo && this.options.lendingInfo.loanId
+      ? { loanId: this.options.lendingInfo.loanId }
+      : {};
```
iisa, line 58: q: is this default an object because `archive_analytics.send_event` expects it to be one?

bfalling: Technically, anything false-y would work, but the default value it gets turned to is an object.

PR #283, opened 2020-04-10
Add keyboard event listener for chrome

### LINKED ISSUE #282: Allow zooming in and out with the +- keys
**Context**

It would be nice to be able to zoom in and out using the `-` and `+` keys (without holding shift, so technically `-` `=`.

**Proposal & Constraints**

Just trigger zoom in and out with these keys. Same as if you hit the zoom in or out button.

I am browsing some image-heavy magazines/books and this would optimize the experience.

ishank-dev:
The mapping for ``=`` and ``-`` are different on different browsers. So I have added the key mappings that will be supported by all the browsers. This will allow zooming in and out with ``-`` and ``+`` keys, which was not possible earlier.
Check [this](https://keycode.info/ ) for reference.
fixes #282


rchrd2:
Is it possible to also verify this works in Safari?

ishank-dev:
Yes, go to this link https://keycode.info/ on your safari browser and press ``=`` and ``-`` and check what number it shows.

rchrd2:
Looks like it's this:
- 189
= 187

ishank-dev:
Yep, so it is same as that of chrome. This PR should fix the issue for safari as well.

ishank-dev:
Update: PR has been tested and I have added the build files in my last commit.
tagging @iisa @cdrini for review.

PR #285, opened 2020-04-13
Adds customizable view mode controls

shaneriley:
This adds options for the controls (currently just the view mode controls, but more/all could be converted) to be toggled, have their class name changed, and use a custom template function to return an HTML string to represent the control.

An example has been added to the BookReader demos at [http://127.0.0.1:8080/BookReaderDemo/toggle_controls.html](http://127.0.0.1:8080/BookReaderDemo/toggle_controls.html). `onePage`, `twoPage`, and `thumbnail` are all visible by default and use internal templates provided by the new `Navbar` instance method `controlFor`. This is called for each of the three controls during the construction of the Navbar template upon initialization. The template functions receive the BookReader instance as a parameter in the event any backreference is needed.

Integration tests are added to the Jest tests for Navbar.


src/js/BookReader/Navbar/Navbar.js
```diff
@@ -15,6 +15,15 @@ export class Navbar {
     this.maxPageNum = null;
   }
 
+  controlFor(controlName) {
+    const option = this.br.options.controls[controlName];
```
bfalling, line 19: QUESTION: Will `this.br.options.controls` always exist at this point?

bfalling: Ah, looking at `src/js/BookReader/options.js`, I think the answer is yes.

PR #293, opened 2020-04-20
Creates leaf template to wrap pages in common markup

shaneriley:
In order to prevent tap-and-hold saving of leaf images, an overlay element was added in the form of a pseudo-element within a common container for leaf images. The container has classes that allow for future styling on a per-mode basis.


src/js/BookReader.js
```diff
@@ -581,6 +581,21 @@ BookReader.prototype.drawLeafs = function() {
   }
 };
 
+BookReader.prototype.leafTemplate = function(styles, index) {
```
cdrini, line 584: Let's place the index first, and makes styles optional. I find it gets a little lost when reading downstream code.

cdrini, line 584: Also, let's make this private. Since this repo is also used by third parties, I don't want to have to support this function being public unless we need it to be :P

```suggestion
/**
 * @private
 */
BookReader.prototype._leafTemplate = function(styles, index) {
```

src/js/BookReader.js
```diff
@@ -1789,15 +1794,17 @@ BookReader.prototype.prefetchImg = function(index) {
   }
 
   if (loadImage) {
-    var img = document.createElement("img");
-    $(img).addClass('BRpageimage').addClass('BRnoselect');
+    const div = this.leafTemplate({}, index);
+    $('<img />', {
+      'class': 'BRpage-frame BRpageimage',
```
cdrini, line 1799: BRpage-frame looks unused
```suggestion
      'class': 'BRpageimage',
```

src/js/BookReader.js
```diff
@@ -1789,15 +1794,17 @@ BookReader.prototype.prefetchImg = function(index) {
...
+    }).appendTo(div);
     if (index < 0 || index > (this._models.book.getNumLeafs() - 1) ) {
       // Facing page at beginning or end, or beyond
-      $(img).addClass('BRemptypage');
+      div.addClass('BRemptypage');
     }
-    img.src = pageURI;
-    img.uri = pageURI; // browser may rewrite src so we stash raw URI here
-    this.prefetchedImgs[index] = img;
+    div[0].uri = pageURI;
```
cdrini, line 1806: Let's keep the comment. I can't find anywhere that uses this `uri` field, but let's keep the comment for now.
```suggestion
    div[0].uri = pageURI; // browser may rewrite src so we stash raw URI here
```

tests/e2e/helpers/base.js
```diff
@@ -104,7 +104,7 @@ export function runBaseTests (br) {
     await t.click(nav.desktop.mode1Up);
     const onePageViewContainer = await Selector('.BRpageview');
     await t.expect(onePageViewContainer.visible).ok();
-    const images = onePageViewContainer.find('.BRpagediv1up');
+    const images = onePageViewContainer.find('.BRmode1up');
```
cdrini, line 107: make it clearer that a page is being selected
```suggestion
    const images = onePageViewContainer.find('.BRpagecontainer.BRmode1up');
```

tests/e2e/helpers/base.js
```diff
@@ -114,7 +114,7 @@ export function runBaseTests (br) {
     await t.click(nav.desktop.modeThumb);
     const thumbnailContainer = await Selector('.BRpageview');
     await t.expect(thumbnailContainer.visible).ok();
-    const images = thumbnailContainer.find('.BRpagedivthumb');
+    const images = thumbnailContainer.find('.BRmodethumb');
```
cdrini, line 117: Ditto
```suggestion
    const images = thumbnailContainer.find('.BRpagecontainer.BRmodethumb');
```

src/js/BookReader.js
cdrini, line 584: Also, let's rename this to be whater the css class is; i.e. `createPageContainer`

src/js/BookReader.js
```diff
@@ -669,22 +684,19 @@ BookReader.prototype.drawLeafsOnePage = function() {
...
       var leftMargin = parseInt((containerWidth - width) / 2);
 
-      var div = document.createElement('div');
-      div.className = 'BRpagediv1up pagediv' + index;
-      div.style.position = "absolute";
-      div.style.top = leafTop + 'px';
-      div.style.left = leftMargin + 'px';
-      div.style.width = width + 'px';
-      div.style.height = height + 'px';
+      const div = this.leafTemplate({
```
cdrini, line 687: And elsewhere.
```suggestion
      const pageContainer = this._createPageContainer({
```

cdrini:
Oh, also, the event capturing "screen" should _only_ be add if the book is `this.protected`; PD books shouldn't have the screen.

src/css/_BRpages.scss
```diff
@@ -2,50 +2,52 @@
...
-.BRpagedivthumb {
-  background-color: $brColorPlaceholderBg;
-  overflow: hidden;
-  box-shadow: 1px 1px 2px #333;
+  img {
+    display: block;
+    width: 100%;
+    height: 100%;
+  }
+  &:after {
```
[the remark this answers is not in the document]
shaneriley: You mean have two elements rather than a parent and a pseudo element? I always take advantage of the pseudo element when it's purely presentational.

src/css/_BRpages.scss
```diff
@@ -2,50 +2,52 @@
   position: relative;
   margin: auto;
 }
-.BRpagediv1up {
+.BRpagediv {
```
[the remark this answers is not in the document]
shaneriley: Yeah, I was just sticking with the existing naming convention.

src/css/_BRpages.scss
```diff
@@ -2,50 +2,52 @@
...
+    left: 0;
+    z-index: 1;
+    content: "";
+  }
+}
+.BRmodethumb {
   cursor: pointer;
 }
-.BRpagedivthumb a {
+.BRmodethumb a {
```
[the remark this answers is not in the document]
shaneriley: Yeah, they can be moved. I usually prefer that method, but it appeared the project's coding convention was to be as shallow as possible.

src/css/_BRpages.scss
```diff
@@ -2,50 +2,52 @@
...
-  background-color: $brColorPlaceholderBg;
-  cursor: pointer;
-}
 .BRPreview .BRpageimage {
   background-color: #eee;
   -webkit-filter: grayscale(100%);
   filter: grayscale(100%);
 }
 .BRemptypage {
-  background-color: #3a3b3b;
```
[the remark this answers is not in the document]
shaneriley: There's two background-color assignments in this style block. The latter will always be the accepted one, so I dropped the unused one.

shaneriley:
We still want images copiable on non-protected books? I may have to rethink the approach, then.

src/css/_BRpages.scss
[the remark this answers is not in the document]
cdrini: Your call. I think being able to name it is probably most important (e.g. naming it `screen`), so I think at least let's add a comment explaining what this does.

src/css/_BRpages.scss
```diff
@@ -2,49 +2,56 @@
...
+  }
+  &.BRemptypage {
+    background: transparent;
+    img {
+      background-image: none;
+    }
+  }
+  &.BRmodethumb {
+    cursor: pointer;
+    a,
```
cdrini, line 24: This anchor is no longer created.
```suggestion
```

src/css/_BRpages.scss
```diff
@@ -2,49 +2,56 @@
   position: relative;
   margin: auto;
 }
-.BRpagediv1up {
+.BRpagecontainer {
+  position: relative;
   overflow: hidden;
   background-color: $brColorPlaceholderBg;
   box-shadow: 1px 1px 2px #333;
```
cdrini, line 9: This shadow should only be applied to 1up (looks odd in 2up)

src/js/BookReader.js
```diff
@@ -389,18 +389,20 @@ BookReader.prototype.init = function() {
...
+    $(document).on('contextmenu dragstart', '.BRpagecontainer1up', function() {
       return false;
     });
     $(document).on('contextmenu dragstart', '.BRpageimage', function() {
       return false;
     });
-    $(document).on('contextmenu dragstart', '.BRpagedivthumb', function() {
+    $(document).on('contextmenu dragstart', '.BRpagecontainerthumb', function() {
       return false;
     });
```
cdrini, line 400: This can all be deleted now. The context menu appears, but it doesn't give you options to download the image because of the screen.

src/js/BookReader.js
```diff
@@ -389,15 +389,6 @@ BookReader.prototype.init = function() {
   }.bind(this));
 
   if (this.protected) {
-    $(document).on('contextmenu dragstart', '.BRpagediv1up', function() {
```
iisa, line 392: yay!

src/js/BookReader.js
```diff
@@ -581,6 +572,25 @@ BookReader.prototype.drawLeafs = function() {
   }
 };
 
+/**
+ * @private
+ */
+BookReader.prototype._createPageContainer = function(index, styles) {
```
iisa, line 578: We can start here. 👍 
I would like to extend the notion further and create a `Page` class when we touch this code next

PR #297, opened 2020-04-27
Handle URL change/suppression on loading

mc2:
See https://webarchive.jira.com/browse/WEBDEV-3158

WEBDEV-3158


mc2:
ES6 version of https://github.com/internetarchive/bookreader/pull/205
With better e2e tests

src/js/BookReader.js
```diff
@@ -277,7 +283,11 @@ BookReader.prototype.initParams = function() {
     // params explicitly set in URL take precedence over all other methods
     var urlParams = this.paramsFromFragment(this.urlReadFragment());
     if (urlParams.mode) {
-      this.prevReadMode = urlParams.mode;
```
shaneriley, line 280: Why was this assignment removed? I didn't see any corresponding changes to how this property is read or interpreted. It seems like it's used when in thumbnail view and a leaf is clicked/tapped, the previous mode is switched to. Does this omission affect that if BookReader starts in thumbnail mode?

src/js/BookReader.js
```diff
@@ -277,7 +283,11 @@ BookReader.prototype.initParams = function() {
...
+      // If there were any parameters
+      if (Object.keys(urlParams).length > 0) {
```
shaneriley, line 287: **Suggestion**: you could use `length` as a truthy value here and drop the comparison to zero.

src/js/BookReader.js
```diff
@@ -277,7 +283,11 @@ BookReader.prototype.initParams = function() {
...
+        this.extendParams(params, urlParams);
```
shaneriley, line 288: Why is there a second call to `extendParams` here? Just outside of this condition, the same call is made. It looks like you could set `params.fragmentChange = !!Object.keys(urlParams).length` and drop the condition.

tests/e2e/helpers/base.js
```diff
@@ -2,6 +2,36 @@ import { ClientFunction, Selector } from 'testcafe';
...
 
+const getHash = ClientFunction(() => document.location.hash);
+const getUrl = ClientFunction(() => window.location.href);
+
+/**
+ * Check URL page paramter in # and path
+ */
+const expectPage = ClientFunction(() => {
+  const hash = document.location.hash;
+  console.log(hash);
```
shaneriley, line 13: Got a console.log still in here.

tests/e2e/helpers/base.js
```diff
@@ -2,6 +2,36 @@ import { ClientFunction, Selector } from 'testcafe';
...
+    return window.location.href.indexOf('/page/') > -1;
+  }
+});
+
+/**
+ * Check URL mode paramter in # and path
+ *
+ * @param mode '1up', '2up', 'thumb'
+ */
+const expectMode = ClientFunction((mode) => {
```
shaneriley, line 26: The return of this function can lead to a false positive when asserting with `.ok()`, as it's looking for any truthy value and `-1` is truthy in JS.

tests/plugins/plugin.resume.test.js
```diff
@@ -39,17 +39,26 @@ describe('Plugin: Remember Current Page in Cookies', () => {
...
 describe('updateResumeValue', () => {
+  /* Mark 2020-05-02
+    Commenting this out while confirming with Neeraj
+    I'm not sure:
+    - What this test really does
+    - If the instantiation of the prototype counts as a call
+    - If so, why would there be a second call
+    Can't find in production
+    In this branch .toHaveBeenCalledTimes() === 1
   test('starts when BookReaderInit is called', () => {
```
shaneriley, line 50: Rather than have it commented out, it can be skipped with `test.skip`. This will provide an output at the end of the run indicating how many skipped tests were encountered.

src/js/BookReader.js
mc2: done

tests/e2e/helpers/base.js
mc2: removed

tests/e2e/helpers/base.js
mc2: fixed

tests/plugins/plugin.resume.test.js
mc2: done

src/js/BookReader.js
mc2: Thanks for catching this. Don't remember why it was removed
Moved setting this variable to BookReader.prototype.init

src/js/BookReader.js
mc2: fixed

src/js/BookReader.js
```diff
@@ -1397,24 +1412,45 @@ BookReader.prototype.jumpToIndex = function(index, pageX, pageY, noAnimate) {
...
+ */
+BookReader.prototype.switchMode = function(mode, options = {}) {
+  // Set configuration, options overriding defaults
+  const {
+    init = false,
+    suppressFragmentChange = false,
+  } = options;
+
+  // Skip checks when init()
+  if (!init) {
```
iisa, line 1441: we may not need this `!init` check? perhaps we can always return "void" if `(mode === this.mode)` or `!this.canSwitchToMode(mode)`
it would be nice to find a way where `switchMode` has no concept of `init` - and just let it exist with certain rules.
not a show stopper, but if you think that it can be easily done, then i suggest we capture it with at least a ticket.

src/js/BookReader.js
```diff
@@ -1397,24 +1412,45 @@ BookReader.prototype.jumpToIndex = function(index, pageX, pageY, noAnimate) {
...
+    }
   }
 
   this.trigger(BookReader.eventNames.stop);
   if (this.enableSearch) this.removeSearchHilites();
 
-  if (this.mode === this.constMode1up || this.mode === this.constMode2up) {
-    this.prevReadMode = this.mode;
-  }
+  this.prevReadMode = this.getPrevReadMode(this.mode);
```
iisa, line 1453: thanks for simplifying this.

src/js/BookReader.js
```diff
@@ -2570,8 +2606,18 @@ exposeOverrideableMethod(BookModel, '_models.book', '_getDataProp');
  * @param {Object}
  */
 BookReader.prototype.updateFromParams = function(params) {
-  if ('undefined' != typeof(params.mode)) {
-    this.switchMode(params.mode);
+  // Set init, fragment change options for switchMode()
+  const {
+    mode = 0,
```
iisa, line 2611: is `0` a "natural" falsey considering mode is a number?

tests/BookReader.test.js
```diff
@@ -18,3 +37,88 @@ test('has registered view type selected events', () => {
...
+  br.options.enablePageResume = false;
+
+  const params = br.initParams();
+  expect(br.getResumeValue).toHaveBeenCalledTimes(0);
+  expect(params.init).toBe(true);
+  expect(params.index).toBe(0);
+  expect(params.fragmentChange).toBe(false);
+});
+
+test('overrides cookie if page fragment when InitParams called', () => {
```
iisa, line 64: suggestion, clarify `if`

tests/BookReader.test.js
```diff
@@ -18,3 +37,88 @@ test('has registered view type selected events', () => {
...
+test('sets prevReadMode when init called', () => {
+  br.init();
+  expect(br.prevReadMode).toBeTruthy();
+});
+
+test('sets prevPageMode if initial mode is thumb', () => {
+  br.urlReadFragment = jest.fn(() => 'page/n4/mode/thumb');
+
+  br.init();
+  expect(br.prevReadMode).toBe(BookReader.constMode1up);
```
iisa, line 85: question: was this the initial behavior showing a 1up mode if last captured mode was thumb or a new thing?

tests/BookReader.test.js
```diff
@@ -18,3 +37,88 @@ test('has registered view type selected events', () => {
...
+
+test('calls switchMode with init option when init called', () => {
+  br.switchMode = jest.fn();
+
+  br.init();
+  expect(br.switchMode.mock.calls[0][1])
+    .toHaveProperty('init', true);
+});
+
+test('has suppressFragmentChange true when init with no input', () => {
```
iisa, line 96: 🙌

tests/BookReader.test.js
```diff
@@ -18,3 +37,88 @@ test('has registered view type selected events', () => {
...
+  br.urlReadFragment = jest.fn(() => '');
+  br.switchMode = jest.fn();
+
+  br.init();
+  expect(br.switchMode.mock.calls[0][1])
+    .toHaveProperty('suppressFragmentChange', false);
+});
+
+test('has suppressFragmentChange false when init with fragment', () => {
+  br.getResumeValue = jest.fn(() => null);
```
iisa, line 117: q: does this mean that when a user goes to a book with url fragment: `/page/22/mode/2up` - you allow the `fragmentChange` event on init and that sets off its respective cascading changes/updates?

tests/e2e/helpers/base.js
```diff
@@ -63,6 +135,7 @@ export function runBaseTests (br) {
   })
 
   test('2up mode - Clicking `Next page` changes the page', async t => {
+    // Note: this will fail on a R to L book if at front cover
```
iisa, line 138: thanks for the clarifier!

tests/e2e/ia-production/ia-prod-base.js
```diff
@@ -6,7 +6,9 @@ const localURL = 'https://archive.org/details/';
 const books = [
   'birdbookillustra00reedrich', // publicDomain
   // 'pianoservicingtu00rebl_0', // borrowable,
-  // 'adventuresoftoms00twaiiala' // has chapter markers
+  // 'adventuresoftoms00twaiiala', // has chapter markers
+  // 'gendaitankashu00meijuoft', // Right to Left book
+  // 'gov.uspto.patents.application.10074026', // multiple files
```
iisa, line 11: should we open these up soon?

doesn't have to be a part of this pr.

src/js/BookReader.js
mc2: Yes

tests/BookReader.test.js
mc2: yes

tests/e2e/ia-production/ia-prod-base.js
mc2: With help from @jbuckner and Hank comments, we should have most if not all of our production e2e tests run in petabox context. That way we can use these UX group test accounts that we can create and protect. Hopefully the BookReader CI can call these (API?) securely and get the benefit without having them in the BookReader repo.

This is pi in the sky...

tests/BookReader.test.js
mc2: It's already implemented in BookReader.prototype.drawLeafsThumbnail()
I noted with @see in the new function.
I'll be lazy and annotate the test in the next branch.
```
/**
 * Return mode or 1up if initial thumb
 * @param {number}
 * @see BookReader.prototype.drawLeafsThumbnail
 */
BookReader.prototype.getPrevReadMode = function(mode) {
```

src/js/BookReader.js
mc2: you're welcome!

tests/BookReader.test.js
mc2: a little long, but clarified, thanks

src/js/BookReader.js
mc2: No, nope. This is where the psghetti was in the init(). Because in the init() `mode` is always going to equal `this.mode` Before we were setting `this.mode` to `null`in init() after needed for other methods that needed it (but as a class var, not passed) in order to 'prime' this method. Bah!

I think that's psghetti. Things should be explicit and not sneaky magic. That said, sure, there might be a better way. At this point, (months and months,) way not worth it. And my brain would explode. But I'd start by refactoring using params and not class vars.
Thanks!

iisa: >Things should be explicit and not sneaky magic.
+1.

yes, we can let this sit and iterate

Issue #263, opened 2020-03-27
Support immersion mode via a querystring parameter


bfalling:
**Context**

We would like to make BookReader more configurable via URL parameters. One need is to be able to open BookReader directly in immersion mode, sometimes called "full-screen mode" (though it's not actually a full-device screen takeover). (Note this is different from the one-page, two-page, and thumbnail (tiled) display layouts, which are also unfortunately called "mode" as well.)

**Proposal & Constraints**

We need to support a new querystring parameter, whose name is TBD, which will allow opening directly to immersion mode.

**Acceptance Criteria**

Using the parameter, BookReader should be displayed in immersion mode. It should work correctly when used simultaneously with the other URL parameters as well: page number, "mode", search queries, etc.

shaneriley:
When it comes to passing options to BookReader via query params, in the past we've written the param parsing on the Petabox side of things, setting the initial options passed in to BookReader based on the presence of query params. A good example is the `?autoplay=1` param. We have options for `flipSpeed` and `flipDelay` that start up autoplay in BookReader, but conditionally setting those options should occur outside of BookReader.

I have a new option added that will support this feature, and to coincide with this addition I think the Petabox init code should check for an optional query param and pass in the value of that option.

mc2:
I should have discussed this, but currently, BookReader.prototype.initParams gets parameters from 
* options.defaults
* cookie
* fragment
I'm extending this so that there's a params.querystring object to hold anything that comes in through the querystring.

So: 
params.querystring.autoplay =1

shaneriley:
So we are moving the query param handling to native BookReader instead of Petabox init of BookReader?

cdrini:
For now, I think leaving it in petabox makes more sense. Longer term we might want to discuss how to handle this, and longer term we _definitely_ need to clean up BR's options so they're not all on a global scope, but changing the options structure in this way would require a breaking change, which I don't think is worth it right now. I definitely don't think we should do it as part of something else; if we're going to change the way BR handles its options, we should create a separate issue with a proposal, and work on it separate of any other feature request. BR is used in other places; we should be thoughtful about introducing breaking changes.

mc2:
not breaking changes, but if we're going to handle `q=[term]` searching in native BookReader, there has to be query param handling

mc2:
> moving the query param handling to native BookReader instead of Petabox init
No, adding query param handling to native BookReader in addition to Petabox init

PR #313, opened 2020-05-18
Adds startFullscreen option

shaneriley:
This adds a `startFullscreen` boolean to optionally call `enterFullscreen` once BookReader has been properly sized. A demo page and e2e test have been added to illustrate.

Completes #263


tests/e2e/start-fullscreen.test.js
```diff
@@ -0,0 +1,10 @@
+import { ClientFunction, Selector } from 'testcafe';
+import BookReader from './models/BookReader';
+
+const { BASE_URL } = process.env;
+
+fixture `Start fullscreen`.page `${BASE_URL}immersion-mode.html`;
+
+test('starts in full-screen when `startFullScreen: true` set', async t => {
+  await t.expect(Selector('#BookReader').getAttribute('class')).match(/br-ui-full/);
```
cdrini, line 9: This should be do-able as a unit test; can we do it there? E2E tests are kind of expensive.

src/js/BookReader.js
```diff
@@ -360,6 +360,10 @@ BookReader.prototype.init = function() {
     }
   }
 
+  if (this.options.startFullscreen) {
```
cdrini, line 363: Let's add this option to `options.js` with a default of `false`, and a docstring :) That way others can know it exists, since that's our current source of truth (or at least the closest thing we have to a source of truth :D )

src/js/BookReader/options.js
```diff
@@ -202,6 +202,8 @@ export const DEFAULT_OPTIONS = {
       className: 'thumb',
     },
   },
+
+  startFullscreen: false,
```
iisa, line 206: Suggestion: add the docstring

shaneriley: This MR was already merged, but a new one for the fix is [here](https://github.com/internetarchive/bookreader/pull/329) and includes a docstring.

PR #315, opened 2020-05-20
Prefetch should skip unviewable pages

cdrini:
Previously, it would prefetch the unviewable pages as well (as in the dummy page; not the actual uri). Now it only prefetches the first unviewable, and then everything else. Also moved this function to Mode2Up, since it's only used there.

### DIFF
--- a/src/js/BookReader/Mode2Up.js
+++ b/src/js/BookReader/Mode2Up.js
@@ -1171,7 +1171,9 @@ export class Mode2Up {
     const { currentIndexL, currentIndexR } = this.br.twoPage;
     const ADJACENT_PAGES_TO_LOAD = 3;
 
-    let lowPage = book.getPage(min(currentIndexL, currentIndexR));
+    // currentIndexL can be -1; getPage returns the last page of the book
+    // when given -1, so need to prevent that.
+    let lowPage = book.getPage(max(0, min(currentIndexL, currentIndexR)));
     let highPage = book.getPage(max(currentIndexL, currentIndexR));
     for (let i = 0; i < ADJACENT_PAGES_TO_LOAD + 1; i++) {
       if (lowPage) {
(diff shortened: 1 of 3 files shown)


src/js/BookReader/Mode2Up.js
```diff
@@ -1170,7 +1170,9 @@ export class Mode2Up {
     const { currentIndexL, currentIndexR } = this.br.twoPage;
     const ADJACENT_PAGES_TO_LOAD = 3;
 
-    let lowPage = book.getPage(min(currentIndexL, currentIndexR));
+    // currentIndexL can be -1; getPage returns the last page of the book
```
iisa, line 1173: q: is `-1` the value for the pair of either front or back cover? I have seen the front cover spread as this[-1, 0].  Just curious if the last page is like this as well [100, -1]?

tests/BookReader/Mode2Up.test.js
```diff
@@ -38,3 +43,41 @@ describe('zoom', () => {
...
     });
 });
+
+describe('prefetch', () => {
+    test('loads nearby pages', () => {
+        const br = new BookReader({ data: SAMPLE_DATA });
+        br.init();
+        const spy = sinon.spy(br, 'prefetchImg');
+        br.prefetch();
+        expect(spy.callCount).toBeGreaterThan(2);
```
iisa, line 53: so is the idea to just pre-fetch all of the images BUT if book is borrowable, we will try to slot the leafNum first and then open to the image?

src/js/BookReader/Mode2Up.js
cdrini: In mode 2 up, -1 denotes the start [-1,0]. However when I implemented getPage, I made the decision that getPage(-1) should return the back cover :P Forgot that Mode2up did this! So now there are like competing definitions for what -1 means. Not the end of the world, but it would be nice to make these consistent at some point.

tests/BookReader/Mode2Up.test.js
cdrini: Or, this change makes it so that instead of just prefetching the preview place holder image like 5 times :P we only prefetch it once, and then prefetch the next available page (so navigating the book is slick).

PR #316, opened 2020-05-21
Slot resolution

cdrini:
Todo:
- [ ] Add tests
- [ ] Catch fetch errors + add timeout
- [x] Display error page if slot used up
- [x] Switch to using new default page

### DIFF
--- a/src/js/plugins/plugin.search.js
+++ b/src/js/plugins/plugin.search.js
@@ -589,7 +589,7 @@ BookReader.prototype._searchPluginGoToResult = async function (pageIndex) {
     })).then(r => r.json());
 
     for (const leafNum of resp.value) {
-      book.makeViewable(book.leafNumToIndex(leafNum));
+      book.getPage(book.leafNumToIndex(leafNum)).makeViewable();
     }
 
     if (!resp.value.length) {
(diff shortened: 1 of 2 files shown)


src/js/plugins/plugin.search.js
```diff
@@ -553,10 +553,11 @@ BookReader.prototype.addSearchResult = function(queryString, pageIndex) {
             <span class="pageDisplay">${pageDisplayString}</span>
           </tr>
           <tr>
-            <td>
-              <img class="searchImgPreview" src="${imgPreviewUrl}" />
-            </td>
-            <td>
+            ${page.isViewable ?
+              `<td><img class="searchImgPreview" src="${page.getURI(16, 0)}" /></td>` :
```
iisa, line 557: what is the significance of 16 & 0?

src/js/plugins/plugin.search.js
```diff
@@ -570,6 +571,30 @@ BookReader.prototype.addSearchResult = function(queryString, pageIndex) {
   }
 };
 
+/**
+ * @private
+ */
```
iisa, line 576: suggestion: add more context to comment

src/js/plugins/plugin.search.js
cdrini: Fixed :+1:

src/js/plugins/plugin.search.js
cdrini: Added :+1:

cdrini:
Still some jankiness around this specific experience, but good enough for a beta release on IA to get more feedback! Merging. Codecov is apparently out to lunch today.

PR #311, opened 2020-05-12
Trying to supress the error

NayamAmarshe:
Created an empty implementation of window.open to suppress the error.
Closes https://github.com/internetarchive/bookreader/issues/303

PR #319, opened 2020-05-22
Fixing window.alert error

NayamAmarshe:
https://github.com/internetarchive/bookreader/pull/311


cdrini:
Moved commit to #311 :+1:

PR #202, opened 2020-02-17
WIP: Added query parameter q for fts inside book

nsharma123:
This PR is based on https://webarchive.jira.com/browse/WEBDEV-3104

PR #321, opened 2020-05-24
Replace search fragment with q parameter

mc2:
Replaces https://github.com/internetarchive/bookreader/pull/202
* Update to ES6
* Fix querystring added to rather than updated

https://webarchive.jira.com/browse/WEBDEV-3104
Combined with petabox MR 
https://webarchive.jira.com/browse/WEBDEV-3162

### DIFF
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -328,13 +328,28 @@ BookReader.prototype.initParams = function() {
 }
 
 /**
+ * Allow mocking of window.location.search
+ */
+BookReader.prototype.getLocationSearch = function () {
+  return window.location.search;
+}
+
+/**
+ * Allow mocking of window.location.hash
+ */
+BookReader.prototype.getLocationHash = function () {
+  return window.location.hash;
+}
+
+/**
  * Return URL or fragment querystring
  */
 BookReader.prototype.readQueryString = function() {
-  if (window.location.search) {
-    return window.location.search;
+  const queryString = this.getLocationSearch();
+  if (queryString) {
+    return queryString;
   }
-  const hash = window.location.hash;
+  const hash = this.getLocationHash();
   const found = hash.search(/\?\w+=/);
   return found > -1 ? hash.slice(found) : '';
 }
@@ -3013,6 +3028,7 @@ BookReader.prototype.fragmentFromParams = function(params, urlMode = 'hash') {
  *
  * @param {Object} params
  * @param {string} currQueryString
+ * @param {string} [urlMode]
  * @return {string}
  */
 BookReader.prototype.queryStringFromParams = function(
@@ -3024,6 +3040,8 @@ BookReader.prototype.queryStringFromParams = function(
   if (params.search && urlMode === 'history') {
     newParams.set('q', params.search)
   }
+  // https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/toString
+  // Note: This method returns the query string without the question mark.
   const result = newParams.toString();
   return result ? '?' + result : '';
 }
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -302,18 +302,25 @@ BookReader.prototype.initParams = function() {
     }
   }
 
-  /*
-  // If we have a querystring
-  params.queryString = new URLSearchParams(this.readQueryString());
-
-  // Check for Search plugin and term
+  // Check for Search plugin
   if (this.options.enableSearch) {
-    this.options.initialSearchTerm =
-      this.options.initialSearchTerm ??
-      params.queryString.get('q') ??
-      params.search;
+    // If initialSearchTerm not set
+    if (!this.options.initialSearchTerm) {
+      // Look for any term in URL
+      if (params.search) {
+        // Old style: /search/[term]
+        this.options.initialSearchTerm = params.search;
+      } else {
+        // If we have a query string: q=[term]
+        const searchParams = new URLSearchParams(this.readQueryString());
+        const searchTerm = searchParams.get('q')
+        if (searchTerm) {
+          this.options.initialSearchTerm = utils.decodeURIComponentPlus(searchTerm);
+        }
+      }
+    }
   }
-*/
+
   // Set for init process, return to false at end of init()
   this.suppressFragmentChange = !params.fragmentChange;
 
@@ -323,7 +330,6 @@ BookReader.prototype.initParams = function() {
 /**
  * Return URL or fragment querystring
  */
-/*
 BookReader.prototype.readQueryString = function() {
   if (window.location.search) {
     return window.location.search;
@@ -332,7 +338,6 @@ BookReader.prototype.readQueryString = function() {
   const found = hash.search(/\?\w+=/);
   return found > -1 ? hash.slice(found) : '';
 }
-*/
 
 /**
  * Determines the initial mode for starting if a mode is not already
@@ -445,7 +450,7 @@ BookReader.prototype.init = function() {
 
   this.trigger(BookReader.eventNames.PostInit);
 
-  // Return to default
+  // If not searching, set to allow on-going fragment changes
   if (!this.options.initialSearchTerm) {
     this.suppressFragmentChange = false;
   }
@@ -712,7 +717,7 @@ BookReader.prototype.drawLeafsOnePage = function() {
   // Based of the pages displayed in the view we set the current index
   // $$$ we should consider the page in the center of the view to be the current one
   let firstIndexToDraw = indicesToDisplay[0];
-  this.updateFirstIndex(firstIndexToDraw);
+  this.updateFirstIndex(firstIndexToDraw, true);
 
   // if zoomed out, also draw prev/next pages
   if (this.reduce > 1) {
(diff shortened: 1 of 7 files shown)


src/js/BookReader.js
```diff
@@ -302,12 +302,38 @@ BookReader.prototype.initParams = function() {
     }
   }
 
+  /*
+  // If we have a querystring
+  params.queryString = new URLSearchParams(this.readQueryString());
+
+  // Check for Search plugin and term
+  if (this.options.enableSearch) {
```
iisa, line 310: Q: is this meant to be runnable code?

src/js/BookReader.js
```diff
@@ -302,12 +302,38 @@ BookReader.prototype.initParams = function() {
...
   this.suppressFragmentChange = !params.fragmentChange;
 
   return params;
 }
 
+/**
+ * Return URL or fragment querystring
+ */
+/*
+BookReader.prototype.readQueryString = function() {
```
iisa, line 327: same question here - is this meant to be runnable?

src/js/BookReader.js
```diff
@@ -424,7 +450,9 @@ BookReader.prototype.init = function() {
   this.trigger(BookReader.eventNames.PostInit);
 
   // Return to default
-  this.suppressFragmentChange = false;
+  if (!this.options.initialSearchTerm) {
+    this.suppressFragmentChange = false;
```
iisa, line 454: Q: what does the comment `Return to default` mean w/r/t `suppressFragmentChange`?

src/js/BookReader.js
```diff
@@ -2954,10 +2988,9 @@ BookReader.prototype.paramsFromFragment = function(fragment) {
  * @param {Object} params
  * @return {string}
```
iisa, line 2989: Suggestion: add JSDoc for new arg

src/js/BookReader.js
```diff
@@ -2985,13 +3018,34 @@ BookReader.prototype.fragmentFromParams = function(params) {
...
+  params,
+  currQueryString,
+  urlMode = 'hash'
+) {
+  const newParams = new URLSearchParams(currQueryString);
+  if (params.search && urlMode === 'history') {
+    newParams.set('q', params.search)
+  }
+  const result = newParams.toString();
+  return result ? '?' + result : '';
```
iisa, line 3046: Q: do you need to append the `?` before the result?

src/js/plugins/plugin.search.js
```diff
@@ -282,6 +291,8 @@ BookReader.prototype.BRSearchCallback = function(results, options) {
   if (firstResultIndex !== null) {
     this._searchPluginGoToResult(firstResultIndex);
   }
+  // Reset
+  // this.suppressFragmentChange = false;
```
iisa, line 295: Q: do we need this one here?

src/js/plugins/plugin.search.js
```diff
@@ -611,11 +622,13 @@ BookReader.prototype._searchPluginGoToResult = async function (pageIndex) {
...
  * Removes all search pins
  */
-BookReader.prototype.removeSearchResults = function() {
+BookReader.prototype.removeSearchResults = function(suppressFragmentChange = false) {
   this.removeSearchHilites(); //be sure to set all box.divs to null
   this.searchTerm = null;
   this.searchResults = null;
-  this.trigger(BookReader.eventNames.fragmentChange);
+  if (!suppressFragmentChange) {
+    this.trigger(BookReader.eventNames.fragmentChange);
```
iisa, line 630: so, unless we flip it true, this will always fire, ya?

src/js/BookReader.js
mc2: Ah. Took me a while to figure out. This was commented out for recent deploy but not committed yet. This would allow q=[term] in standalone BookReader. Putting it back in and testing. Sorry about the confusion.

src/js/BookReader.js
mc2: As above

src/js/BookReader.js
mc2: improved comment:
`// If not searching, set to allow on-going fragment changes`

src/js/BookReader.js
mc2: Done.

src/js/BookReader.js
mc2: Y: URLSearchParams.toString() returns query string with no preceding '?'

src/js/plugins/plugin.search.js
mc2: Oops. Test code. Good catch! Removed.

src/js/plugins/plugin.search.js
mc2: Yes.

src/js/BookReader.js
iisa: yah. i saw the string interpolation later in my review. ok 👍

src/js/plugins/plugin.search.js
```diff
@@ -154,20 +157,26 @@ BookReader.prototype.search = function(term, options) {
...
 
-  this.removeSearchResults();
+  this.removeSearchResults(options.suppressFragmentChange);
 
   this.searchTerm = term;
   // strip slashes, since this goes in the url
   this.searchTerm = this.searchTerm.replace(/\//g, ' ');
 
-  this.trigger(BookReader.eventNames.fragmentChange);
+  if (! options.suppressFragmentChange) {
```
bfalling, line 177: SUGGESTION: Remove space after `!`

src/js/plugins/plugin.url.js
```diff
@@ -137,10 +137,12 @@ BookReader.prototype.urlUpdateFragment = function() {
...
   }, {});
 
-  const newFragment = this.fragmentFromParams(params);
+  const newFragment = this.fragmentFromParams(params, urlMode);
   const currFragment = this.urlReadFragment();
+  const currQueryString = window.location.search;
+  const newQueryString = this.queryStringFromParams(params, currQueryString, urlMode);
 
-  if (currFragment === newFragment) {
+  if (currFragment === newFragment && currQueryString === newQueryString) {
```
bfalling, line 145: QUESTION: This is the critical comparison to determine whether the URL should be rewritten. How dependent is it on parameter order? Is it smart enough to still do a correct match if it's the same parameters but in a different order? If not, should it be?

And: Should there be unit tests to check this?

mc2: * dependent is it on parameter order? yes
* smart enough to match if different order? no
* should it be? no, according to spec: https://url.spec.whatwg.org/#interface-urlsearchparams, internal 'List' is ordered, and original query string used
* should we check for out-of-spec implementations? no

Reviewing the code I can see a few places to add units tests, they'd test implicitly for order.

src/js/plugins/plugin.search.js
mc2: done

src/js/plugins/plugin.url.js
mc2: tests added

src/js/BookReader.js
```diff
@@ -302,12 +302,58 @@ BookReader.prototype.initParams = function() {
     }
   }
 
+  // Check for Search plugin
+  if (this.options.enableSearch) {
+    // If initialSearchTerm not set
+    if (!this.options.initialSearchTerm) {
+      // Look for any term in URL
+      if (params.search) {
+        // Old style: /search/[term]
```
iisa, line 311: Q: at this point - will we even have `/search/term` in URL?

src/js/BookReader.js
```diff
@@ -302,12 +302,58 @@ BookReader.prototype.initParams = function() {
...
+        this.options.initialSearchTerm = params.search;
+      } else {
+        // If we have a query string: q=[term]
+        const searchParams = new URLSearchParams(this.readQueryString());
```
iisa, line 315: Q: is there a reason why we aren't adding it to `params.search` ?

src/js/BookReader.js
```diff
@@ -2985,13 +3041,36 @@ BookReader.prototype.fragmentFromParams = function(params) {
...
 
+/**
+ * Create, update querystring from the params object
+ *
+ * @param {Object} params
+ * @param {string} currQueryString
+ * @param {string} [urlMode]
+ * @return {string}
+ */
+BookReader.prototype.queryStringFromParams = function(
```
iisa, line 3059: Question:  This new core code seems to exclusively munge the URL - can we put this in the URL Plugin?

src/js/BookReader.js
iisa: will `params` be out of sync?

mc2:
Replaced by https://github.com/internetarchive/bookreader/pull/340

src/js/BookReader.js
mc2: >Q: at this point - will we even have /search/term in URL?
in the hash URL styles, yes
> will params be out of sync in any way?
No, not resetting

src/js/BookReader.js
mc2: This is an init case, don't need it. options.initialSearchTerm is what's needed
This particular line creates a URLSearchParams object

src/js/BookReader.js
mc2: We'll look at this later

PR #328, opened 2020-06-01
Search inside cleanup

iisa:
**1) search term highlight does not show when the page is unviewable**

- Go to: https://www-isa.archive.org/details/goldengalleonsof0000grah
- inside book search, input & initiate search for: `island`
- pins drop -> click around -> ...
  - ... -> if page is viewable, search is highlighted
  - ... -> if page isn't viewable, "try again later" placeholder is visible with no highlight

ezgif com-video-to-gif

**2) Both mobile & desktop search inputs sync in value on search term change**

- Go to: https://www-isa.archive.org/details/goldengalleonsof0000grah
- inside book search, input & initiate search for: `island`
- shrink viewport width until it's in mobile view
- open menu 🍔  -> then click `Search` panel
- In search term form, input & initiate on term `shipwreck`
- no results, cool
- grow viewport width until it is "desktop" size again
- desktop input should also have `shipwreck` as value

ezgif com-video-to-gif (2)

**3) Mobile menu closes when a panel is open and the viewport stretches to desktop**
- Go to: https://www-isa.archive.org/details/goldengalleonsof0000grah
- inside book search, input & initiate search for: `island`
- shrink viewport width until you get mobile menu
- open menu 🍔  -> then click `Search` panel
- keep panel open
- grow viewport width until it is "desktop" size again
- Mobile menu closes properly.

ezgif com-video-to-gif (1)

### DIFF
--- a/src/js/plugins/plugin.mobile_nav.js
+++ b/src/js/plugins/plugin.mobile_nav.js
@@ -9,6 +9,8 @@
  * <link rel="stylesheet" href="../BookReader/mmenu/dist/addons/navbars/jquery.mmenu.navbars.css" />
  */
 
+import * as utils from '../BookReader/utils.js';
+
 jQuery.extend(BookReader.defaultOptions, {
   enableMobileNav: true,
   mobileNavTitle: 'Internet Archive',
@@ -103,7 +105,7 @@ BookReader.prototype.initToolbar = (function (super_) {
         }
       });
 
-      window.addEventListener('resize', (e) => {
+      const closeMobileMenu = (e) => {
         // Need to close the mobile menu to reset DOM & Style
         // driven by menu plugin
         const width = $( window ).width();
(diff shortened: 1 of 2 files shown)


jbuckner:
Tested in Safari and worked as expected. Code looks good 👍

src/js/plugins/plugin.mobile_nav.js
```diff
@@ -102,6 +102,17 @@ BookReader.prototype.initToolbar = (function (super_) {
           $mmenuEl.data('mmenu').open();
         }
       });
+
+      window.addEventListener('resize', (e) => {
```
shaneriley, line 106: With window resize events, it's usually a good idea to throttle how often the callback is fired to avoid an impact on performance. I'm not sure if there's already a function to do so, but the function from Petabox is pretty standard and light.

src/js/plugins/plugin.mobile_nav.js
```diff
@@ -102,6 +104,19 @@ BookReader.prototype.initToolbar = (function (super_) {
...
+        // driven by menu plugin
+        const width = $( window ).width();
+        const mobileMenuIsOpen = $mmenuEl.data('mmenu').getInstance().vars.opened;
+        // $brBreakPointMobile: 800px;
+        if (mobileMenuIsOpen && (width >= 800)) {
+          $mmenuEl.data('mmenu').close ();
+        }
+      };
+
+      window.addEventListener('resize', utils.debounce(closeMobileMenu, 900));
```
shaneriley, line 119: oh_yeah

PR #329, opened 2020-06-01
Start fullscreen post resize

shaneriley:
This fixes an issue where fullscreen was started before the pages were initially resized to fit the BookReader container and a corresponding class was removed to display the images. A new demo page was added to illustrate the ability to enable fullscreen mode on init for items that start in a mode other than 2up.

### DIFF
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -386,13 +386,13 @@ BookReader.prototype.init = function() {
     }
   }
 
-  if (this.options.startFullscreen) {
-    this.enterFullscreen();
-  }
-
   this.resizeBRcontainer();
   this.updateFromParams(params);
   this.initUIStrings();
+
+  if (this.options.startFullscreen) {
+    this.enterFullscreen();
+  }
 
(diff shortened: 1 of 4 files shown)


src/js/BookReader.js
```diff
@@ -427,6 +423,12 @@ BookReader.prototype.init = function() {
   this.suppressFragmentChange = false;
 
   this.init.initComplete = true;
+
+  // Must be called after this.init.initComplete set to true to allow
+  // BookReader.prototype.resize to run.
+  if (this.options.startFullscreen) {
```
iisa, line 429: ok 👍 thanks for the comment and context.

src/js/BookReader/options.js
```diff
@@ -203,6 +203,10 @@ export const DEFAULT_OPTIONS = {
     },
   },
 
+  /**
+   * @type {Boolean}
+   * Optional: if true, starts in fullscreen mode
+   */
```
iisa, line 209: thanks

PR #346, opened 2020-06-04
Mock SearchInsideAccess to fix IA demos

cdrini:
These apparently need some SearchInsideAccess.js which is not being imported by the PHP

cc: @iisa


cdrini:
Yaaay travis passed! \o/ Not waiting for codecov

PR #356, opened 2020-06-15
Generic demo

### LINKED ISSUE #353: Consolidate demo-ia-plato and demo-ia-olivertwist
These are largely identical demos that differ only in the OCAID specified. Note the oliver twist demo also uses `/` style urls instead of `#`, but these don't work properly with the demo servers, so let's just use the `#` style URLs for the demos.

I recommend using `document.writeln` as much as possible to make this simple.

## Summary of Requirements
- [ ] URL parameter, `&ocaid=FOO` that can be used to specify/use any OCAID
- [ ] BookReaderDemo/index.html still lists both as demos, but with updated URLs
- [ ] E2E tests should use the new scheme

Pyrojet99:
closes #353 
Creation of a generic demo html file which uses pages of a specified book from IA servers based on its OCAID.
E2E tests use dem-generic.html instead of the specific files


src/js/plugins/plugin.url.js
```diff
@@ -145,6 +145,13 @@ BookReader.prototype.urlUpdateFragment = function() {
     return;
   }
 
+  //filtering query parameters to select only book search param (?q=foo)
+  //this needs to be updated/URL system modified if future query params are to be added
+  const queries = newQueryString.replace("?","").split("&");
+  const searchQueryIndx = queries.findIndex((el) => el.match(/^q=/));
+  const newQueryStringSearch = searchQueryIndx == -1 ? "" : queries[searchQueryIndx];
```
cdrini, line 152: Refactor this to use `URLSearchParams` instead of manual parsing, and do this inside the `else` clause below (since that's the only place it's needed).

tests/e2e/base.test.js
```diff
@@ -6,8 +6,8 @@ import { runMobileSearchTests } from './helpers/mobileSearch';
 
 
 const { BASE_URL } = process.env;
-const localPages = [
-  'demo-ia-plato.html',
+const idBooks = [
```
cdrini, line 9:
```suggestion
const OCAIDS = [
```

tests/e2e/base.test.js
```diff
@@ -16,18 +16,19 @@ const localPages = [
   // /BookReaderDemo/demo-ia-olivertwist.html/page/n13/mode/2up
   // 'demo-ia-olivertwist.html',
 ];
+const genericDemo = 'demo-generic.html?ocaid='
```
cdrini, line 19:
```suggestion
const DEMO_PATH = 'demo-generic.html?ocaid='
```

tests/e2e/base.test.js
```diff
@@ -16,18 +16,19 @@ const localPages = [
...
 
-localPages.forEach(function(page) {
-  const url = `${BASE_URL}${page}`;
+idBooks.forEach(function(book) {
```
cdrini, line 21:
```suggestion
idBooks.forEach(ocaid => {
```

BookReaderDemo/index.html
```diff
@@ -25,6 +25,8 @@
       <li><a href="demo-plugin-menu-toggle.html">Plugin: Full screen menu toggle</a></li>
       <li><a href="immersion-mode.html">Start in immersion (fullscreen) mode</a></li>
       <li><a href="immersion-1up.html">Start in immersion mode on 1up default item</a></li>
+      <li><a href="demo-generic.html?ocaid=theworksofplato01platiala">demo-ia-plato with generic demo</a></li>
+      <li><a href="demo-generic.html?ocaid=adventuresofoli00dick">demo-ia-olivertwist with generic demo</a></li>
```
cdrini, line 29: Replace the previous files, `demo-ia-plato` and `demo-ia-olivertwist` with these links.

BookReaderDemo/demo-generic.html
```diff
@@ -0,0 +1,80 @@
...
+    
+    <link rel="stylesheet" href="BookReaderDemo.css"/>
+</head>
+<body>
+
+<div id="BookReader">
+    Internet Archive BookReader Demo<br/>
+    <noscript>
+    <p>
+        The BookReader requires JavaScript to be enabled. Please check that your browser supports JavaScript and that it is enabled in the browser settings.  You can also try one of the <a href="https://archive.org/details/theworksofplato01platiala"> other formats of the book</a>.
```
cdrini, line 46:
```suggestion
        The BookReader requires JavaScript to be enabled. Please check that your browser supports JavaScript and that it is enabled in the browser settings.
```

BookReaderDemo/demo-generic.html
```diff
@@ -0,0 +1,80 @@
+<!DOCTYPE html>
```
cdrini, line 1: Rename this to `demo-internetarchive`

BookReaderDemo/demo-generic.html
```diff
@@ -0,0 +1,80 @@
...
+// Temporarily fix build by mocking SearchInsideAccess; long term fix likely needs to be
+// on petabox side
+function SearchInsideAccess(br, identifier, callback) { this.init = callback; }
+</script>
+<script
+  id="pageUrl"
+  type="text/javascript"
+></script>
+
+<script>
```
cdrini, line 64: Clean this up.

src/js/plugins/plugin.url.js
cdrini, line 152: Also, let's add a unit test for this.

src/js/plugins/plugin.url.js
```diff
@@ -154,12 +154,27 @@ BookReader.prototype.urlUpdateFragment = function() {
...
+ * filtering query parameters to select only book search param (?q=foo)
+   this needs to be updated/URL system modified if future query params are to be added
+ * @param {string} url
+ * @return {string}
+ * */
+BookReader.prototype.urlParamsFiltersOnlySearch = function(url) {
+  const queries = new URLSearchParams(url);
+  const searchedTerm = queries.get('q');
+  const newQueryStringSearch = searchedTerm == null ? "" : 'q=' + searchedTerm;
+  return newQueryStringSearch;
```
cdrini, line 174: This might break if the query has special characters in it (like `&`), since `.get(...)` returns the _unencoded_ parameters. Add a unit test for it; using `URLSearchParams#toString` should make sure we're good :+1:

```suggestion
  const params = new URLSearchParams(url);
  return params.has('q') ? new URLSearchParams({ q: params.get('q') }).toString() : '';
```

src/js/plugins/plugin.url.js
```diff
@@ -154,12 +154,27 @@ BookReader.prototype.urlUpdateFragment = function() {
...
   } else {
-    window.location.replace('#' + newFragment + newQueryString);
+    const newQueryStringSearch = this.urlParamsFiltersOnlySearch(this.readQueryString());
+    window.location.replace('#' + newFragment + newQueryStringSearch);
   }
 
   this.oldLocationHash = newFragment;
 };
 
+/**
```
cdrini, line 164:
```suggestion
/**
@private
```

src/js/plugins/plugin.url.js
Pyrojet99: DONE

src/js/plugins/plugin.url.js
Pyrojet99: DONE

BookReaderDemo/demo-internetarchive.html
```diff
@@ -56,10 +57,16 @@
...
+
+<script id="pageUrl" type="text/javascript"></script>
+
+<script>
+  url1 = "https://archive.org/bookreader/BookReaderJSLocate.php?id=";
+  url2 = "&subPrefix=";
+  currentUrl = new URL(window.location.href);
+  bookId = currentUrl.searchParams.get("ocaid");
+  document.getElementById('pageUrl').src = url1 + bookId + url2;
+</script>
```
cdrini, line 69: Use `var` + reduce some of the helper variables.

tests/plugins/plugin.url.test.js
```diff
@@ -120,4 +120,13 @@ describe('Plugin: URL controller', () => {
     expect(window.history.replaceState).toHaveBeenCalledTimes(0);
   })
 
+  test('only q= param is selected from url query params', () => {
+    const INTIAL_URL="http://127.0.0.1:8080/BookReaderDemo/demo-internetarchive.html?ocaid=adventuresofoli00dick&q=foo"
+    br.options.urlMode = 'hash';
+    br.init();
+    const result = br.urlParamsFiltersOnlySearch(INTIAL_URL);
+    const expectedRes = new URLSearchParams({ q: "foo" }).toString();
+    expect(result).toBe(expectedRes);
```
cdrini, line 129: In general, unit tests should be kind of "dumb"; just using flat data. Otherwise you end up with sort of a tautology. Do this without `URLSearchParams`, and have another test to handle a url with a special character in it's `q` (e.g. `?q=foo%20boo`)

cdrini: Actually, the function is essentially a static function (since it doesn't use `this`, so you can call it as `BookReader.prototype.urlParamsFiltersOnlySearch('https://...')` without having to initialize BR in any way.

src/js/plugins/plugin.url.js
```diff
@@ -154,12 +154,26 @@ BookReader.prototype.urlUpdateFragment = function() {
...
+ * @private
+ * Filtering query parameters to select only book search param (?q=foo)
+   This needs to be updated/URL system modified if future query params are to be added
+ * @param {string} url
+ * @return {string}
+ * */
+BookReader.prototype.urlParamsFiltersOnlySearch = function(url) {
+  const params = new URLSearchParams(url);
+  return params.has('q') ? new URLSearchParams({ q: params.get('q') }).toString() : '';
```
cdrini, line 173: This should include the `?` character (+ update tests)

```suggestion
  return params.has('q') ? `?${new URLSearchParams({ q: params.get('q') })}` : '';
```

PR #359, opened 2020-06-17
Use browser's path for br-resume coookie

mc2:
For multiple files in items, cookie paths written by code get broken and don't match with browser.
Use what the browser sees as the path, trimming any /page/... or /mode/...  
**Note:** Required for multiple files in items petabox code

https://webarchive.jira.com/browse/WEBDEV-3542

**Solution Summary:**
- Implement regex trimming `/path/` or `/mode/` from `location.pathname` to use for cookie path
- Wrap in method for testability
- Refactor legacy test using `.call()` and breaking `this`
- Add unit tests of new wrapped regex

**Testing**
Should set/get cookies perfectly, independent of new petabox code to handle multiple files in items


src/js/plugins/plugin.resume.js
```diff
@@ -40,14 +40,29 @@ BookReader.prototype.getResumeValue = function() {
...
- * Set's the page resume value, for remembering reader's page
+ * Return cookie path using pathname up to /page/... or /mode/...
+ * using window.location.pathname for urlPathPart:
+ * - matches encoding
+ * - ignores querystring part
+ * - ignores fragment part (after #)
+ * @param {string} urlPathPart - window.location.pathname
+ */
+BookReader.prototype.getCookiePath = function(urlPathPart) {
+  return urlPathPart.match('.+?(?=/page/|/mode/|$)')[0];
```
iisa, line 51: Thanks for the tests!
Q: is `page|mode` the only parts of the URL that immediately proceed the item identifier?

tests/plugins/plugin.resume.test.js
```diff
@@ -66,11 +66,40 @@ describe('updateResumeValue', () => {
...
+    expect(setItemSpy.args[0][3]).toEqual('/details/foo');
+    // Restore function
+    br.getCookiePath = saveFn;
+  });
 
-    updateResumeValue.call(fakeBr, 16);
-    expect(setItemSpy.args[0][3]).toEqual('/');
+  test('handles cookie path from URL with decoration', () => {
+    const complexPathWithPage = '/details/2008ELMValidityStudyFinalReportRevised/Executive%20Summary%20for%20the%20EPT%26ELM%20Validity%20Studie_20100603%20-%20Copy/page/n1/mode/2up';
+    const complexPath = '/details/2008ELMValidityStudyFinalReportRevised/Executive%20Summary%20for%20the%20EPT%26ELM%20Validity%20Studie_20100603%20-%20Copy';
```
iisa, line 81: thanks again.

src/js/plugins/plugin.resume.js
mc2: Yes, for now. '/search/ used to be there but no longer.

PR #369, opened 2020-07-09
Fix toc entries in desktop missing page num

cdrini:


cdrini:
Auto-merging this since it's a small fix, so I can get a deploy out today.

PR #373, opened 2020-07-14
Refactor: simpler webpack.config.js

cdrini:
We'd over-abstracted this; we can just use a basic setup :)

This will also make it easier to dedupe the shared code inside BookReader once we go to webpack 5 (still in beta :( )! https://webpack.js.org/configuration/entry-context/#dependencies


cdrini:
Thanks @shaneriley !

PR #375, opened 2020-07-15
Creates view mode cycler

shaneriley:
This replaces the individual view mode buttons with a single button that cycles through the available view modes. An update was also made to the method of disabling switching to particular view modes.

Example of new interaction:

view_mode_switch


src/js/BookReader.js
```diff
@@ -2149,99 +2151,67 @@ BookReader.prototype.createPopup = createPopup;
...
+      if ($.inArray(this.mode, [this.constMode1up, this.constModeThumb]) >= 0) {
+        this.scrollDown();
+      } else {
+        this.next();
+      }
+    },
+    book_top: this.first,
+    book_bottom: this.last,
+    book_leftmost: this.leftmost,
+    book_rightmost: this.rightmost,
```
cdrini, line 2184: Will these work? We might need to do `() => this.first()` so that `this` is bound correctly?

src/js/BookReader.js
```diff
@@ -2149,99 +2151,67 @@ BookReader.prototype.createPopup = createPopup;
...
+        window.open(url);
+      } else {
+        this.toggleFullscreen();
+      }
+    },
+  };
 
   jIcons.filter('.fit').bind('fit', function() {
     // XXXmang implement autofit zoom
   });
```
cdrini, line 2207: Let's copy into the new object as is (with the comment) for consistency :+1:

src/js/BookReader.js
```diff
@@ -2149,99 +2151,67 @@ BookReader.prototype.createPopup = createPopup;
...
-
-  jIcons.filter('.twopg').bind('click', function() {
-    self.switchMode(self.constMode2up);
-  });
-
-  jIcons.filter('.thumb').bind('click', function() {
-    self.switchMode(self.constModeThumb);
-  });
+  const jIcons = this.$('.BRicon').add('.BRmobileMenu .BRicon');
+  const navigationControls = {
```
cdrini, line 2158:
```suggestion
  // Map of jIcon class -> click handler
  const navigationControls = {
```

src/js/BookReader/options.js
```diff
@@ -189,17 +189,10 @@ export const DEFAULT_OPTIONS = {
   getEmbedCode: null,
 
   controls: {
-    onePage: {
+    viewmode: {
```
cdrini, line 192: Ahhh, this is a breaking change 😓 Is there a way we could use the old syntax here? And then in the code e.g. `if (controls?.onePage?.visible) { ... }`

src/js/BookReader/Navbar/Navbar.js
```diff
@@ -24,6 +24,55 @@ export class Navbar {
     return `<button class="BRicon ${option.className} desktop-only js-tooltip"></button>`;
   }
 
+  bindViewModeButton() {
```
cdrini, line 27:
```suggestion
  /** @private */
  _bindViewModeButton() {
```

src/js/BookReader.js
shaneriley: Tested them out, and they do need to have their context bound. Forgot I'd be in a jQuery context once I called it from within a jQuery callback.

src/js/BookReader.js
shaneriley: All of the events on the `navigationControls` object are bound to a click event. This one was a custom 'fit' event, so I left it on its own.

src/js/BookReader/options.js
shaneriley: I was trying to retain the option structure, but the issue I ran into was that I'd have to add properties to the existing control objects to know which controls to remove from the array when the viewmode button is created. Since this control object allows you to override every property, there was no way to map the particular control to a view mode.

This option was added specifically for the audio player's liner notes BR instance, and when Petabox is upgraded to use this version, the options will change to achieve the same result. Beyond that, I was developing with Brenton's comment of ultimately developing this project for our own use in the back of my mind. If we need to make breaking changes so we can develop the features we want, then that's what we do.

On the other hand, to make the old options work and the new viewmode switcher work, it could be an either/or option. If `viewmode` is one of the controls, this ignores the other individual view mode control options. That will take a little work and some conditional logic in the Navbar template, but it could be done.

shaneriley:
Relevant test page: https://deploy-preview-375--lucid-poitras-9a1249.netlify.app/bookreaderdemo/viewmode-cycle.html

src/js/BookReader/options.js
cdrini: :+1: The current solution lgtm!

PR #407, opened 2020-08-03
Make testcafe run concurrently in Travis

cdrini:


cdrini:
On a sample size of 1, cut like 20s of the testcafe command's time.

cdrini:
Sample size 2: ~25s cut. Good enough!

PR #412, opened 2020-08-07
Create npm-publish.yml

cdrini:


cdrini:
:D Hahaha yep! The release flow is still the same; after we click that 'fun' link, this action will automatically publish the newly created release to npm! 🎉

PR #419, opened 2020-08-13
Bookreader: do not cycle to back of book if at cover page

iisa:
### Bookreader: do not cycle to back of book if at cover page
Currently, if you are at the first page of a book, and click the “previous page” button, you will cycle to the back of the book.
But if you go to the back cover, and click “next page”, you do not cycle back to the front.
We want to not cycle at either ends of the book.

#### Testing
Go to each of these books and confirm:
- if at book front cover && you click on “previous page” => nothing happens, you stay at front cover
- if at book back cover && you click on “next page” => nothing happens, you stay at back cover
- you can click forward or backward as usual
- you can skip through the book using the book width as usual

Links:
- stream: https://www-isa2.archive.org/stream/fishhatcherymana00pipe
- borrowable: https://www-isa2.archive.org/details/wolfganggoethe1ran
- right to left: https://www-isa2.archive.org/details/gendaitankashu00meijuoft
- br netlify: https://deploy-preview-419--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala#mode/2up


BookReader/BookReader.js
```diff
@@ -1,2 +1,2 @@
...
...Imgs[o]).animate({width:"0px"},this.br.flipSpeed,"easeInSine",(function(){$(r.br.leafEdgeTmp).animate({left:"".concat(h+c,"px")},r.br.flipSpeed,"easeOutSine"),r.br.$(".BRgutter").css({left:"".concat(h-.5*r.br.twoPage.bookSpineDivWidth,"px")}),$(r.br.prefetchedImgs[t]).animate({width:"".concat(c,"px")},r.br.flipSpeed,"easeOutSine",(function(){$(r.br.prefetchedImgs[e]).css("zIndex",2),$(r.br.prefetchedImgs[e]).css("display",""),$(r.br.prefetchedImgs[t]).css("display",""),$(r.leafEdgeR).css({width:"".concat(r.br.twoPage.edgeWidth-i,"px"),left:"".concat(h+c,"px")}),$(r.leafEdgeL).css({width:"".concat(i,"px"),left:"".concat(h-u-i,"px")}),$(r.br.twoPage.coverDiv).css({width:"".concat(r.coverWidth(u+c),"px"),left:"".concat(h-u-i-r.br.twoPage.coverInternalPadding,"px")}),$(r.br.leafEdgeTmp).remove(),r.br.leafEdgeTmp=null,r.br.twoPage.currentIndexL=e,r.br.twoPage.currentIndexR=t,r.br.twoPage.scaledWL=u,r.br.twoPage.scaledWR=c,r.br.twoPage.gutter=h,r.br.updateFirstIndex(r.br.twoPage.currentIndexL),r.br.displayedIndices=[e,t],r.br.pruneUnusedImgs(),r.br.prefetch(),r.br.animating=!1,r.br.enableSearch&&r.br.updateSearchHilites(),r.setMouseHandlers(),r.br.animationFinishedCallback&&(r.br.animati
```
iisa, line 1: compiled to test in pb

tests/e2e/helpers/base.js
```diff
@@ -104,16 +116,18 @@ export function runBaseTests (br) {
   test('2up mode - Clicking `Previous page` changes the page', async t => {
     const { nav, BRcontainer} = br;
 
+    const goPrevious = await isRightToLeft() ? nav.desktop.goNext : nav.desktop.goPrevious;
```
shaneriley, line 119: These two ternaries are already used a few times, and I suspect we'll need them more often. Can we abstract this to methods that allow us to request the next or previous elements? Then internally it can determine whether to send the left or right direction flip button.

src/js/BookReader.js
```diff
@@ -1812,6 +1812,9 @@ BookReader.prototype.next = function() {
 };
 
 BookReader.prototype.prev = function() {
+  const isOnFrontPage = this.firstIndex < 1;
```
jbuckner, line 1815: Nice use of an intermediate variable here. It gives context to this logic. 👏

tests/e2e/helpers/base.js
iisa: yes, absolutely. and i think we need to rename the nav buttons (ex. `goNext` => `goRight`) as well in order to make this happen.
i'd like to follow that work behind this PR so we can properly do it and won't balloon this task out of scope.

i do think it is absolutely necessary to do as well,

iisa: hey @shaneriley lmk if we are still ok with moving that work to its own separate PR - i'd like to get this into master branch before embarking so i don't have to fork.

shaneriley: Yeah, we can defer to a separate task.

PR #420, opened 2020-08-17
Splits search plugin controller and view, updates mobile menu styles

shaneriley:
In order to make it easier to eventually replace the search menu HTML and CSS with a web component, the view needed to be split from the search plugin controller. This creates a new file, plugin.search.view.js, that is imported by the search plugin and provides a view constructor for the search form and search results. New triggered events were added to the search plugin to make this possible, and are documented at the top of the search plugin controller.

To match the styling of the new search menu and recent changes to the styling of the nav bar, the mobile menu and its icons have been updated to match. In the process of updating the icons, I've cleaned out all of the unnecessary code generated by Sketch and added ARIA labels.

See https://webarchive.jira.com/browse/WEBDEV-3688 (search plugin split/refactor) https://webarchive.jira.com/browse/WEBDEV-3689 (mobile menu redesign) for more info and for a video demo of the redesigned mobile nav.

Testing can be done on https://www-shaner.archive.org


src/js/plugins/plugin.search.js
```diff
@@ -330,37 +239,16 @@ BookReader.prototype.BRSearchCallbackErrorMobile = function(results) {
...
  */
-BookReader.prototype._BRSearchCallbackError = function(results, $el, fade) {
-  this.$('.BRnavpos .search').remove();
-  this.$('.BRmobileSearchResultWrapper').empty(); // Empty mobile results
-
+BookReader.prototype._BRSearchCallbackError = function(results) {
   this.searchResults = results;
-  let timeout = 2000;
   if (results.error) {
-    if (/debug/.test(window.location.href)) {
```
iisa, line 340: ahh the things we squirrel away...

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,317 @@
+class SearchView {
+  constructor(params) {
+    if (!params.selector) {
+      console.warn('BookReader::Search - SearchView must be passed a valid CSS selector');
+      return;
+    }
+
+    this.br = params.br;
+
+    this.matcher = new RegExp('{{{(.+?)}}}', 'g');
```
iisa, line 10: suggestion: add comment and example for what we are matching for

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,317 @@
...
+    this.cacheDOMElements();
+    this.bindEvents();
+
+    if (this.br.options.initialSearchTerm) {
+      this.setQuery(params.query);
+    }
+  }
+
+  cacheDOMElements() {
+    this.dom.results = this.dom.searchTray.querySelector('[data-id="results"]');
```
iisa, line 23: Q: are these the full list of search dom?

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,317 @@
...
+  }
+
+  clearSearchFieldAndResults() {
+    this.toggleResultsCount(false);
+    this.removeResultPins();
+    this.emptyMatches();
+    this.setQuery('');
+  }
+
+  renderSearchTray(selector) {
```
iisa, line 64: ah ok, i see why you sent the selector string through instead of the dom node.

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,317 @@
...
+              tooltip.style.setProperty('transform', `translateX(-${targetOffset.left - boxSizeAdjust}px)`);
+            }
+            $('.BRsearch,.BRchapter').removeClass('front');
+            $(event.target).addClass('front');
+          },
+          (event) => $(event.target).removeClass('front'))
+        .click(function (event) {
+          // closures are nested and deep, using an arrow function breaks references.
+          // Todo: update to arrow function & clean up closures
+          // to remove `bind` dependency
```
iisa, line 194: ok sounds good.

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,317 @@
...
+    });
+
+    $(this.dom.results).on('click', 'li', (e) => {
+      this.br._searchPluginGoToResult(+e.currentTarget.dataset.pageIndex);
+      this.br.updateSearchHilites();
+    });
+  }
+}
+
+export default SearchView;
```
iisa, line 317: what a beaut.

tests/e2e/helpers/mobileSearch.js
```diff
@@ -29,10 +29,11 @@ export function runMobileSearchTests(br) {
 
       //testing successful search
       await t
-        .selectText(nav.searchBox.child('.BRsearchInput'))
+        .selectText(nav.searchBox.find('[name="query"]'))
```
iisa, line 32: perhaps we still need this to be a bit more specific? do both inputs have that name?

src/js/BookReader.js
```diff
@@ -2648,31 +2650,18 @@ BookReader.prototype.updateFromParams = function(params) {
...
     // $$$ this assumes page numbers are unique
     if (params.page != this._models.book.getPageNum(this.currentIndex())) {
       this.jumpToPage(params.page);
     }
   }
 
-  // process /search
-  // @deprecated for urlMode 'history'
-  // Continues to work for urlMode 'hash'
-  if (this.enableSearch && 'undefined' != typeof(params.search)) {
```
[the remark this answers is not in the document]
shaneriley: In my testing post-refactor, the search request kicked off twice and rendered two sets of results markers. What's the format of the hash-param URLs? I don't see any examples in the BookReader dev doc, and the test page uses a hash URL like https://deploy-preview-420--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=artofcaricaturin006061mbp#page/n22/mode/2up/search/bristol and works fine. In Petabox, the `?q` param is used instead and works as well.

src/js/plugins/plugin.search.js
```diff
@@ -31,6 +49,11 @@ BookReader.prototype.setup = (function (super_) {
     this.server = options.server;
     this.subPrefix = options.subPrefix;
     this.bookPath = options.bookPath;
+
+    this.searchView = new SearchView({
```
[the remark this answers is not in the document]
shaneriley: Yeah, (BookReader core)[https://github.com/internetarchive/bookreader/blob/master/src/js/BookReader.js#L201] uses class constructors like this to create the nav bar and toolbar.

src/js/plugins/plugin.search.js
shaneriley: Yeah, with a sourcemap and breakpoints, we really don't need inline debugging conditionals.

src/js/plugins/plugin.search.js
```diff
@@ -267,38 +211,16 @@ BookReader.prototype.search = function(term = '', options) {
...
-}
-
-/** @deprecated */
-BookReader.prototype.BRSearchCallbackErrorDesktop = function(results, options) {
-  const $el = $(this.popup);
-  this._BRSearchCallbackError(results, $el, true);
+  // BookReader's wrapper on jQuery's trigger method is broken. Props are not
+  // passed down, since the method only takes two arguments. The actual props
+  // passed is the BookReader instance.
+  $(document).trigger('BookReader:SearchCallback', { results, options, instance: this });
```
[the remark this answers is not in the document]
shaneriley: The comment above explains why I can't use BookReader's wrapper around `$.fn.trigger`. That's what I described to you in our call last week. The view listens to this event and gets the object as a parameter. The search results are the local variable `results`.

src/js/plugins/plugin.search.view.js
shaneriley: This is the search results container in the search panel.

src/js/plugins/plugin.search.view.js
shaneriley: Yeah, trying to use as little jQuery as possible in new BR code. Still had to use it for the event system, but we can set and trigger arbitrary events on the document without it. Refactoring that would require touching all of BookReader, though.

src/js/plugins/plugin.search.view.js
shaneriley: This was a comment from the existing search plugin code. I didn't quite understand it, but figured I could leave it until we come back and remove jQuery.

src/js/plugins/plugin.search.view.js
shaneriley: ✨ 🦄 🍰

tests/e2e/helpers/mobileSearch.js
shaneriley: They do, however the local `nav` is `br.nav.mobile`, so it's scoped to the mobile search query input.

iisa: ah ok, is that why the desktop search input didn't have the search value that was passed in on page load?

src/js/plugins/plugin.search.js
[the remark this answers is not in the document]
iisa: speaking in javascript, can we move the callback to a variable above so that it keeps scope?

It would be nice to extend `br.trigger` to accept an optional data object as well.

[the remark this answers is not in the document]
shaneriley: The second argument isn't a callback, it's an object passed through to the event listener. In the docs, it's the `[extraParameters]` parameter. https://api.jquery.com/trigger/ That's why I can't use the BookReader wrapper around `$.fn.trigger`. The wrapper calls `$(document).trigger('BookReader:' + name, this, props)`, so the `extraParameters` that the event listener gets is the BookReader instance. I want to pass in the matches, so I can't use the method wrapped around jQuery's method. Likewise, there'd be no way to extend BookReader's wrapper around it, because it will ultimately pass only the BookReader instance through.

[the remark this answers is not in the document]
iisa: perhaps we can make the fork inside of `br.trigger` - i am quite cautious of the 2 different ways we will be firing events. i really like the use of data emission here and that is something we should definitely bring into the fold.

@cdrini - interested in your input here.

src/js/plugins/plugin.search.view.js
iisa: pardon, i meant the full list declaration in `cacheDOMElements`

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,328 @@
...
+    // The element added to the mobile menu that is animated into view when
+    // the "search" nav item is clicked
+    this.dom.mobileSearch = this.buildMobileDrawer();
+    // Search input within the top toolbar. Will be removed once the mobile menu is replaced.
+    this.dom.toolbarSearch = this.buildToolbarSearch();
+  }
+
+  toggleSearchTray(bool) {
+    const state = typeof bool !== 'boolean' ? this.dom.searchTray.classList.contains('hidden') : bool;
+    this.dom.searchTray.classList.toggle('hidden', !state);
```
cdrini, line 44: This should work
```suggestion
  toggleSearchTray(bool=this.dom.searchTray.classList.contains('hidden')) {
    this.dom.searchTray.classList.toggle('hidden', !bool);
```

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,328 @@
...
+  }
+
+  toggleResultsCount(bool) {
+    this.dom.resultsCount.classList.toggle('visible', bool);
+  }
+
+  updateResultsCount(results) {
```
cdrini, line 51:
```suggestion
  /**
   * @param {number} results
   */
  updateResultsCount(results) {
```

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,328 @@
...
+  }
+
+  clearSearchFieldAndResults() {
+    this.toggleResultsCount(false);
+    this.removeResultPins();
+    this.emptyMatches();
+    this.setQuery('');
+  }
+
+  renderSearchTray(selector) {
```
cdrini, line 75: Doc string for what this selector is used for; it looks like it gets converted to an ID?

src/js/plugins/plugin.search.js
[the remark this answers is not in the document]
cdrini: Since we can assume no one uses `props` (since it does literally nothing :) ), maybe we can change it to this?

```js
/**
 * @param {EVENTS} name
 * @param {*} [props]
 */
BookReader.prototype.trigger = function(name, props = this) {
  $(document).trigger('BookReader:' + name, props);
};
```

src/js/plugins/plugin.search.js
```diff
@@ -152,8 +109,7 @@ BookReader.prototype.initToolbar = (function (super_) {
  * @param {string} term
  * @param {SearchOptions} options
```
cdrini, line 110:
```suggestion
 * @param {SearchOptions} overrides
```

src/js/plugins/plugin.search.js
```diff
@@ -307,20 +229,7 @@ BookReader.prototype.BRSearchCallbackErrorDesktop = function(results, options) {
  * @param {SearchInsideResults} results
```
cdrini, line 229: Use this type param in SearchView wherever results are used

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,328 @@
...
+    if (this.br.options.initialSearchTerm) {
+      this.setQuery(params.query);
+    }
+  }
+
+  cacheDOMElements() {
+    this.dom = {};
+
+    // The parent search tray in mobile menu
+    this.dom.searchTray: this.renderSearchTray(params.selector);
```
cdrini, line 26: Um, typo? :D 

```suggestion
    this.dom.searchTray = this.renderSearchTray(params.selector);
```

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,328 @@
...
+    const pinsVisibleState = bool ? 'visible' : 'hidden';
+    instance.refs.$BRfooter.find('.BRsearch').css({ visibility: pinsVisibleState });
+  }
+
+  buildMobileDrawer() {
+    const mobileSearch = document.createElement('li');
+    mobileSearch.innerHTML = `
+      <span>
+        <span class="DrawerIconWrapper">
+          <img class="DrawerIcon" src="${this.br.imagesBaseURL}icon_search_button.svg" alt="Icon of a magnifying glass" />
```
cdrini, line 122: > You should always have text labels next to icons. Assuming you do, the icon should not have an alt-text. Let me explain why!

https://axesslab.com/alt-texts/

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,328 @@
...
+    this.renderModalMessage('Sorry, there was an error with your search.<br />The text may still be processing.');
+    this.delayModalRemovalFor(4000);
+  }
+
+  renderBookNotIndexedModal() {
+    this.renderModalMessage(`
+      <p>
+         This book hasn't been indexed for searching yet.
+         We've just started indexing it, so search should be available soon.
+         Please try again later. Thanks!
```
cdrini, line 236:
```suggestion
         Please try again later. Thanks! 🙇 
```

(just kidding :P )

src/js/plugins/plugin.search.js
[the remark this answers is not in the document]
shaneriley: Fine with me. The inner workings of that method are not long for this world anyway if we're going to be removing jQuery.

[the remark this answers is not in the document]
shaneriley: Note that, if anyone using BR relies on that instance being passed through, those event listeners will now break.

src/css/_BRsearch.scss
```diff
@@ -199,6 +169,180 @@
...
     }
   }
+
+  #BRsearch_tray {
+    --menuSliderBg: #151515;
+    --primaryTextColor: #fff;
+    --activeButtonBg: #282828;
+    --searchResultText: #adaedc;
+    --searchResultBg: #272958;
+    --searchResultBorder: #fff;
```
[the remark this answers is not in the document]
shaneriley: Starting the switch to native CSS variables instead for most cases.

src/css/_BRsearch.scss
```diff
@@ -199,6 +169,180 @@
...
+      -webkit-appearance: none;
+      appearance: none;
+      -webkit-mask: url('data:image/svg+xml;base64,PHN2ZyB2aWV3Qm94PSIwIDAgNDAgNDAiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIgYXJpYS1sYWJlbGxlZGJ5PSJjbG9zZVRpdGxlSUQgY2xvc2VEZXNjSUQiPjxwYXRoIGQ9Ik0yOS4xOTIgMTAuODA4YTEuNSAxLjUgMCAwMTAgMi4xMkwyMi4xMjIgMjBsNy4wNyA3LjA3MmExLjUgMS41IDAgMDEtMi4xMiAyLjEyMWwtNy4wNzMtNy4wNy03LjA3IDcuMDdhMS41IDEuNSAwIDAxLTIuMTIxLTIuMTJsNy4wNy03LjA3My03LjA3LTcuMDdhMS41IDEuNSAwIDAxMi4xMi0yLjEyMUwyMCAxNy44NzhsNy4wNzItNy4wN2ExLjUgMS41IDAgMDEyLjEyMSAweiIgY2xhc3M9ImZpbGwtY29sb3IiIGZpbGwtcnVsZT0iZXZlbm9kZCIvPjwvc3ZnPg==') 0 0 no-repeat;
+      mask: url('data:image/svg+xml;base64,PHN2ZyB2aWV3Qm94PSIwIDAgNDAgNDAiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIgYXJpYS1sYWJlbGxlZGJ5PSJjbG9zZVRpdGxlSUQgY2xvc2VEZXNjSUQiPjxwYXRoIGQ9Ik0yOS4xOTIgMTAuODA4YTEuNSAxLjUgMCAwMTAgMi4xMkwyMi4xMjIgMjBsNy4wNyA3LjA3MmExLjUgMS41IDAgMDEtMi4xMiAyLjEyMWwtNy4wNzMtNy4wNy03LjA3IDcuMDdhMS41IDEuNSAwIDAxLTIuMTIxLTIuMTJsNy4wNy03LjA3My03LjA3LTcuMDdhMS41IDEuNSAwIDAxMi4xMi0yLjEyMUwyMCAxNy44NzhsNy4wNzItNy4wN2ExLjUgMS41IDAgMDEyLjEyMSAweiIgY2xhc3M9ImZpbGwtY29sb3IiIGZpbGwtcnVsZT0iZXZlbm9kZCIvPjwvc3ZnPg==') 0 0 no-repeat;
```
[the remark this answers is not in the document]
shaneriley: What do you mean, explain what a mask image is doing?

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,328 @@
+class SearchView {
```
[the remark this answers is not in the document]
shaneriley: TTS is also within the subdirectory. Should I move both?

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,328 @@
...
+    const items = matches.map((match) => `
+      <li data-page="${match.par[0].page}" data-page-index="${this.br.leafNumToIndex(match.par[0].page)}">
+        <h4>Page ${match.par[0].page}</h4>
+        <p>${match.text.replace(this.matcher, '<mark>$1</mark>')}</p>
+      </li>
+    `);
+    this.dom.results.innerHTML = items.join('');
+  }
+
+  togglePinsFor(instance, bool) {
```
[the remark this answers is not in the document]
shaneriley: It's no longer needed. At the time, I had the BR instance being passed in, and it happened enough that I just cached it on the view instance in the constructor instead.

src/js/plugins/plugin.search.js
shaneriley: That's going to translate even though there's no definition of `SearchInsideResults` in this file?

src/js/plugins/plugin.search.view.js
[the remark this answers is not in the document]
cdrini: I'm not sure I follow; I'm thinking the pattern should yield:

```
plugins/
   plugin.foo.js
   tts/
       plugin.tts.js
       AbstractTTSEngine.js
   search/
      plugin.search.js
      SearchView.js
```

They're like little subprojects. Also update webpack config to point to correct entry file for the plugin

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,328 @@
...
+    this.matcher = new RegExp('{{{(.+?)}}}', 'g');
+    this.cacheDOMElements();
+    this.bindEvents();
+
+    if (this.br.options.initialSearchTerm) {
+      this.setQuery(params.query);
+    }
+  }
+
+  cacheDOMElements() {
```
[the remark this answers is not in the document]
shaneriley: A reference to the DOM element is cached on the `dom` object, so you don't have to re-traverse the DOM to locate them.

src/js/plugins/plugin.search.view.js
```diff
@@ -0,0 +1,328 @@
...
+        <div class="loader tc mt20"></div>
+      </div>
+      <ul data-id="results"></ul>
+    `;
+    return searchTray;
+  }
+
+  renderMatches(matches) {
+    const items = matches.map((match) => `
+      <li data-page="${match.par[0].page}" data-page-index="${this.br.leafNumToIndex(match.par[0].page)}">
```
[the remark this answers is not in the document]
shaneriley: This isn't pointing to any CSS. Did you mark the wrong file?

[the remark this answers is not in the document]
cdrini: This is referenced in the css using `li` /etc; let me see if I can find where.

src/js/BookReader.js
```diff
@@ -496,10 +496,10 @@ BookReader.prototype.init = function() {
 
 /**
  * @param {EVENTS} name
- * @param {*} [props]
+ * @param {array | object} [props]
  */
-BookReader.prototype.trigger = function(name, props = undefined) {
-  $(document).trigger('BookReader:' + name, this, props);
+BookReader.prototype.trigger = function(name, props = {}) {
```
cdrini, line 501: This should default to `this`, otherwise we have to update a _bunch_ of code

```suggestion
BookReader.prototype.trigger = function(name, props = this) {
```

src/css/_BRsearch.scss
```diff
@@ -199,6 +169,181 @@
...
+    li h4 {
+      grid-column: 2 / 3;
+      padding: 0 0 20px 0;
+      margin: 0;
+      font-weight: normal;
+    }
+
+    li p {
+      grid-column: 2 / 3;
+      padding: 0 0 15px 0;
```
cdrini, line 321: Here; this code (+neighbouring) is hard to read because I don't know what `li` represents here

shaneriley: I don't understand how this is difficult. These styles are scoped to list items that exist within the parent selector block, which is `#BRsearch_tray`. That selector is nested within `#{$brScope}`. That means that, once compiled, the selector chain for styling search tray list items and their children use the selector chain `.BookReader #BRsearch_tray li, .BRmobileMenu #BRsearch_tray li, .BRfloat #BRsearch_tray li`

cdrini: It's not a blocker; having something like `li.search-result p` would make the CSS self-documenting and easier to read. As is, it's unclear and requires the reader to jump into the HTML code to try decipher where this `li` lives. It also makes the css harder to find from the js/html because there's nothing I can `grep` for to find this css; I'd have to _know_ beforehand that this CSS exists. Or I'd have to try to read all the css in this file to see if any applies to it.

cdrini: It's kind of the equivalent of naming a variable `num`. You know the type, but you don't know what it actually is, and you have to spend time trying to figure out what it is by reading a lot of neighbouring code.

shaneriley: This generates more HTML and CSS for the UA to download, when a selector with that level of specificity is not needed. Additionally, this is written to style ANY list items (and certain child elements) with these properties, not just ones that you remember to add a class to. If we're going to apply styles to everything using class names only, we need to first rewrite the HTML and SCSS throughout the project (there are many element selectors throughout the SCSS) and then specify in the contributing guide that CSS is to be applied via class selectors.

On top of all of that, I will reiterate that this is basically throwaway code. As soon as the search results WebComponent is released, it's taking the place of this view constructor and CSS. At that time, nearly all of the CSS selectors within the view will be element-level.

shaneriley:
e2e tests are passing locally, and had been passing throughout development of this feature. Merging despite the Travis failure. If it continues to be an issue, we will address it separately.

PR #421, opened 2020-08-18
E2E test: updates to navigation model & base tests

iisa:
Added new helpers to navigation base test that gets the "next" and "previous" button given a book's page progression.

Approach:
There is a clear scope delineation between `ClientFunction` and local test scope.  `ClientFunction` allows one to reach into the light DOM while local test scope only can only reference TestCafe selectors.

I was trying to consolidate the `window.br.pageProgression` check in the button helpers, but you can't mix the two and you will lose the scope of the other.  So, we fetch for the book's page progression first and then calculate which button to serve.

Acceptance: e2e tests run

Follow up work from #419


tests/e2e/helpers/base.js
```diff
@@ -94,8 +118,8 @@ export function runBaseTests (br) {
     const initialUrl = await getUrl();
 
     // Set Cookie by page navigation, wait for cookie
-    const goForward = await isRightToLeft() ? nav.desktop.goPrevious : nav.desktop.goNext;
-    await t.click(goForward)
+    const goNext = getNext(nav, await bookPageProgression());
```
shaneriley, line 121: For consistency, you could instantiate a `pageProgression` variable like you do in line 143 and pass it to `getNext`

iisa: done

PR #433, opened 2020-09-02
Fix wrong openlibrary URL

cdrini:
This was causing a redirect to trigger


cdrini:
Automerging since it's a 1-liner with very few consequences.

PR #434, opened 2020-09-07
webdev-3628 HTML escaping to prevent XSS

nsharma123:
when book item has inline JS, resulting some popups and alerts execution on bookreader thearter. we want to prevent those alerts by `escapeHTML()` function while render the title in HTML.
https://webarchive.jira.com/browse/WEBDEV-3628

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@internetarchive/bookreader",
-  "version": "4.15.0",
+  "version": "4.15.1-alpha.2b5e76c",
   "description": "The Internet Archive BookReader.",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@internetarchive/bookreader",
-  "version": "4.15.1-alpha.2b5e76c",
+  "version": "4.15.0",
   "description": "The Internet Archive BookReader.",
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


src/js/BookReader/Toolbar/Toolbar.js
```diff
@@ -285,10 +285,11 @@ export class Toolbar {
 
     // A loop to build fields
     for (const {extraValueClass = '', label, value} of br.metadata) {
+      const newValue = label === 'Title' ? escapeHTML(value) : value;
```
shaneriley, line 288: **Suggestion**: change the variable name to `escapedValue`

package.json
```diff
@@ -46,7 +46,7 @@
     "sinon": "9.0.3",
     "testcafe": "1.8.6",
     "webpack": "4.44.1",
-    "webpack-cli": "3.3.9"
+    "webpack-cli": "^3.3.9"
```
shaneriley, line 49: The dependencies should be locked to a specified version, so this caret should be removed.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@internetarchive/bookreader",
-  "version": "4.15.0",
+  "version": "4.15.1-alpha.2b5e76c",
```
shaneriley, line 3: This needs to be reverted. Canary versions can be published for testing, but should not be part of the commit history. The version will be updated once a new release is published.

shaneriley (changes requested):

PR #437, opened 2020-09-09
Run travis on master branch pushes!

cdrini:
Looks like I forgot master is also a branch :D 

I enabled "build on pushed branches" in https://travis-ci.com/github/internetarchive/bookreader/settings, and set it to only build master via .travis.yml


cdrini:
Auto-merging since this is tiny and highly unlikely to be controversial

PR #443, opened 2020-09-11
Search plugin fixes

shaneriley:
This fixes the issues addressed in https://webarchive.jira.com/browse/WEBDEV-3688. Review app is at https://www-shaner.archive.org/details/goody and test steps are noted in the ticket comments.

### DIFF
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -2681,6 +2681,16 @@ BookReader.prototype.updateFromParams = function(params) {
     }
   }
 
+
+  // process /search
+  // @deprecated for urlMode 'history'
+  // Continues to work for urlMode 'hash'
+  if (this.enableSearch && 'undefined' != typeof(params.search)) {
+    if (this.searchTerm != params.search) {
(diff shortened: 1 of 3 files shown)


src/js/BookReader.js
```diff
@@ -2681,6 +2681,16 @@ BookReader.prototype.updateFromParams = function(params) {
     }
   }
 
+
+  // process /search
+  // @deprecated for urlMode 'history'
+  // Continues to work for urlMode 'hash'
+  if (this.enableSearch && 'undefined' != typeof(params.search)) {
+    if (this.searchTerm != params.search) {
```
bfalling, line 2689: SUGGESTION: Prefer `!==`?

bfalling (approved):
One suggestion. Otherwise, LGTM.

src/js/BookReader.js
shaneriley: Yeah, I resurrected this code from before my refactor and didn't verify that it wasn't type coercing. I'm guessing not, since this should always be treated as a string, even if it is only comprised of numbers.

PR #444, opened 2020-09-14
Fix: Polyfill CustomEvent constructor

cdrini:
This was causing IE11 to brick. Apparently not supported by core-js :(
https://github.com/zloirock/core-js/issues/354


shaneriley (approved):

cdrini:
Tested on IE11 ✅

PR #447, opened 2020-09-15
Fix text selection on Safari

cdrini:
Where to begin. So the way this works, is the svg text layer has `pointer-events: none` (so that you can right click on non-text and access image options). The word elements (usually `<tspan>`) override this with `pointer-events: all`, so that you can have the cursor hover effect, and so the event listeners work. The problem is that for some reason, on Safari, the `pointer-events: all` override wasn't working on `<tspan>` elements. So the solution I ended up with was on Safari, apply that override on the _paragraph_ element. This has some problems, in that it makes it a little more difficult to pan when in 2up zoomed-in, but should otherwise be fine.

Other things I tried:
- Using `<text>` for the word instead of `<tspan>` (as we do on Firefox). We can't do this on Chrome/Safari, because those browsers add a newline character after `<text>` elements when copying the text, so you get one word per line. I tried modifying the `oncopy` result to make it clean up the copied text, but I found no clean/reliable way to differentiate the word newlines from the paragraph newlines.
- Having HTML embedded in the SVG (via `<foreignObject>`). This would likely work, but was a huge change a would require a lot of testing. Something we might want to explore though at some point, because it did result in smoother selecting.
- Having a `<text>` for each line (currently we have one per paragraph). This would likely work, but would make the copy UX less nice since you'd have newlines instead of paragraphs. Also would require a ton of code changes, which would require a lot of testing.

### DIFF
--- a/src/js/plugins/plugin.text_selection.js
+++ b/src/js/plugins/plugin.text_selection.js
@@ -1,5 +1,5 @@
 //@ts-check
-import { isFirefox, isSafari } from './tts/utils';
+import { isFirefox, isSafari } from '../../util/browserSniffing.js';
 
 const BookReader = /** @type {typeof import('../BookReader').default} */(window.BookReader);
 
--- a/src/js/plugins/plugin.text_selection.js
+++ b/src/js/plugins/plugin.text_selection.js
@@ -1,11 +1,11 @@
 //@ts-check
-import { isFirefox } from './tts/utils';
+import { isFirefox, isSafari } from './tts/utils';
 
 const BookReader = /** @type {typeof import('../BookReader').default} */(window.BookReader);
 
 export class TextSelectionPlugin {
 
-  constructor(avoidTspans = isFirefox()) {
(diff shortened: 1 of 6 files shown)


cdrini:
Tested: Safari 13.1 on Catalina, Chrome 85 on Win 10, FF 80 on Win 10

jbuckner (approved):
Code looks good and tested in Safari 14, Catalina. I'm not totally sure what the previous bug was (I think it was always working for me in Safari), but looks good 👍

cdrini:
Awesome, thanks for the review! One of the pieces of feedback we got while testing was that you couldn't right click to save an image any more (because the SVG layer was sitting over top the image). Giacomo fixed this by make the svg `pointer-event:none` and the text `pointer-events:all` (so you right clicking a "blank" part of the svg would send the click event down to the lower layer). Unfortunately it wasn't working in Safari, so everything was `pointer-events:none` there!

P.S. If you ever want to do a comparison test, remove the `deploy` prefix in the netlify url to see how master behaves; e.g. https://lucid-poitras-9a1249.netlify.app/

Issue #446, opened 2020-09-14
Add "semi" always eslint + fix violators


cdrini:
Eslint docs: https://eslint.org/docs/rules/semi

### Summary of Requirements
- [ ] New rule added to eslint rc
- [ ] `npm run lint:fix` used to fix any violators

### Stakeholders

nikhilgupta2001:
i would like to work on this issue

iisa:
ok sounds good @nikhilgupta2001

nikhilgupta2001:
mam do you have a slack channel or other platform for connecting with community if yes please provide link i would like to be in touch

Yashs911:
Hey, @nikhilgupta2001 are you working on this issue. If not then can I create a PR?

nikhilgupta2001:
Yah i am working on it

Yashs911:
Hey, @cdrini I would like to work on this issue.

cdrini:
If the PR stagnates for more than a week, I'll close it so that others can give it a try :) Would you be interested in tackling #470 , perhaps?

ArunTeltia:
@cdrini please have a look the assignee was not doing pr, so i thought i can try this issue

Yashs911:
@cdrini can I open a PR for this issue as its already 8 days and there is no activity. and do I need to exclude BookReader/ and plugins/ both or only BookReader/ while creating a PR?

nikhilgupta2001:
I have updated my repo to latest and made the required changes in eslintrc.js for linting . now should I directly add and commit please guide me @cdrini @iisa

hkumar0132:
I would like to work on this issue if it's still open @cdrini

SwapnilSoni1999:
Is this issue still open?

duggiefresh:
I think this issue can be closed by https://github.com/internetarchive/bookreader/pull/789. Thanks!

PR #448, opened 2020-09-17
#446 

nikhilgupta2001:
i have changed the setting to always require semicolon if there is any problem in this let me know.


cdrini (changes requested):
Awesome! This is looking good :) A few notes:

It looks like this is your first PR! I'd recommend reading https://github.com/internetarchive/openlibrary/wiki/Git-Cheat-Sheet if you happen to be new to git.

A few notes:
1) Don't worry about codecov; it's just confused :P
2) It looks like your fork has outdated code; rebase it and try running your commands again (see the Git Cheat Sheet)
3) Don't commit the `package.json` or the `package-lock.json`; there shouldn't be any changes there
4) Don't commit any of the `BookReader/` files; those are auto-generated (See https://github.com/internetarchive/bookreader#development ).

Let us know if you have any questions!

PR #449, opened 2020-09-18
Feature/text sel options

cdrini:
Add new options to change text selection url used to fetch djvu xml.

Testing can happen on https://www-drini.archive.org/details/goodytwoshoes00newyiala , which has:

```js
{
  vars: { bookId: '...', server: '...', bookPath: '...' },
  plugins: {
    textSelection: { enabled: true, singlePageDjvuXml: '...blah...' }
  }
}
```

The the `vars` and mustache-style templates for the options are a little awkward, but here's what I considered:

- Tried having explicit objects (like `{ url: '', parameters: { path: '' } }`, etc). This resulted in too much code on the BR side to handle it. And it became brittle, and very verbose.
- No vars, and have petabox construct the url itself. This works, but is a little repetitive. With `vars`, they can be set once, and use in any settings in a semi-logical way. They might also be useful in the future if a var changes mid-way. If it's in petabox, you'd have to do a hard reload.

Open questions/concerns:
- I don't like how `pageIndex` looks exactly like the other vars, when it's in fact different.
- Note this will use the old, big djvu on the netlify app. Coordinating these was a little awkward, so this will have to get released, then the petabox side will need to get released (with the new version), then I can open a new PR to make this use the single-page option.

### DIFF
--- a/src/js/plugins/plugin.text_selection.js
+++ b/src/js/plugins/plugin.text_selection.js
@@ -1,11 +1,24 @@
 //@ts-check
 import { isFirefox, isSafari } from '../../util/browserSniffing.js';
+import { applyVariables } from '../../util/strings.js';
+/** @typedef {import('../../util/strings.js').StringWithVars} StringWithVars */
 
 const BookReader = /** @type {typeof import('../BookReader').default} */(window.BookReader);
 
+export const DEFAULT_OPTIONS = {
+  enabled: true,
+  /** @type {StringWithVars} The URL to fetch the entire DJVU xml. Supports options.vars */
+  fullDjvuXmlUrl: null,
+  /** @type {StringWithVars} The URL to fetch a single page of the DJVU xml. Supports options.vars. Also has {{pageIndex}} */
+  singlePageDjvuXmlUrl: null,
+};
+/** @typedef {typeof DEFAULT_OPTIONS} TextSelectionPluginOptions */
+
 export class TextSelectionPlugin {
 
-  constructor(avoidTspans = isFirefox(), pointerEventsOnParagraph = isSafari()) {
+  constructor(options = DEFAULT_OPTIONS, optionVariables, avoidTspans = isFirefox(), pointerEventsOnParagraph = isSafari()) {
+    this.options = options;
+    this.optionVariables = optionVariables;
     /**@type {PromiseLike<JQuery<HTMLElement>|undefined>} */
     this.djvuPagesPromise = null;
     // Using text elements insted of tspans for words because Firefox does not allow svg tspan strech.
@@ -23,13 +36,12 @@ export class TextSelectionPlugin {
     }
   }
 
-  /**
-   * @param {string} ocaid
-   */
-  init(ocaid) {
+  init() {
+    // Only fetch the full djvu xml if the single page url isn't there
+    if (this.options.singlePageDjvuXmlUrl) return;
     this.djvuPagesPromise = $.ajax({
       type: "GET",
-      url: `https://cors.archive.org/cors/${ocaid}/${ocaid}_djvu.xml`,
+      url: applyVariables(this.options.fullDjvuXmlUrl, this.optionVariables),
       dataType: "xml",
 
       error: function (e) {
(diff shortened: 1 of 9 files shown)


src/js/plugins/plugin.text_selection.js
```diff
@@ -212,7 +237,7 @@ export class BookreaderWithTextSelection extends BookReader {
   _createPageContainer(index, styles = {}) {
     const $container = super._createPageContainer(index, styles);
     // Disable if thumb mode; it's too janky
-    if(this.enableTextSelection && this.mode != this.constModeThumb){
+    if(this.textSelectionPlugin && this.mode != this.constModeThumb){
       this.textSelectionPlugin.createTextLayer(index, $container);
```
jbuckner, line 241: Can you use an optional here too, ie:
```
if(this.mode != this.constModeThumb){
  this.textSelectionPlugin?.createTextLayer(index, $container); 
}
```

jbuckner (commented):
Question/comment about the url generation with the `applyVariables()` method... it feels like there's a shared responsibility of the url for the djvu endpoint. ie the consumer passes in the base url template with the mustache-style variables and inside book reader, those get merged and added to with the `pageIndex`.

What should be the authoritative source be for that URL?

If the consumer (ie petabox) should be the authoritative source, one idea that came to mind would be using a "url generation delegate" that is responsible for generating the URLs. For instance, you could do something like this:

Consumer:
```
const br = new BookReader();
br.urlGenerationDelegate = this;  // make the consumer the delegate (or another object if it makes sense)

// I'm assuming that `abc123` is known to the consumer
generateFullDjvuUrl() {
  return `https://cors.archive.org/cors/abc123/abc123_djvu.xml`
}

// BR calls this method and passes up the pageIndex so the consumer can generate a URL for it
generateSinglePageUrlFor(pageIndex) {
  return `https://cors.archive.org/cors/abc123/abc123_djvu.xml?pageIndex=${pageIndex}`
}
```

Then inside BR when you want need to make the XHR calls:
```
const url = this.urlGenerationDelegate.generateSinglePageUrlFor(pageIndex);
return $.ajax({
  type: "GET",
  url: url,
  ...
})
```

You can also expand the `urlGenerationDelegate` if additional URLs need to be generated for different purposes.

jbuckner:
Thinking some more about the delegate idea, you could actually extend the delegate's responsibilities to fetch and return the djvu instead of just generating a URL. This would let you easily mock out the djvu by passing in a mock delegate and also allow for alternate backends instead of just XHR requests. 

eg, your consumer could look like this:

```
// consumer
const br = new BookReader()
br.djvuFetcherDelegate = this;

async fetchFullDjvu() {
  return fetch('https://url.to.djvu')
}

async fetchSinglePageDjvu(pageIndex) {
  return fetch('https://url.to.djvu?pageIndex=${pageIndex})
}

// inside BR:
const djvu = await this.djvuFetcherDelegate.fetchSinglePageDjvu(3);
```

src/js/plugins/plugin.text_selection.js
cdrini: Ah, good idea! I thought this one stood out :D

cdrini:
I think the responsibility should lie on the client; in an ideal world, BookReader should have no internal knowledge about what IA is doing. The `vars` thing is kind of nice because it basically let's petabox extend BR's data store (currently it does this by adding any/all parameters to br.options :/ ).

I was thinking of using a function! Long term that would be a great way to do it (it would also be more flexible, because the user could e.g. do OCR on the fly if they really wanted to or something :P ). The reason I shifted away from it is because a lot of this config is in petabox's PHP code 😬 And also a lot of the BR-related JS code doesn't pass through any compilation, so it's stuck in ES5 :/ Also, I figured we could very easily extend this in the future to optionally take a function.

src/js/plugins/plugin.text_selection.js
cdrini: Fixed

jbuckner:
When you say "responsibility should lie on the client", do you mean the consumer of book reader (ie petabox in our case)?

In the petabox case, when BR is instantiated, it has to pass that configuration from the PHP to the javascript somehow, right? Couldn't you instantiate that fetcher function with the same configuration and pass that in? Then BR could call the function with the new page index and it'd have it return the new url or fetch the djvu.

I might be misunderstanding the full scope of the issue, but I'm a bit hesitant about the mustache-style solution. For me, this feels like a generalized solution for a targeted problem. 

It puts a lot of responsibility on the consumer to know that, when they pass in `fullDjvuXmlUrl` and `optionsVariables` that it's going to get merged together to form a URL in the end and then an XHR request will be made with it. Internally, you also have to worry about the url format passed in by the consumer since your method is taking these values and merging them together (and adding to them).

By making the consumer the authoritative source and passing in a function/delegate that returns the URL or full djvu file, there's no ambiguity of what's going on.

jbuckner (approved):

PR #452, opened 2020-09-23
Refactor BookReaderSimpleJS files

jimchamp:
This is far from complete.  I've left links to the old demos and included the refactored demos at the bottom of index.html so that it's easier to compare them.

Extra options are successfully being passed from the query string and used to instantiate the book reader.  If you look at the network logs in your devtools, you'll see that the files are being loaded in a different order.

Some additional notes about the BookReaderSimpleJS demo files can be found [here](https://docs.google.com/document/d/1VoCE505Gljz2A27W_AGV6hjob1qOC-lG6z6a5rBKUyM/edit?skip_itp2_check=true).


jamesachamp:
This update should complete the refactor of the demos that do not require plugins and unique styling.

Remaining demos that need refactored are as follows:

- demo-fullscreen.html
- demo-fullscreen-mobile.html
- demo-vendor-fullscreen.html
- immersion-mode.html

BookReaderDemo/simple-bookreader-demo.html
```diff
@@ -0,0 +1,129 @@
...
+            );
+        }
+    </script>
+
+    <!-- BookReader and plugins -->
+    <link id="bookreader-css" rel="stylesheet" href="../BookReader/BookReader.css"/>
+    <script src="../BookReader/BookReader.js"></script>
+
+    <!-- Link to plugins that were passed in the query string -->
+    <script id="import-plugins">
```
jamesachamp, line 47: The suggested `document.write()` approach wrote directly to the body, so I went with this approach.

PR #454, opened 2020-09-25
Search results navigation

shaneriley:
This creates a new search results navigation view conditionally rendered when search results have been returned. Buttons to navigate to next and previous results surround text that shows the current result and total number of results. A button on the right side of the view clears the search results when clicked.

Completes [WEBDEV-3802](https://webarchive.jira.com/browse/WEBDEV-3802)


src/css/_BRsearch.scss
```diff
@@ -345,6 +355,58 @@
...
+      font-weight: normal;
+      span {
+        vertical-align: middle;
+      }
+    }
+    .pagination {
+      position: absolute;
+      left: 50%;
+      top: 50%;
+      transform: translate(-50%, -50%);
```
bfalling, line 381: QUESTION: Teach me: Why use these positioners to do the centering vs. flex box?

src/js/plugins/search/view.js
```diff
@@ -122,6 +126,102 @@ class SearchView {
...
+    this.currentMatchIndex = this.matches.indexOf(matchingSearchResult);
+  }
+
+  find1upMatchingSearchResult() {
+    return this.matches.find((m) => this.br.currentIndex() === m.par[0].page - 1);
+  }
+
+  find2upMatchingSearchResult() {
+    return this.matches.find((m) => this.br._isIndexDisplayed(m.par[0].page - 1));
+  }
```
bfalling, line 216: QUESTION: For either of these functions, will `m.par[0].page` ever be 0? Is that okay given we're subtracting 1 from the value?

bfalling (approved):
A few questions. Otherwise, LGTM.

src/css/_BRsearch.scss
shaneriley: The element won't center if you simply rely on flexbox. There are three elements in the parent. The first is wider than the last, and the space between them and the middle one will be equal. As a result, the middle element is not centered. Attached are two screenshots illustrating first the flexbox space-between centering and the positioning present in this PR.

Screen Shot 2020-09-28 at 6 05 36 AM
Screen Shot 2020-09-28 at 6 04 33 AM

src/js/plugins/search/view.js
shaneriley: The page number is the human-readable number, not the index. Since internally BookReader is only concerned with the index of the page and its data, I need to subtract one from the value.

PR #469, opened 2020-10-01
interstitial popup: clean up css so that it flows with parent container

iisa:
This PR cleans up css and hard positioning on progress modals.
Now, it flows with the parent container.

To test:
- pull down branch, run `npm install && npm run serve-dev` to run demo server & compile latest code.
- go to Internet Archive demo -> search 'virtue'
- watch progress bar come into view
- shrink & grow browser window and confirm that progress popup does not leave the center area of the bookreader

Or go to www-isa.archive.org/details/goody?q=catches
- when running search on page load, confirm that the progress popup stays inside of the bookreader container & does not bleed out of screen

ezgif com-gif-maker

### DIFF
--- a/src/js/plugins/search/view.js
+++ b/src/js/plugins/search/view.js
@@ -265,7 +265,7 @@ class SearchView {
     toolbarSearch.classList.add('BRtoolbarSection', 'BRtoolbarSectionSearch');
     toolbarSearch.innerHTML = `
       <form class="BRbooksearch desktop">
-        <input type="search" name="query" class="BRsearchInput" val="" placeholder="Search inside"></input>
+        <input type="search" name="query" class="BRsearchInput" val="" placeholder="Search inside"/>
         <button type="submit" class="BRsearchSubmit">
           <img src="${this.br.imagesBaseURL}icon_search_button.svg" />
         </button>
--- a/src/js/plugins/search/view.js
+++ b/src/js/plugins/search/view.js
@@ -265,7 +265,7 @@ class SearchView {
     toolbarSearch.classList.add('BRtoolbarSection', 'BRtoolbarSectionSearch');
     toolbarSearch.innerHTML = `
       <form class="BRbooksearch desktop">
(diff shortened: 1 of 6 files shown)


src/js/plugins/search/view.js
```diff
@@ -265,7 +265,7 @@ class SearchView {
     toolbarSearch.classList.add('BRtoolbarSection', 'BRtoolbarSectionSearch');
     toolbarSearch.innerHTML = `
       <form class="BRbooksearch desktop">
-        <input type="search" name="query" class="BRsearchInput" val="" placeholder="Search inside"/>
+        <input type="search" name="query" class="BRsearchInput" val="" placeholder="Search inside"></input>
```
bfalling, line 268: QUESTION: Why the closing `</input>` tag? I believe the HTML5 spec says it shouldn't be there.

bfalling (approved):
One question. Otherwise, LGTM.

bfalling:
QA Results (Mac Safari):
1) When I resized the window in the midst of the search progress, sometimes the progress modal floated vertically above the book. It stayed horizontally centered, which is good. And being vertically above the book didn't look bad, but it definitely wasn't vertically centered in the viewport. Is this by design?
2) From within search results, if I used the Next Page button (not Next Search Result), then I was correctly taken to the next page in the book, and the search result label correctly changed to show how many search results, instead of e.g. "2 / 12". However, if I then hit the Next Search Result button, I was expecting to see the next search result page _after_ the page I was on. Instead, it jumped backwards to the very first search result.
3) Using the "works of Plato" IA demo, if I search for the word "influenced", the search result on page 3 incorrectly selects all of the words on the page starting with "influenced" and continuing until the end of the page. Perhaps this is a full-text search issue.
4) When on the first search result, the Previous Search Result arrow should probably be grayed out. Similarly, when on the last search result, the Next Search Result arrow should be grayed out.

src/js/plugins/search/view.js
iisa: that's what I said to myself 😅 
the chrome dev console was barking about the need for a closing tag at this line, removed.

PR #472, opened 2020-10-02
Added comma dangle rule in src/js

### LINKED ISSUE #154: Add eslint rule comma-dangle: "only-multiline"
See https://eslint.org/docs/rules/comma-dangle . Should _only_ apply to the JS in `src/`.

Yashs911:
Fixes #154

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,6 +22,7 @@ module.exports = {
         "sourceType": "module"
     },
     "rules": {
+        "comma-dangle": ["error", "only-multiline"],
         "space-before-blocks": ["error"],
         "keyword-spacing": [2, {"before": true, "after": true}],
         "indent": ["error", 2],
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -46,5 +46,13 @@ module.exports = {
                 "no-var": ["off"],
             }
         },
+        {
+            "files": [
+                "src/js/**/*.js"
+            ],
+            "rules": {
+                "comma-dangle": ["error", "only-multiline"]
```
cdrini, line 54: Hmmm, could we add this with the other rules, above? Or is there a reason you created it as a separate section?

Yashs911: @cdrini in the PR it was mentioned only to JS in src/
If you want I can add this rule with other rules

cdrini: Let's try adding it to everything :+1:

cdrini (approved):
Lgtm!

PR #474, opened 2020-10-02
Remove old bt CSS

cdrini:
No refs to these classes in src


cdrini:
Auto-merging; tiny change

PR #475, opened 2020-10-02
Make Travis runs for any master branch

cdrini:


cdrini:
Auto-merging; small dev-only change

PR #476, opened 2020-10-02
Remove no longer used jquery.bt.min.js

### LINKED ISSUE #385: Delete BookReader/jquery.bt.min.js
No longer used as a result of @shaneriley 's changes 🎉

cdrini:
Closes #385


iisa (approved):
ok!

PR #484, opened 2020-10-03
add semi rule in eslintrc

ArunTeltia:
tried solving #446 issue with all test case passed


Yashs911 (commented):
Hi, @ArunTeltia I guess the assignee's PR was not closed yet, and I was the next person to claim it. But as you have already opened a new PR use "semi": ["error", "always", { "omitLastInOneLineBlock": true}] and try if it passes all tests.

ArunTeltia:
okay thank you for telling me that I will try this ,sorry for the trouble @Yashs911

PR #489, opened 2020-10-05
Skips subsequent creation of search views when setup called 1+n times

shaneriley:
The search plugin's `setup` method is being called twice, causing two instances of the search views to be created. This checks for the existence of a search view and skips creating a new instance, while also adding a call to teardown the search navigation when search results are returned to ensure one search navigation is being rendered for search results.

Testing steps and review app can be found in [WEBDEV-3832](https://webarchive.jira.com/browse/WEBDEV-3832)


jbuckner (approved):
Looks good :shipit:

PR #494, opened 2020-10-09
Adds data attribute to 2up page containers for side designation

shaneriley:
Since 2up mode page containers are in reverse DOM order when flipping pages forward, there is no accurate method of determining which container is the left positioned container and which is right. This adds a data attribute based on the leaf data to determine which side the container is positioned on.


iisa (approved):
glad the pageSide metadata helped! LGTM 🍰

bfalling (approved):
LGTM

PR #498, opened 2020-10-15
Waits to resize until first result flipped to

shaneriley:
Fixes [WEBDEV-3879](https://webarchive.jira.com/browse/WEBDEV-3879).

### DIFF
--- a/src/js/plugins/search/view.js
+++ b/src/js/plugins/search/view.js
@@ -141,7 +141,6 @@ class SearchView {
       </div>
     `);
     this.dom.searchNavigation = $(`.${selector}`);
-    this.br.resize();
   }
 
   resultsPosition() {
@@ -459,8 +458,9 @@ class SearchView {
    * @param {Event} e
    * @param {object} properties
    *   @param {object} properties.results
+   *   @param {object} properties.options
    */
-  handleSearchCallback(e, { results }) {
+  handleSearchCallback(e, { results, options }) {
     this.matches = results.matches;
     this.setCurrentMatchIndex();
     this.teardownSearchNavigation();
@@ -470,6 +470,13 @@ class SearchView {
     this.renderPins(results.matches);
     this.updateResultsCount(results.matches.length);
     this.toggleSearchPending(false);
+    if (options.goToFirstResult) {
+      $(document).one('BookReader:pageChanged', () => {
+        this.br.resize();
+      });
+    } else {
+      this.br.resize();
+    }
   }
 
   /**


jbuckner (approved):
👍

iisa:
hmmm, options.goToFirstResult - i thought this be a default behavior? if so, we shouldn’t need to toggle this
Q: which of these functions change the page?
handleSearchCallback(e, { results, options }) {
     this.matches = results.matches;
     this.setCurrentMatchIndex();
     this.teardownSearchNavigation();
    this.renderSearchNavigation();
    this.bindSearchNavigationEvents();
    this.renderMatches(results.matches);
     this.renderPins(results.matches);
     this.updateResultsCount(results.matches.length);
     this.toggleSearchPending(false);
     if (options.goToFirstResult) {
       $(document).one('BookReader:pageChanged', () => {
         this.br.resize();
       });
     } else {
       this.br.resize();
     }
I think we can just gracefully resize here, is there a set timeout we can do to remove this condition?

iisa (changes requested):
want to talk through some clarifiers before we merge in..

shaneriley:
`goToFirstResult` is default behavior. In fact, there's no way to override the behavior. I don't know why it's on an `options` object, but I didn't want to get into a refactor to fix a critical bug.

Regarding using a `setTimeout` instead of waiting for the page transition to be over, that's going to have the potential to continue to exhibit the same bug. If we have a setTimeout of 80ms because that's plenty of time in our testing, but network conditions in production may cause the image to take longer to load than the timeout, resulting in the same issue. This method is a surefire way of knowing that it's safe to resize the book container.

iisa:
Q: which of those functions in the callback changes the page to the book?

This current cascade was a direct result of adding the page direction (L or R) onto the DOM - https://github.com/internetarchive/bookreader/commit/f55804b22ed5510a67a4b3cc1bbf58b390a12e58

I'd hate for us to just tackle the symptom and not look for a stable way to add proper attributes to the container, or something else that is more passive than propagating an options flag (that is actually a default).

shaneriley:
The issue had nothing to do with adding the data attribute indicating page side. Commit 45d3e35 is where the issue arose. The resize call was added at the end of `renderSearchNavigation`, and in the event there's no existing page to land on when BookReader initializes, when results are returned the reader flips to the page for the first result. When this series of events occurs, the resize event is called at the same time the jump to index is called by the search controller, which kicks off a load of the new pages to be displayed. Since they're not loaded yet, when the resize method processes, it resizes the pages to be 0 height because there's no image contents to give them dimensions. The only way to know that the jump to the first search result will occur is to check for the property `goToFirstResult`, and if it's true, we need to wait until the page loading and flipping has occurred before we resize.

iisa:
ah ok! thanks for the clarifier.

what can we do to bring this block closer together, so we don't need a conditional around a bad option flag?
does this only happen in 2up mode? if so, can we isolate against that?
```
if (options.goToFirstResult) {
       $(document).one('BookReader:pageChanged', () => {
         this.br.resize();
       });
     } else {
       this.br.resize();
     }
```

shaneriley:
We've only seen it happen in 2up mode, but I don't know if it would happen in the other modes if they were made the starting mode. I would imagine it would definitely manifest in thumbnail mode, since that fetches far more images than 2up mode when it changes index. 1up is still a possibility based on network conditions on both ends. If we replaced the check for the `goToFirstResult` flag with a check against a view mode, then we won't resize in most cases (e.g.: starting a search after book is rendered).

iisa (approved):
ok, let's do it and keep you movin ⚡

PR #500, opened 2020-10-16
#446 I have made the required changes please have a look @iisa @cdrini 

nikhilgupta2001:
Let me know if there is something wrong in PR


ArunTeltia:
I think @nikhilgupta2001 you have deleted the BookReader folder and that is an essential part of code(As it is stated in the readme that we don't have to change anything in the BookReader folder)
It would be best if you, first of all, make your forked repo to be even with the InternetArchive/Bookreader 
Try writing this 
```
git remote add upstream https://github.com/internetarchive/bookreader.git
git fetch --all --prune
git checkout master 
git reset --hard upstream/master 
git push origin master --force 
```
I assume that will make your repo even
and then try changing the code in .eslintrc

nikhilgupta2001:
@ArunTeltia then how should I stop commiting Bookreader. because they told not to update gitignore .and which bookreader they dont want to be not commited the one inside src or Bookreader folder inside the main directory.
thanks):

ArunTeltia:
actually what its written that you dont have to change a thing in BookReader folder because It is a production thing, You have got it wrong I guess ,We are not suppose to touch that code that's what it means to not commit changes in BookReader folder 
Hope this helps :D

nikhilgupta2001:
so after changing in eslintrc.js i should directly add and commit ???

ArunTeltia:
Yes I presume that will do the work

PR #501, opened 2020-10-19
Remove exit animation TTS toolbar

shaneriley:
When focus is given to the search input in the top right of BookReader and tab is pressed twice, the contents of the book container are shifted upward to show the TTS toolbar offscreen. This hides the TTS toolbar when not active, preventing the controls from being part of the tab index. In order to both toggle display and animate position in and out, timeouts are employed to toggle the animation shortly after visibility is given and visibility is removed immediately after the animation out is completed.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>BookReader Vendor Native Fullscreen Demo</title>
+    <title>BookReader Demos</title>
 
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
     <meta name="apple-mobile-web-app-capable" content="yes">
--- a/src/css/_controls.scss
+++ b/src/css/_controls.scss
@@ -91,10 +91,21 @@
     height: 30px;
   }
 
+  @keyframes slideUp {
+    from {
+      opacity: 0;
+      transform: translateY(10px);
+    }
+    to {
+      opacity: 1;
+      transform: translateY(0);
+    }
+  }
+
   .read-aloud {
     display: none;
     position: absolute;
-    bottom: -80px;
+    bottom: 100%;
     right: 0;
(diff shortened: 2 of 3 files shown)


bfalling (approved):
I wish we had a good way to do this more deterministically, i.e. via states rather than relying on timeouts.

shaneriley:
Because the element needs to first be visible before starting the animation, the two state changes need to happen in series. The transition from on screen to off and then to hidden can be done as a transition end event listener and callback, but that approach didn't match the hidden to visible transition code and I'd rather have them look as similar as possible to illustrate they're just moving in opposite directions. If you'd rather, `ttsStop` can be changed to look like 
```js
  this.refs.$BRReadAloudToolbar.one('transitionend', function() {
    this.classList.remove('visible');
  }).removeClass('animate');
```

cdrini:
@shaneriley Could we do something like this? https://codepen.io/cdrini/pen/bGewoYO

shaneriley:
@cdrini your example doesn't animate back out, otherwise I would have done the same.

cdrini:
I personally don't think the animate-out is worth the extra code-complication. It might be worth getting feedback from Jim.

(See also: my most popular stackexchange response ever :P https://ux.stackexchange.com/questions/97760/why-are-animations-out-often-neglected-even-from-google/97789 )

ArunTeltia:
for the animation back effect you can use ```transition-timing-function: linear, step-end;```
check this pen 
https://codepen.io/arunteltia/pen/VwjKdNQ

shaneriley:
That still doesn't work when toggling the display property value. If the display property is set to none at either end of the state toggle, transitions are ignored because the element isn't technically rendered and therefore cannot animate properties that aren't applied to it.

ArunTeltia:
Okay, Thank you @shaneriley now I got the problem :D 
I complete misjudged the problem here
sorry about the wrong suggestion :P

PR #503, opened 2020-10-21
Fix 1up fullscreen loading

iisa:
Problem: starting 1up mode in fullscreen did not start at the requested page.

### Primary issue:
- while the outer bookreader container changes size to fullscreen, somewhere in the `resize` chain changes the page.

### Current Solution:
- pin current index at start of fullscreen toggle
- **then** set it after entrance animation, so that the bookreader's containers are already enlarged.
  - this prevents extra calls that change the page due to resizing functionality
- added a boolean to whether or not we set the keyboard Fullscreen Escape shortcut: `bindKeyboardControls`
  - this is to accommodate /stream page's basic default fullscreen view (they never leave it)
    - /stream page will now use main fullscreen toggle fn: `br.toggleFullscreen` to initiate Fullscreen instead of doing a work around
- some code rearrangement, to move the ESC keypress bind out of `enterFullscreen` & `exitFullscreen`
- unit tests around `toggleFullscreen`

### Testing:

go to: https://www-isa.archive.org/stream/chrysostom_pauline_homilies_field_vol_5#page/n34/mode/1up => goes to n34

go to: https://www-isa.archive.org/details/chrysostom_pauline_homilies_field_vol_5/page/n34/mode/1up?view=theater => ultimately goes to n34

(internal: https://webarchive.jira.com/browse/WEBDEV-3912)

### DIFF
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -1584,19 +1584,43 @@ BookReader.prototype.isFullscreen = function() {
   return this.isFullscreenActive;
 };
 
-BookReader.prototype.toggleFullscreen = function() {
-  if (this.isFullscreen()) {
+/**
+ * Fullscreen mode manager
+ * - toggles fullscreen
+ * - binds keyboard controls
+ * - fires custom event
+ * @param { boolean } bindKeyboardControls
+ */
+BookReader.prototype.toggleFullscreen = function(bindKeyboardControls = true) {
+  this._fullscreenCloseHandler = function _fullscreenCloseHandler(e) {
+    if (e.keyCode === 27) this.toggleFullscreen();
+  }.bind(this);
+
+  const handleExit = () => {
     this.exitFullScreen();
-  } else {
-    this.enterFullscreen();
+    $(document).unbind('keyup', this._fullscreenCloseHandler);
   }
+
+  const handleEnter = () => {
+    this.enterFullscreen();
+    if (bindKeyboardControls) {
+      $(document).keyup(this._fullscreenCloseHandler);
+    }
+  }
+  if (this.isFullscreen()) {
+    handleExit();
+  } else {
+    handleEnter();
+  }
+
   this.trigger('fullscreenToggled');
 };
 
 BookReader.prototype.enterFullscreen = function() {
+  const currentIndex = this.currentIndex();
   this.refs.$brContainer.css('opacity', 0);
 
-  var windowWidth = $(window).width();
+  const windowWidth = $(window).width();
   if (windowWidth <= this.onePageMinBreakpoint) {
     this.switchMode(this.constMode1up);
   }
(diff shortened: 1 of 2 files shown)


src/js/BookReader.js
```diff
@@ -1584,44 +1584,61 @@ BookReader.prototype.isFullscreen = function() {
...
+ * Fullscreen mode manager
+ * - toggles fullscreen
+ * - binds keyboard controls
+ * - fires custom event
+ * @param { boolean } bindKeyboardControls
+ */
+BookReader.prototype.toggleFullscreen = function(bindKeyboardControls = true) {
+  this._fullscreenCloseHandler = function _fullscreenCloseHandler(e) {
+    if (e.keyCode === 27) this.toggleFullscreen();
+  }.bind(this);
```
cdrini, line 1597: This will be re-defined every time `toggleFullscreen` is called, which would make the `unbind` always fail (because it's a new function each time). We should also be able to use arrow fns here to avoid `bind`-ing.

```suggestion
  this._fullscreenCloseHandler = this._fullscreenCloseHandler || (e) => {
    // ESC key
    if (e.keyCode === 27) this.toggleFullscreen();
  };
```

src/js/BookReader.js
```diff
@@ -1584,44 +1584,61 @@ BookReader.prototype.isFullscreen = function() {
...
+  }
+  const handleEnter = () => {
     this.enterFullscreen();
+    if (bindKeyboardControls) {
+      $(document).keyup(this._fullscreenCloseHandler);
+    }
+  }
+
+  if (this.isFullscreen()) {
+    handleExit();
```
cdrini, line 1610: These two new functions, `handleExit` and `handleEnter` are a little confusing since there are already `enterFullscreen` and `exitFullScreen` functions. Can we dissolve them and place the keyup bind/unbind in `enterFullscreen` and `exitFullScreen`? Also, that code should be run whenever full screen is entered; not just when it's toggled.

tests/BookReader/BookReaderPublicFunctions.test.js
```diff
@@ -7,6 +7,48 @@ afterEach(() => {
   jest.restoreAllMocks();
 });
 
+describe('BookReader.prototype.toggleFullscreen', ()  => {
```
cdrini, line 10: Woot woot unit test! :)

cdrini (changes requested):
This could some subtle issues if code call the `enterFullscreen` directly instead of `toggle`; let's make sure the keybinding also happens when `enter` is called! You can pass the `bindKeyboardControls` from `toggle` to `enter`, and you can define `_fullscreenCloseHandler` inside the `enter` function using the `||` to ensure it's only defined once.

iisa:
hey @cdrini - thanks for the code review, i have addresse the points we mentioned during our chat:
- `toggleFullscreen` method stays slim
- pass through `bindKeyboardControls` boolean to `enterFullscreen`
- move event triggers to `enterFullscreen` and `exitFullscreen`

we should be good to go. behavior is still the same in `www-isa.archive.org/details/goody/page/n22?view=theater`  - which is good enough for this round 👍

iisa:
conferring with @cdrini - good to go..

PR #504, opened 2020-10-23
Add TransformManager for continuous zoom

cdrini:
Still a number of bugs:
- Doesn't work in 1up
- Doesn't replace old zoom where possible
- Doesn't work in chrome with trackpad pinch-zoom


cdrini:
Closed by #806

PR #524, opened 2020-10-28
Replace colorbox and touch-punch and jquery.browser copies with npm dependencies

cdrini:


iisa (approved):
OK.
Q: what is color box used for?
Q: can we import touch punch where we put the mmenu? 

just some questions around dependency and import locations, other than that LGTM

cdrini:
Colorbox is used for the modal dialogs (e.g. share/info dialogs). I don't really fully understand how/why touch-punch is used yet 😅 But it's used on archive.org, so I added it to all the demos. Finding a place where we could import it via JS would be great!

PR #527, opened 2020-10-30
Replace mmenu copy with npm dep

cdrini:
Note the copy script/logic is temporary until v5, where this will be deleted from the folder and imported as js/css (#528). Even more temporary considering we're not planning on using mmenu for future versions!

- Tested that the copied files were the same as what was already there (except for the one which wasn't, which I excluded from the copy)


scripts/copy-modules.js
```diff
@@ -0,0 +1,10 @@
+const cpx = require('cpx2');
+
+// mmenu
+cpx.copySync('node_modules/jquery.mmenu/dist/{css,js}/**/*', 'BookReader/mmenu/dist');
+cpx.copySync('node_modules/jquery.mmenu/dist/addons/navbars/*', 'BookReader/mmenu/dist/addons/navbars');
+cpx.copySync('node_modules/jquery.mmenu/dist/addons/offcanvas/*', 'BookReader/mmenu/dist/addons/offcanvas');
+cpx.copySync('node_modules/jquery.mmenu/dist/addons/screenreader/*', 'BookReader/mmenu/dist/addons/screenreader');
+cpx.copySync('node_modules/jquery.mmenu/dist/addons/searchfield/*', 'BookReader/mmenu/dist/addons/searchfield');
+// This file has custom modifications -_- We'll delete it in v5 since it's unused.
+// cpx.copySync('node_modules/jquery.mmenu/src/addons/searchfield/jquery.mmenu.searchfield.js', 'BookReader/mmenu');
```
iisa, line 10: do we need this commented out `copySync`?

iisa (approved):
1 question, otherwise LGTM

scripts/copy-modules.js
cdrini: This file is the only file that's _not_ getting copied, so I though it would be nice to note it as such :)

PR #528, opened 2020-10-30
Remove mmenu public files

cdrini:
We only need mmenu for mobile nav, so only import it there.

#527 needs to go first; you just need to look at the last commit here.


src/js/plugins/plugin.mobile_nav.js
```diff
@@ -1,15 +1,11 @@
...
- * NOTE additional script and style tags must be included.
- *
- * <script src="../BookReader/mmenu/dist/js/jquery.mmenu.min.js"></script>
- * <script src="../BookReader/mmenu/dist/addons/navbars/jquery.mmenu.navbars.min.js"></script>
- * <link rel="stylesheet" href="../BookReader/mmenu/dist/css/jquery.mmenu.css" />
- * <link rel="stylesheet" href="../BookReader/mmenu/dist/addons/navbars/jquery.mmenu.navbars.css" />
  */
 
 import * as utils from '../BookReader/utils.js';
+import 'jquery.mmenu/dist/js/jquery.mmenu.min.js';
```
iisa, line 7: yasss!!!!

iisa (approved):
LGTM

PR #531, opened 2020-10-30
Remove dependOn for jquery and friends

cdrini:
dependOn let's webpack include code/polyfills in shared nodes of the dependency tree. But since a client (and in fact IA) might be using its own copy of jquery, this was causing colorbox to break when integrated into IA.

This causes a small jump in file size--and a peculiarly large jump in file size to jquery.browser.min.js 3KB -> 98KB! So I disabled it for this one and will confirm that it is still working correctly on IA.

This won't matter once we import most of these via JS.


cdrini:
Confirmed working; auto-merging this as a hotfix since it blocks a release and fixes a bug in master.

PR #532, opened 2020-10-31
Fix some text selection issues

cdrini:
Fix:
- Defining both full djvu xml url and single page djvu xml url causes neither to work; broke textsel on netlify
- No longer need options overrides now text sel is in prod
- Avoid requests for special index=-1 cover; causing a bunch of errors in Sentry


iisa (approved):
LGTM

PR #545, opened 2020-11-03
Refactor/move dragscroll to src

cdrini:
This file has way too many modifications from the library; it's src code.


jbuckner (approved):
Code looks fine. Where was dragscroll living before? I don't see the "old" one.

cdrini:
It was living in the `BookReader` dir (i.e. the "dist" or "build" dir). Now it's generated into that directory.

PR #546, opened 2020-11-03
Fix: Tests should import dragscrollable from src

cdrini:


cdrini:
Auto-merging this one; should've been part of #545, and only touches tests.

PR #552, opened 2020-11-09
Fix/samsung br not loading

cdrini:


cdrini:
Auto-merging; trivial change, and hotfixes production issue.

PR #553, opened 2020-11-09
Circumvent bug in BR JSIA code

cdrini:
Sentry was added to IA, and they code like `if (Sentry) { ... }`. Should be `if (window.Sentry)`. Temp fix until that's resolved.


cdrini:
Merged as part of #552

PR #555, opened 2020-11-10
Update Text selection error handling to handle empty xml file

iisa:
### Problem: 
Some items do not have a dejavu.xml, and when that happens, the xml parser borks inside of promise chain.

#### Repro:
- go to: https://archive.org/stream/EQuran00003/E-Quran-00003#page/n54/mode/1up
- open console - you see lots of errors
image

### Solution:
- wrap response in a try/catch to parse incoming XML

Go to: https://www-isa2.archive.org/stream/EQuran00003/E-Quran-00003#page/n54/mode/1up

Error no longer happens in console

Go to a book that has dejavu.xml available:
- https://www-isa2.archive.org/stream/goody#page/n22/mode/1up
- you can select text

### DIFF
--- a/tests/plugins/plugin.text_selection.test.js
+++ b/tests/plugins/plugin.text_selection.test.js
@@ -20,6 +20,7 @@ const FAKE_XML_MULT_WORDS = `<OBJECT data="file://localhost//tmp/derive/goodytwo
 <WORD coords="1600,2768,1700,2640">test3</WORD></LINE></PARAGRAPH></OBJECT>`;
 const FAKE_XML_5COORDS = `<OBJECT data="file://localhost//tmp/derive/goodytwoshoes00newyiala//goodytwoshoes00newyiala.djvu" height="3192" type="image/x.djvu" usemap="goodytwoshoes00newyiala_0001.djvu" width="2454">
 <PARAGRAPH><LINE><WORD coords="1216,2768,1256,2640,2690">test</WORD></LINE></PARAGRAPH></OBJECT>`;
+const FAKE_XML_EMPTY = '';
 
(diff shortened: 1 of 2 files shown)


tests/plugins/plugin.text_selection.test.js
```diff
@@ -20,6 +20,7 @@ const FAKE_XML_MULT_WORDS = `<OBJECT data="file://localhost//tmp/derive/goodytwo
 <WORD coords="1600,2768,1700,2640">test3</WORD></LINE></PARAGRAPH></OBJECT>`;
 const FAKE_XML_5COORDS = `<OBJECT data="file://localhost//tmp/derive/goodytwoshoes00newyiala//goodytwoshoes00newyiala.djvu" height="3192" type="image/x.djvu" usemap="goodytwoshoes00newyiala_0001.djvu" width="2454">
 <PARAGRAPH><LINE><WORD coords="1216,2768,1256,2640,2690">test</WORD></LINE></PARAGRAPH></OBJECT>`;
+const FAKE_XML_EMPTY = '';
```
jbuckner, line 23: So we get back an empty string instead of a 404 for missing dejavu files?

jbuckner (approved):
One question, but looks good 👍

tests/plugins/plugin.text_selection.test.js
iisa: yes. I took a look at the API server code and the logic is quite linear (+ python), it isn't a quick thing to sort out 😅  
But the smoking gun was empty XML file; this lil tweak handling the empty case will help us contain the non-render-blocking errors.

PR #556, opened 2020-11-12
Search Plugin Updates

iisa:
1. always display search results markers (adding them to mobile) - WEBDEV-3921
2. make search results nav tap targets at least 30px by 30px - WEBDEV-3982
3. extend click events
  - add new click event to help toggle side menu
  - add new test file for Search view
  - move search plugin test files to their own directory
  - parts of WEBDEV-3981, WEBDEV-3983

PR #557, opened 2020-11-13
Clear field and search result when click on X button in side pane - 3984

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-3984


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@internetarchive/bookreader",
-  "version": "4.20.1",
+  "version": "4.20.2",
```
iisa, line 3: hi @nsharma123 - let's not hard code the version for BookReader Repo as we have a formal release process.

If you can, please rsync the build directory to your review app for testing and QA. thanks!

src/js/plugins/search/view.js
```diff
@@ -525,7 +525,8 @@ class SearchView {
       .on(`${namespace}SearchCallbackError`, this.handleSearchCallbackError.bind(this))
       .on(`${namespace}SearchCallbackBookNotIndexed`, this.handleSearchCallbackBookNotIndexed.bind(this))
       .on(`${namespace}SearchCallbackEmpty`, this.handleSearchCallbackEmpty.bind(this))
-      .on(`${namespace}pageChanged`, this.updateSearchNavigation.bind(this));
+      .on(`${namespace}pageChanged`, this.updateSearchNavigation.bind(this))
+      .on(`${namespace}SearchResultsCleared`, this.clearSearchFieldAndResults.bind(this));
```
iisa, line 529: For BookReader's event system, we are using a 1 way directional flow.  Only BookReader emits events, and its consumer will use BR's public functions to update bookreader's state.

This is actually a "mirror" to some of my fixes in my PR #556.

let's sync on Monday to make sure we are on the same page around these quick fixes. thanks!

iisa (changes requested):
see comments. thank you!

package.json
nsharma123: Its just come in this commit, I was not indent to coming this here. Will remove

src/js/plugins/search/view.js
nsharma123: Yeah, I checked your PR that time but just waiting for you to merge so that I can use that event :) and them my whole week off :wink:

Issue #460, opened 2020-09-30
Move `pageChanged` event name into events.js


iisa:
Task:

- Update events 'list' to include new field: `pageChanged = 'pageChanged'`;
- find all the places `pageChanged` is used, and point it to its new reference (`events.pageChanged`)
- Make sure that the event `BookReader:pageChanged` is emitted when changing the page
- Make sure all tests pass

https://github.com/internetarchive/bookreader/blob/master/src/js/BookReader/events.js

nynaalekhya:
Hi, 
I would like to take up this issue.  Could this be assigned to me @iisa

iisa:
hi @nynaalekhya i apologize for not getting you assigned quickly.  We will go with the current PR that is open since it is being worked on.
Would you like to take this one instead? https://github.com/internetarchive/bookreader/issues/487

ArunTeltia:
@iisa mam can you assign a issue to me as well please

DhairyaBahl:
@iisa if this issue is still available kindly assign it to me. Thanks :)

Chaitanya31612:
@iisa Please review my pull request and guide for further changes. #559

Chaitanya31612:
> @iisa Please review my pull request and guide for further changes. #559

@iisa I've made pull request of requested changes. please review it and guide for any other change whatsoever is required.

harshhx:
I would like to take this issue.

hkumar0132:
I would like to work on this issue if it's still open @iisa 
I have been doing web development in javascript for a year now and would like to enter the open source world.

codewithsantosh:
Mam,Can i solve this issue

vishwjeet-singh:
Mam, May I work on this issue?

Priyaraj17:
Is this issue open? I would like to work on it.

GeraAlcantara:
close the issue please if is resolve

VishavSingla:
Is this issue still open? I would like to work on it.

devansh-dek:
Is the issue still open? Can I work on It?

PR #559, opened 2020-11-22
Move  event name into events.js #460

Chaitanya31612:
Done the following changes:

- Update events 'list' to include new field: pageChanged = 'pageChanged';
- find all the places pageChanged is used, and point it to its new reference (events.pageChanged)
- Make sure that the event BookReader:pageChanged is emitted when changing the page
- Make sure all tests pass

### DIFF
--- a/src/js/BookReader/events.js
+++ b/src/js/BookReader/events.js
@@ -17,4 +17,5 @@ export const EVENTS = {
   /* currently 3 represents thumbnail view */
   '3PageViewSelected': '3PageViewSelected',
   mobileNavOpen: 'mobileNavOpen',
+  pageChanged = 'pageChanged'
 };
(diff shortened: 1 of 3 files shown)


src/js/BookReader/events.js
```diff
@@ -17,4 +17,5 @@ export const EVENTS = {
   /* currently 3 represents thumbnail view */
   '3PageViewSelected': '3PageViewSelected',
   mobileNavOpen: 'mobileNavOpen',
+  pageChanged = 'pageChanged'
```
iisa, line 20: hi, i believe this will break the object.

iisa (changes requested):
please update issue with fix request in comment.

Also, please use format that the other `trigger` use in passing in value.

src/js/BookReader/events.js
Chaitanya31612: Ok, making the required changes. Thanks

Chaitanya31612: @iisa I've made the required change, please review it.
Thanks.

Chaitanya31612:
> please update issue with fix request in comment.
> 
> Also, please use format that the other `trigger` use in passing in value.

I've mentioned the pull request link in the comment of the issue. Also, I've changed the format for passing the value to as used in other places in the same file.

PR #560, opened 2020-11-24
Suppress framment changeing should be prevent on initial pin click

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-3766

### DIFF
--- a/src/js/plugins/search/plugin.search.js
+++ b/src/js/plugins/search/plugin.search.js
@@ -378,6 +378,7 @@ BookReader.prototype._searchPluginGoToResult = async function (pageIndex) {
       makeUnviewableAtEnd = true;
     }
   }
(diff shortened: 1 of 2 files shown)


src/js/plugins/search/plugin.search.js
```diff
@@ -378,6 +378,7 @@ BookReader.prototype._searchPluginGoToResult = async function (pageIndex) {
       makeUnviewableAtEnd = true;
     }
   }
+  this.suppressFragmentChange = false;
```
iisa, line 381:
```suggestion
  /* this updates the URL */
  this.suppressFragmentChange = false;
```

iisa (approved):
LGTM @nsharma123 

I added a comment above to note what is happening.

PR #572, opened 2020-12-01
Added more controls in toobar on desktop width

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-3987


iisa (commented):
checked in with @nsharma123 directly and we will: 
- remove `window` dependency in css (media query, etc)
- update render code so that it's easily switchable between limited & full nav options.

BookReader/BookReader.css
```diff
@@ -1030,10 +1030,16 @@ i.BRicon {
 .BRfooter button:hover .tooltip {
   display: block;
 }
+.BRfooter button.viewmode {
```
iisa, line 1033: this is a built file, we need to remove this bc it gets generated during versioning. let's get this out of the PR.

iisa (commented):
looks good! we just need to remove the built CSS file that got committed.

PR #574, opened 2020-12-07
Faster serve dev (>10x!)

cdrini:
- Use webpack dev mode for dev server
- Make live-server only watch the files it needs to watch
- Make webpack ignore built files
- Speed up jest --watch by making it only watch src/tests dirs (My VSCode jest extension no longer crashes!)

Making a JS change went from ~10-20s to cause a refresh to ~600ms :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,8 +58,12 @@
     "setupFiles": [
       "./env-setup.js"
     ],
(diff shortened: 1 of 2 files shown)


iisa (approved):
🍰

PR #575, opened 2020-12-07
Continuous Zoom Preparatory Refactor

### LINKED ISSUE #270: Separate Mode1Up out of BookReader.js into own file
There are a lot of methods inside BookReader.js that are only relevant to Mode1Up. Let's pull these out into a separate file, Mode1Up.js (similar to Mode2Up.js).

## Summary of Requirements
- Law of refactor: **No observable changes to UI or public APIs**. Need to be very strict about this, since BookReader is used by more than just IA, and we don't want to throw a breaking change as part of this.
- The new file should pass eslinting
- There should be a new jest file for the new file with at least 1 test.

## Process
Each step should be a separate commit. This type of refactor can easily become very complicated, these steps ensure that (1) each commit is in a run-able state, so you can test the site, and (2) that the git history is organized enough to easily see what changes were made when the code was actually modified. This _greatly_ reduces the complexity, risk of error, mental load, and time need to complete the refactor.

**Note:** Do not include the built JS files inside any of these commits. Feel free to add them as separate commits before pushing. That will let the run with testcafe/travis.

1. Create a new file `Mode1Up.js`, that has a function `exposeMode1Up(BookReader)`, and cut all the appropriate methods into the body of this function. **Make no other modifications to the code**. Inside BookReader.js, import that function and call it. Confirm the site works.
2. Eslint the new file, and fix any violations.
3. Add docstring with type annotations to every method.
4. Convert to an es6 class. This is going to be a big step. See https://github.com/internetarchive/bookreader/pull/237/commits/3a751c2d31001438caec6902ecfffc5753994270 for an example. Inside BookReader.js, you need to expose the class' methods for overriding (see the commit above), and add the mode into `_modes`. Confirm the site works.
5. Fix type errors. A refactor can _easily_ introduce type errors, typos, etc. Add `// @ts-check` to the top of the file, and fix as many type errors as possible. Confirm the site works.
5. Create a new jest test file for the new file.

cdrini:
Closes #270 

- Es6ify + add some tests to some zoom related functions
- Pull out Mode1Up into its own file

### DIFF
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -1684,6 +1684,9 @@ exposeOverrideableMethod(Mode1Up, '_modes.mode1Up', 'prepare', 'prepareOnePageVi
 /** @deprecated not used outside BookReader */
 BookReader.prototype.drawLeafsOnePage = Mode1Up.prototype.drawLeafs;
 exposeOverrideableMethod(Mode1Up, '_modes.mode1Up', 'drawLeafs', 'drawLeafsOnePage');
+/** @deprecated not used outside BookReader */
+BookReader.prototype.zoom1up = Mode1Up.prototype.zoom;
+exposeOverrideableMethod(Mode1Up, '_modes.mode1Up', 'zoom', 'zoom1up');
 
 /************************/
 /** Mode2Up extensions **/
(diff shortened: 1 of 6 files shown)


jbuckner (approved):
Nice job! Looks good 👍

PR #577, opened 2020-12-18
Replace /stream/ URLs with equivalent /details/ URLs

nsharma123:
JIRA:WEBDEV-4070

This is part of /stream/ to /details/ redirection stuff. Currently #page/n1/mode/2up not being converted in sub-path something like /page/n1/mode/2up.

These changes is to convert the URL from hash segment to sub-path sagment.


iisa (approved):
👍

iisa (approved):
👍

Issue #275, opened 2020-04-01
Investigate IIIF demo error


iisa:
### Problem:
The IIIF demo page in ./BookReaderDemo/demo-iiif.html throws an error as follows:

```
Uncaught TypeError: Cannot read property 'width' of undefined (IIIFBookReader.js:69)
    at L.brInstance.getPageHeight (IIIFBookReader.js:69)
    at e.<computed> (classes.js:31)
    at e.value (BookModel.js:72)
    at e.value (Mode2Up.js:376)
    at e.value (Mode2Up.js:473)
    at L.s (classes.js:19)
    at L.switchMode (BookReader.js:1441)
    at L.init (BookReader.js:365)
    at Object.success (IIIFBookReader.js:120)
    at fire (jquery-1.10.1.js:3074)
```

### Task:
- [x] Confirm the task by finding the last commit before all of the plugins were converted to ES6 and visiting the demo page
- [ ] Update this issue to raise flag and discuss next steps.

kristopolous:
It's caused by the page number not being set (currentIndexL) prior to making the call. then it gets an NaN for indexing an array and we see this issue. It looks like either there needs to be a default value, this initial case has to be accounted for, or the order of the operations need to change ... I'll see if I can find a really low impact fix for this.

spacejamjim:
@kristopolous I understand there are no promises, but thank you for finally picking this up. The IIIF feature would be an amazing addition

Pyrojet99:
- https://github.com/internetarchive/bookreader/commit/f2c52671be541b4ef5828133a65059971582dc08 is the last working commit for IIIF demo

- https://github.com/internetarchive/bookreader/commit/03c10ef1162995821c1c1cba195533171e8e37dc IIIF demo does not work, but the error is not shown either

- https://github.com/internetarchive/bookreader/commit/d7335d4710f26df24b078c29ae10908db10de2b0 first commit in which IIIF demo shows the error

(these 3 commits are sequential)

spacejamjim:
Most likely already known but the `IIIFBookReader.js` as it stands uses IIIF 1.0 `full/full/0/native.jpg` versus the `default.jpg` which would be a quick fix. 
Also,  thanks a ton.

PR #581, opened 2020-12-22
fix IIIF demo error

nynaalekhya:
@iisa 
IIIF demo error fix
iifdemo


nynaalekhya:
@iisa  
Please take a look at this. Closes #275  commit 638be16

PR #607, opened 2021-01-07
4.26.1

iisa:
https://github.com/internetarchive/bookreader/releases/new?tag=v4.26.1&title=v4.26.1&body=Patch%3A%20Text%20selection%20layer%20does%20not%20get%20made%20in%20thumbnail%20view%20%40dualcnhq


iisa:
GTG w @cdrini

PR #608, opened 2021-01-12
Port book navigator

dualcnhq:


jbuckner:
Just a thought, but since the Book Navigator has a different development style than BookReader and has its own set of dependencies, have you considered having a separate Book Navigator repo that pulls in BookReader via NPM?

cdrini:
Yeah; we want BR to be more like BookNavigator in terms of development style longer term :P And neither one can exist without the other, so separating adds unnecessary development friction.

cdrini:
ItemNavigator we likely want to pull into a separate repo at some point though, but could be a secondary step.

package.json
```diff
@@ -22,7 +22,21 @@
     "url": "https://github.com/internetarchive/bookreader/issues"
   },
   "homepage": "https://github.com/internetarchive/bookreader#readme",
-  "dependencies": {},
+  "dependencies": {
+    "@internetarchive/ia-book-downloads": "^0.1.5",
```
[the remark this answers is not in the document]
jbuckner: Don't you need the dependencies to be here to get them to be included with the built package?

src/js/ia-topnav/ia-topnav.js
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable no-undef */
```
cdrini, line 1: Ooooh, ia-topnav is too much :P We don't this one.

package.json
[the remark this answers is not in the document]
cdrini: Yes and no-ish? Currently BR compiles into single JS files which are imported in petabox as `<script>` tags. Longer term we likely want to change that so that it can be imported like a more traditional npm repo, but we currently don't do that.

src/js/ia-topnav/ia-topnav.js
dualcnhq: noted

BookReaderDemo/demo-internetarchive.html
```diff
@@ -2,70 +2,74 @@
...
 
-<div id="BookReader">
-    Internet Archive BookReader Demo<br/>
-    <noscript>
-    <p>
-        The BookReader requires JavaScript to be enabled. Please check that your browser supports JavaScript and that it is enabled in the browser settings.
-    </p>
-    </noscript>
-</div>
+      <div id="bookreader-container" class="xs-col-12">
```
cdrini, line 42: `xs-col-12` looks like boostrap; I don't think this will work in BR? We might be able to remove it.

BookReaderDemo/demo-internetarchive.html
```diff
@@ -2,70 +2,74 @@
...
+
+  <!-- CSS only -->
+  <link rel="stylesheet" href="BookReaderDemo.css" />
+
+  <link rel="stylesheet" href="../BookReaderTemplate/styles/BookReader-ia.css" />
+
 </head>
+
 <body>
+  <div id="theatre-ia">
```
cdrini, line 39: Can we have these elements created via JS? That way the interface to set up BookReader is the same (you just need a `#BookReader` element)

BookReaderTemplate/BookNavigator/BookNavigator.js
```diff
@@ -0,0 +1,332 @@
+import { html, LitElement } from 'lit-element';
```
cdrini, line 1: Let's place these in `src/`; code always lives in the `src` directory here. (e.g. `src/BookNavigator`, `src/iaux-foo-bar`). We'll at some point in the future want to remove the `js` directory, since things are organized by structural component instead of by file type :+1: (If that causes issues, we can also do e.g. `src/templates/BookNavigator`)

BookReaderTemplate/BookNavigator/util/debouncer.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Wait until some time has passed before executing a callback.
+ *
+ * @param {Function} callback
```
cdrini, line 4: Note to self: In a future PR we should dissolve this to use the existing debounce function.

cdrini (changes requested):
- Rebase your branch; might be a touch messy because it looks like some more commits have jumped in. If you do:

```
# Get up-to-date master-v5 branch
git branch -D master-v5
git fetch origin master-v5

git checkout my-branch
git rebase -i master-v5
```

Make sure to remove all commits except your first commit (" Initial setup of moving bookNav files to bookreader ") and you should be good! You might want to make a backup of this branch `git checkout -b port-booknavigtor--backup` just in case :)

BookReaderDemo/demo-internetarchive.html
dualcnhq: removed this one

BookReaderTemplate/BookNavigator/BookNavigator.js
dualcnhq: updated now

package.json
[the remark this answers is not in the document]
dualcnhq: updated

src/BookReaderTemplate/BookReaderTemplate.js
```diff
@@ -0,0 +1,98 @@
...
+      }
+    </style>
+
+    <div id="theatre-ia-wrap" class="container container-ia width-max ">
+      <div id="theatre-ia" class="width-max">
+        <div class="row">
+          <div class="xs-col-12">
+            <div id="IABookReaderMessageWrapper" style="display:none;"></div>
+            <item-navigator
+              item=${encJson}
```
cdrini, line 78: This is currently using a dummy metadata file (see sample.json). This means it is out of sync with what's actually displayed in the BR 😬 (You can see this by clicking the "Share > Twitter" button, and see that it uses a ComputerWorld link instead of Plato https://deploy-preview-608--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala )

src/BookReaderTemplate/BookReaderTemplate.js
```diff
@@ -0,0 +1,98 @@
...
+              </div>
+            </item-navigator>
+          </div>
+        </div>
+      </div>
+    </div>
+  `;
+
+// Render the template to the document
+debugger;
```
cdrini, line 97:
```suggestion
```

src/iaux-bookmark-edit/src/styles/ia-bookmark-edit.js
```diff
@@ -0,0 +1,124 @@
+import { css } from 'lit-element';
```
cdrini, line 1: Note to self: Isa is actively changing some of the bookmark files, I believe? Might need to update these soon.

webpack.config.js
```diff
@@ -1,26 +1,24 @@
 // @ts-check
```
cdrini, line 1: Want to drop the style changes on this one; this file tends to get edited a decent amount, so will likely cause annoying merge conflicts.

cdrini:
So primary goal: we want to be able to delete these files from petabox, and have it use the code here.

In order to do that we need BookReader to be initialized by petabox with all the panels, etc.

- Q: In petabox, if we remove all the files, how would we specify the metadata for item-navigator?
- Q: What will petabox call? It's currently essentially doing what BookReaderTemplate.js is doing now. How much of the responsibility for initializing do we now want inside BR? Ideally, we want third-party users of BookReader (non-IA people) to be able to quickly initialize a BookReader that looks ~identical to what IA is using, which suggests we likely want most of the init logic here.
    - So if that's what we want, then we need an easy way to init it.
    - Maybe a good next step is to call BookReaderTemplate `render` from `BookReader.init`? At that point all the dynamic info (e.g. metadata) will be ready.
    - Then we remove item-navigator/etc from petabox, and just essentially call BookReader.init :/

Might be useful to create a petabox branch and start deleting these files, and see what breaks and what hooks need to add to make the system work.

dualcnhq:
moving changes to this [PR](https://github.com/internetarchive/bookreader/pull/629)

PR #609, opened 2021-01-19
Added alt attribute on bookreader pages

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-4059

Added alt attributes on all bookreader images including loader, 1up, 2up and thumbnail images.


jbuckner:
Shouldn't these `alt` tags have some content to them? They're all blank.

nsharma123:
> Shouldn't these `alt` tags have some content to them? They're all blank.

I think we don't sufficient context to provide in alt='', that's why all are blank https://webaim.org/techniques/alttext/#context

jbuckner:
I think when they're talking about context there, they are saying that if the image already has enough information around it, like a title and description, you can use an empty `alt` tag. The images in this PR look like placeholder images, which probably aren't labelled otherwise so they'd probably benefit from alt tags.

PR #611, opened 2021-01-21
initializer updates, new options: `autoResize` && `useSrcSet`

iisa:
- `autoResize = true` by default, sets in-br window change resizer
- `useSrcSet = true` by default, sets srcSets to images
+ move postInit event trigger after `this.init.initComplete = true`

### DIFF
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -489,18 +489,18 @@ BookReader.prototype.init = function() {
     }
   });
 
-  $(window).bind('resize', this, function(e) {
-    e.data.resize();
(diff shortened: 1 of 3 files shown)


jbuckner (approved):
Looks good 👍 

So this is giving the consumer more control over configuring BookReader's layout and images?

iisa:
> Looks good 👍
> 
> So this is giving the consumer more control over configuring BookReader's layout and images?

yep! this is in preparation to move this functionality to a ResizeObserver in v5 as well.. (exciting)

PR #614, opened 2021-01-29
BookReader: Show more UI at desktop widths

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-3987
Showing full controls on desktop.
Showing toggle button on mobile devices.

https://webarchive.jira.com/browse/WEBDEV-4042 (on mobile device)
Clicking on thumbnail leaf from toggle button should always open 1up mode.
Clicking on thumbnail leaf from toggle button should reflect correct UI in navbar.

### DIFF
--- a/src/js/BookReader/options.js
+++ b/src/js/BookReader/options.js
@@ -223,7 +223,7 @@ export const DEFAULT_OPTIONS = {
       className: 'thumb',
     },
     viewmode: {
-      visible: false,
+      visible: true,
       className: 'viewmode',
       excludedModes: [],
     },
--- a/src/js/BookReader/options.js
+++ b/src/js/BookReader/options.js
@@ -207,25 +207,58 @@ export const DEFAULT_OPTIONS = {
   getEmbedCode: null,
 
   controls: {
-    onePage: {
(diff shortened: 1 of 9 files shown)


src/js/BookReader/options.js
```diff
@@ -207,26 +207,59 @@ export const DEFAULT_OPTIONS = {
...
     thumbnail: {
-      label: 'Thumbnail view',
       visible: true,
+      label: 'Thumbnail view',
       className: 'thumb',
+      iconClassName: 'thumb'
     },
     viewmode: {
-      visible: false,
+      visible: true,
```
iisa, line 241: does `view mode` need a `label` field, too?

iisa (approved):
we will move navbar "switching" logic to its own function && call it during init & resize.

this way, we can "house" function inside of Navbar.js as well.

src/js/BookReader/options.js
nsharma123: No, it will automatically added later when BR is initialised and it know on what page view you are currently.

iisa (approved):

PR #623, opened 2021-02-03
PRs into master-v5 should also have tests run

cdrini:


cdrini:
Auto-merging; trivial devops thing

PR #624, opened 2021-02-03
Dont limit BR max-width in demo

cdrini:
Testcafe failing with `AssertionError: expected 800 to deeply equal 1366`


cdrini:
Auto-merging; tiny change only affects demos + fixes Testcafe \o/

PR #625, opened 2021-02-05
2up draw - reduction factors + making sure actual numbers are used when calculating

iisa:
Problem:
If container DOM is not set, the reduction factors of 2up mode are calculating against `NaN` and breaks all the things.

Solution:
- remove redundant `this.reduce` calculation
- pass `this.reduce` through to image url creation.

To Test: go to www-isa.archive.org/details/goody
open dev tools
refresh
look at BookReaderImages.php requests
confirm that the URL does not have scale=1

### DIFF
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -1680,9 +1680,9 @@ BookReader.prototype._scrollAmount = function() {
   return parseInt(0.9 * this.refs.$brContainer.prop('clientHeight'));
 };
 
-BookReader.prototype.prefetchImg = function(index, optionalReduction) {
-  var pageURI = this._getPageURI(index, optionalReduction);
-  const pageURISrcset = this.options.useSrcSet ? this._getPageURISrcset(index) : [];
+BookReader.prototype.prefetchImg = function(index) {
+  var pageURI = this._getPageURI(index, this.reduce);
+  const pageURISrcset = this.options.useSrcSet ? this._getPageURISrcset(index, this.reduce) : [];
 
   // Load image if not loaded or URI has changed (e.g. due to scaling)
   var loadImage = false;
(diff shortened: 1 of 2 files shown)


src/js/BookReader.js
```diff
@@ -1671,8 +1680,8 @@ BookReader.prototype._scrollAmount = function() {
   return parseInt(0.9 * this.refs.$brContainer.prop('clientHeight'));
 };
 
-BookReader.prototype.prefetchImg = function(index) {
-  var pageURI = this._getPageURI(index);
+BookReader.prototype.prefetchImg = function(index, optionalReduction) {
+  var pageURI = this._getPageURI(index, optionalReduction);
   const pageURISrcset = this.options.useSrcSet ? this._getPageURISrcset(index) : [];
```
cdrini, line 1685:
```suggestion
  const pageURISrcset = this.options.useSrcSet ? this._getPageURISrcset(index, this.reduce) : [];
```

src/js/BookReader.js
```diff
@@ -1671,8 +1680,8 @@ BookReader.prototype._scrollAmount = function() {
   return parseInt(0.9 * this.refs.$brContainer.prop('clientHeight'));
 };
 
-BookReader.prototype.prefetchImg = function(index) {
-  var pageURI = this._getPageURI(index);
+BookReader.prototype.prefetchImg = function(index, optionalReduction) {
+  var pageURI = this._getPageURI(index, optionalReduction);
```
cdrini, line 1684:
```suggestion
  var pageURI = this._getPageURI(index, this.reduce);
```

cdrini (approved):
Lgtm! Seems high priority since requesting the large images at this scale is causing perf issues :)

PR #627, opened 2021-02-05
Add cache for text selection page text

cdrini:
This was loading twice on init, and multiple times on resize. Let's cache it!

TODO: Add unit test


iisa (approved):
LGTM!

PR #629, opened 2021-02-09
Feature item nav

dualcnhq:
## Changes:

- moved BR iaux-book* components for submenus to `src` directory
- changed import references to point to `src` book-nav and item-nav components
- create entrypoint plugin for loading `<item-nav>`

----
**Note:**
- btoa encoding issue on private books


iisa (commented):
wow!!! brava!!
thank you @dualcnhq for the care and drive to push this forward. this is a solid pass. 💪 

I've got some suggestions, small ones, that I wanted to run by you sprinkled in the files.  LMK what you think

package.json
```diff
@@ -22,11 +22,22 @@
...
-  "dependencies": {},
   "private": false,
   "devDependencies": {
     "@babel/core": "7.12.9",
     "@babel/preset-env": "7.12.7",
+    "@internetarchive/ia-activity-indicator": "0.0.1",
+    "@internetarchive/ia-icons": "1.2.2",
+    "@internetarchive/ia-menu-slider": "0.1.10",
+    "@internetarchive/ia-sharing-options": "^0.1.2",
+    "@internetarchive/ia-topnav": "1.1.1",
```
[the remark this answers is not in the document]
dualcnhq: updated now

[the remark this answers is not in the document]
dualcnhq: will need to move `ia-sharing-options` to this repo and `ia-topnav` too

src/iaux-book-search-results/src/styles/icon_close.js
```diff
@@ -0,0 +1,3 @@
+import { css } from 'lit-element';
```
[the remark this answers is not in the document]
dualcnhq: it's not a double import. this code has the same output of `ia-icon/close` but the menu panel uses this file directly

src/iaux-book-downloads/src/ia-book-downloads.js
```diff
@@ -0,0 +1,68 @@
+import { html, LitElement } from 'lit-element';
```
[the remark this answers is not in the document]
dualcnhq: done

package.json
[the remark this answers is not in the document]
dualcnhq: going to keep `ia-sharing-options` then will remove `ia-topnav`

.gitattributes
```diff
@@ -1,24 +0,0 @@
-# keep these lines sorted in ascending order.
```
cdrini, line 1: Let's not delete this file :)

package.json
```diff
@@ -22,11 +22,21 @@
...
     "@babel/preset-env": "7.12.7",
+    "@internetarchive/ia-activity-indicator": "0.0.1",
+    "@internetarchive/ia-icons": "1.2.2",
+    "@internetarchive/ia-menu-slider": "0.1.10",
+    "@internetarchive/ia-sharing-options": "^0.1.2",
+    "@internetarchive/icon-bookmark": "1.1.3",
+    "@internetarchive/icon-collapse-sidebar": "^1.1.0",
+    "@internetarchive/icon-edit-pencil": "^1.1.5",
+    "@internetarchive/icon-magnify-minus": "^1.2.3",
+    "@internetarchive/icon-magnify-plus": "^1.2.3",
```
cdrini, line 37: We try to avoid `^` dependencies, so that all our environments are consistent (otherwise like if you npm install a month after me, you might get slightly different versions).

package.json
```diff
@@ -44,6 +54,8 @@
     "jquery-ui-touch-punch": "0.2.3",
     "jquery.browser": "0.1.0",
     "jquery.mmenu": "5.6.5",
+    "lit-element": "^2.4.0",
+    "lit-html": "^1.3.0",
```
cdrini, line 58: Ditto

src/BookReaderTemplate/template.js
```diff
@@ -0,0 +1,124 @@
+/**
```
cdrini, line 1: Call this file BookReaderTemplate.js to match the class and be consistent with how other files are named when they have "main" class (e.g. `BookNavigator.js`)

src/BookReaderTemplate/template.js
```diff
@@ -0,0 +1,124 @@
...
+    this.fetchData();
+  }
+
+  /**
+   * Fetch metadata response from public metadata API
+   * convert response to base64 data
+   * set base64 data to props
+   */
+  async fetchData() {
+    const ocaid = location.href.match(/ocaid=([^&#]+)/i)[1];
```
cdrini, line 92:
```suggestion
    const ocaid = new URLSearchParams(location.search).get('ocaid');
```

src/BookReaderTemplate/template.js
```diff
@@ -0,0 +1,124 @@
...
+    const url = 'https://archive.org/metadata/' + ocaid;
+    const response = await fetch(url);
```
cdrini, line 94:
```suggestion
    const response = await fetch('https://archive.org/metadata/' + ocaid);
```

src/css/_BRTemplate.scss
```diff
@@ -0,0 +1,7 @@
+#item-navigator-modal {
+  display: none
+}
+
+item-navigator-modal[mode=open] {
+  display: block;
+}
```
cdrini, line 7: Add comment about why these are necessary.

```suggestion
#item-navigator-modal, item-navigator-modal[mode=open] {
  display: block;
}
```

src/js/plugins/plugin.item_nav.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * plugin to load BookReaderTemplate
+ */
+import { BookReaderTemplate } from "../../BookReaderTemplate/template";
+
+window.customElements.define('bookreader-template', BookReaderTemplate);
```
cdrini, line 6: I was initially thinking we might want `BookReaderTemplate` constructed/initialized from within this file; and use the `BookReader.prototype.init` as a hook. That would let us support the current "interface" of `<div id="BookReader" />`. I think our main goal was to have it so that petabox could use bookreader "binaries" (everything in the BookReader folder), _but_ avoid using item navigator.

I think we're having trouble predicting the future on this one and getting stuck; let's run with this as is, and see what we need to change :+1:

cdrini (changes requested):

package.json
jbuckner: Isn't that what `package-lock.json` is for?

cdrini: If I (very vaguely) recall correctly, `package-lock.json` isn't used for `npm i`? So running `npm i` will install from package.json using the appropriate version ranges. Running `npm ci` will install from package-lock.json, but also remove any existing node_modules (read: slow). The petabox repo uses npm shrinkwrap, which does something else which I can't remember :P Pinning them in the package.json seemed like the easiest approach, that gets us 90% of what we need without complicating things too much.

cdrini: (90% because there is still some wiggle room for secondary dependencies having different versions, but meh; super unlikely to cause in issue if a secondary dependency has like a patch update. I believe the npm-shrinkwrap approach does cover that last 10%, but also adds some complexity. Makes sense for a larger repo like petabox, but for BR its not worth it.)

jbuckner: I looked into this and I think we should be using the `^` for these dependencies. When developers check out the the book reader repo, the `package-lock.json` file will guarantee that `npm install` installs the proper version, but still allows consumers to update transitive dependencies. 

By locking these versions down, we cannot install minor updates of these icons in petabox because BR is requiring these specific versions and we include these same icons in other dependencies. Locking things down in the `package.json` kind of breaks semantic versioning.

Here's a good article about the different purposes of the files: https://medium.com/@hossam.hilal0/package-json-vs-package-lock-json-vs-npm-shrinkwrap-json-33fcddc1521a

package.json
```diff
@@ -22,19 +24,29 @@
...
+    "@internetarchive/ia-icons": "1.2.2",
+    "@internetarchive/ia-menu-slider": "0.1.10",
+    "@internetarchive/ia-sharing-options": "0.1.2",
+    "@internetarchive/icon-bookmark": "1.1.3",
+    "@internetarchive/icon-collapse-sidebar": "1.1.0",
+    "@internetarchive/icon-edit-pencil": "1.1.5",
+    "@internetarchive/icon-magnify-minus": "1.2.3",
+    "@internetarchive/icon-magnify-plus": "1.2.3",
+    "@internetarchive/icon-search": "1.2.3",
+    "@internetarchive/modal-manager": "0.1.0",
```
jbuckner, line 41: _Question_ How come these are all in `devDependencies`?

cdrini: Because BookReader doesn't export as a normal npm module; it exports built files. So everything is a dev dependency in that sense. We'll need to change this at some point as we transition to allowing BR to be imported via JS.

package.json
cdrini: Good article! I don't think we should do that as part of this PR (unless it blocks the migration), but seems good to me :+1:

package.json
jbuckner: Hmm I think that's going to be a problem integrating this into petabox and offshoot. By doing this, it's going to register all of these custom elements twice, once inside petabox, then again in this package, and you get an error if you register a custom element twice.

iisa: +1 on that, @jbuckner.
the first phase is asset swap on PB, then creating a custom component to load in offshoot

package.json
jbuckner: Sorry, what seems good to you?

cdrini: Switching to `^`

package.json
jbuckner: Just closing the loop here after our chat on Friday. Since these, along with `lit-html` and `lit-element` below, are all dependencies of the exported `module`, they should go into `dependencies` so consumers can properly install them.

jbuckner (commented):
Two small things. I didn't go too far into the MR since it's all just copied over existing code. I think the only main thing is the `dependencies` in `package.json`

src/js/plugins/plugin.item_nav.js
dualcnhq: resolving for now, we can get back to this in the future

package.json
dualcnhq: resolving this one

iisa (approved):
LGTM!

iisa:
@dualcnhq will merge to `master-v5` branch when checks are complete

PR #631, opened 2021-02-12
exit fullscreen - switch to 2up if available

iisa:
Problem: even if 2up is not available, when exiting fullscreen, bookreader will put book at 2up mode

Solution: check to see if 2up mode is available before switching modes


bfalling (approved):
LGTM

PR #645, opened 2021-03-04
More image caching

iisa:
### Dev notes:
 we are doing our best to cancel any pending requests
 we are saving up to 50 images that 2up & thumbnail can use
 we reuse images often, and resize as much as possible.
 
### Testing:
load a book with network tab on..
flip through pages,
toggle side menu,
toggle fullscreen,
— if current image scale is better or equal, we keep the image
thumbnail view: go to thumbnail from 2up, remember your page -> inspect that page in thumbnail view, it should be the same URL + check in network tab for that URL - no secondary image request gets made

https://webarchive.jira.com/browse/WEBDEV-4255

### DIFF
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -140,7 +140,7 @@ BookReader.prototype.setup = function(options) {
 
   this.displayedIndices = [];
   this.imgs = {};
-  this.prefetchedImgs = {}; //an object with numeric keys cooresponding to page index
+  this.prefetchedImgs = {}; //an object with numeric keys cooresponding to page index, reduce
 
   this.animating = false;
   this.flipSpeed = options.flipSpeed;
@@ -1682,44 +1682,83 @@ BookReader.prototype._scrollAmount = function() {
   return parseInt(0.9 * this.refs.$brContainer.prop('clientHeight'));
 };
 
-BookReader.prototype.prefetchImg = function(index) {
-  var pageURI = this._getPageURI(index, this.reduce);
+/**
+ * Used by 2up
+ * Fetches the image for requested index & saves in `this.prefetchedImgs`
+ * Does not re-request if image is in the 
+ *
+ * @param {Number} index 
+ */
+BookReader.prototype.prefetchImg = async function(index, fetchNow = false) {
+  let pageURI = this._getPageURI(index, this.reduce);
   const pageURISrcset = this.options.useSrcSet ? this._getPageURISrcset(index, this.reduce) : [];
 
   // Load image if not loaded or URI has changed (e.g. due to scaling)
-  var loadImage = false;
+  let loadImage = false;
+  const wasPrefetchedSmaller = this.prefetchedImgs[index]?.reduce > this.reduce;
   if (undefined == this.prefetchedImgs[index]) {
     loadImage = true;
-  } else if (pageURI != this.prefetchedImgs[index].uri) {
+  } else if (wasPrefetchedSmaller) {
+    console.log("PREFETCH: curr reduce rate > br.reduce - index, this.prefetchedImgs[index]?.reduce, this.br.reduce: ", index, this.prefetchedImgs[index]?.reduce, this.reduce);
+    loadImage = true;
+  } else if (!wasPrefetchedSmaller && (pageURI != this.prefetchedImgs[index]?.uri)) {
     loadImage = true;
   }
 
-  if (loadImage) {
-    const pageContainer = this._createPageContainer(index);
-    $('<img />', {
-      'class': 'BRpageimage',
-      'alt': 'Book page image',
-      src: pageURI,
-      srcset: pageURISrcset,
-    }).data('reduce', this.reduce).appendTo(pageContainer);
-    if (index < 0 || index > (this._models.book.getNumLeafs() - 1) ) {
-      // Facing page at beginning or end, or beyond
-      pageContainer.addClass('BRemptypage');
+  if (!loadImage) {
+    return;
+  }
+
+  if (wasPrefetchedSmaller) {
+    console.log('wasPrefetchedSmaller', index, this.prefetchedImgs['index']);
+    $(this.prefetchedImgs[index]).find('img').attr('src', '');
+  }
+    
+  const $pageContainer = this._createPageContainer(index);
+  const $imgShell = this._modes.mode2Up.createPageImgShell(index, pageURI, this.reduce, $pageContainer);
+
+  /** set uri in img tag to start request & save in `this.prefetchedImgs` */
+  const fetchImageAndRegister = () => {
+    const $imgEl = $($imgShell).find('img');
+    $imgEl.attr('src', pageURI);
+    if (pageURISrcset.length) {
+      $imgEl.attr('srcSet', pageURISrcset);
     }
-    pageContainer[0].uri = pageURI; // browser may rewrite src so we stash raw URI here
-    this.prefetchedImgs[index] = pageContainer[0];
+    $($imgEl).load(() => {
+      console.log('** PREFETCH LOADED: index, this.reduce', index, this.reduce);
+    })
+    this.prefetchedImgs[index] = $imgShell;
+  };
+
+  if (fetchNow || (index == this.twoPage.currentIndexL) || (index == this.twoPage.currentIndexR)) {
+    console.log("*** PREFETCH NOW: - index, this.reduce ", index, this.reduce);
+    fetchImageAndRegister();
+  } else {
+    // stagger request
+    const time = 500;
+    setTimeout(() => {
+      console.log(`*** PREFETCH ${time}: staggered - index, this.reduce`, index, this.reduce);
+      fetchImageAndRegister();
+    }, time);
   }
 };
 
+/** used in 2up */
 BookReader.prototype.pruneUnusedImgs = function() {
+  console.log('___ PRUNE ___', this.br.reduce);
+  const prefetchIsCapped = Object.keys(this.prefetchedImgs).length == 50;
   for (var key in this.prefetchedImgs) {
     if ((key != this.twoPage.currentIndexL) && (key != this.twoPage.currentIndexR)) {
       $(this.prefetchedImgs[key]).remove();
     }
     if ((key < this.twoPage.currentIndexL - 4) || (key > this.twoPage.currentIndexR + 4)) {
-      delete this.prefetchedImgs[key];
+      if (prefetchIsCapped || (this.prefetchedImgs[key]?.reduce > this.reduce)) {
+        console.log("PREFETCH IS CAPPED or reduction rate is small -- key, prefetchIsCapped, this.prefetchedImgs[key]?.reduce", key, prefetchIsCapped, this.prefetchedImgs[key]?.reduce);
+        delete this.prefetchedImgs[key];
+      }
     }
   }
+  console.log('___ END PRUNE ___', this.br.reduce);
 };
 
 /************************/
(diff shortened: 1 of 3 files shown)


src/js/BookReader.js
```diff
@@ -889,12 +889,14 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
 
         const img = document.createElement("img");
         const thumbReduce = floor(book.getPageWidth(leaf) / this.thumbWidth);
-
+        // use prefetched img src first if previously requested & available img is good enough
+        const prefetchedImg = this.prefetchedImgs[leaf] || {};
+        const imageURI = prefetchedImg.reduce <= thumbReduce ? prefetchedImg.uri || this._getPageURI(leaf, thumbReduce);
```
iisa, line 894: Thumbnail: added check of prefetchedImgs before requesting new image to draw view.

src/js/BookReader.js
```diff
@@ -889,12 +889,14 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
 
         const img = document.createElement("img");
         const thumbReduce = floor(book.getPageWidth(leaf) / this.thumbWidth);
-
+        // use prefetched img src first if previously requested & available img is good enough
+        const prefetchedImg = this.prefetchedImgs[leaf] || {};
+        const imageURI = prefetchedImg.reduce <= thumbReduce ? prefetchedImg.uri : this._getPageURI(leaf, thumbReduce);
```
iisa, line 894: Thumbnail: added check of prefetchedImgs if img has been fetched && if its reduction factor is better than the thumbnail one. If so, then we reuse to save a request.

src/js/BookReader.js
```diff
@@ -1682,42 +1684,80 @@ BookReader.prototype._scrollAmount = function() {
...
 
-BookReader.prototype.prefetchImg = function(index) {
-  var pageURI = this._getPageURI(index, this.reduce);
+/**
+ * Used by 2up
+ * Fetches the image for requested index & saves in `this.prefetchedImgs`
+ * Does not re-request if image is in the
+ *
+ * @param {Number} index
+ * @param {Boolean} fetchNow
```
iisa, line 1693: new flag to allow for non-viewed pages to immediately be called. this is used to set up flipping action on 2up

src/js/BookReader.js
```diff
@@ -1682,42 +1684,80 @@ BookReader.prototype._scrollAmount = function() {
...
+    $imgEl.attr('src', pageURI);
+    if (pageURISrcset.length) {
+      $imgEl.attr('srcSet', pageURISrcset);
     }
-    pageContainer[0].uri = pageURI; // browser may rewrite src so we stash raw URI here
-    this.prefetchedImgs[index] = pageContainer[0];
+    $($imgEl).load(() => {
+      console.log('** PREFETCH DONE: index, this.reduce', index, this.reduce);
+    }).error(() => {
+      console.log('** PREFETCH ERROR: index, this.reduce', index, this.reduce);
```
iisa, line 1734: (saving room for error handling... - currently unsure what it is right now, so we can comment out if needed)

src/js/BookReader.js
```diff
@@ -1682,42 +1684,80 @@ BookReader.prototype._scrollAmount = function() {
...
+    this.prefetchedImgs[index] = $imgShell;
+  };
+
+  if (fetchNow || (index == this.twoPage.currentIndexL) || (index == this.twoPage.currentIndexR)) {
+    fetchImageAndRegister();
+  } else {
+    // stagger request
+    const time = 500;
+    setTimeout(() => {
+      fetchImageAndRegister();
```
iisa, line 1745: fire adjacent prefetches a lil after to reduce batch request load

src/js/BookReader.js
```diff
@@ -1682,42 +1684,80 @@ BookReader.prototype._scrollAmount = function() {
...
 BookReader.prototype.pruneUnusedImgs = function() {
+  const prefetchIsCapped = Object.keys(this.prefetchedImgs).length == 50;
   for (var key in this.prefetchedImgs) {
     if ((key != this.twoPage.currentIndexL) && (key != this.twoPage.currentIndexR)) {
       $(this.prefetchedImgs[key]).remove();
     }
     if ((key < this.twoPage.currentIndexL - 4) || (key > this.twoPage.currentIndexR + 4)) {
-      delete this.prefetchedImgs[key];
+      if (prefetchIsCapped || (this.prefetchedImgs[key]?.reduce > this.reduce)) {
+        delete this.prefetchedImgs[key];
```
iisa, line 1759: This is the reason why so many images that have already been fetched at a given scale must be re-fetched and are never cache-able => bc their respective DOM nodes were yanked.

this keeps a bigger count of images & their doms to use.  Capping at 50.

src/js/BookReader/Mode2Up.js
```diff
@@ -99,17 +99,21 @@ export class Mode2Up {
...
-    const { prefetchedImgs, twoPage } = this.br;
+    const { prefetchedImgs, displayedIndices } = this.br;
     const { reduce: idealReductionFactor } = this.getIdealSpreadSize( this.br.twoPage.currentIndexL, this.br.twoPage.currentIndexR );
 
     const currentImagesAreLarger = this.br.reduce <= idealReductionFactor;
-    const pagesInViewArePrefetched = prefetchedImgs[twoPage.currentIndexL] && prefetchedImgs[twoPage.currentIndexR];
+    const leftDisplayed = prefetchedImgs[displayedIndices[0]] || {};
+    const rightDisplayed = prefetchedImgs[displayedIndices[1]] || {};
+    const LeftImgIsPrefetchedToScale = leftDisplayed && (leftDisplayed.reduce <= this.br.reduce);
+    const RightImgIsPrefetchedToScale = rightDisplayed && (rightDisplayed.reduce <= this.br.reduce);
```
iisa, line 114: updating this to check currently displayed pages - if images are of decent size given a view, then let's keep them

src/js/BookReader/Mode2Up.js
```diff
@@ -175,10 +182,18 @@ export class Mode2Up {
...
+    const sameReducer = startingReduce == this.br.reduce;
+    const sameStart = startingIndices == this.br.displayedIndices;
+    const hasNewDisplayPagesOrDimensions = !sameStart || (sameStart && !sameReducer);
 
-    this.br.pruneUnusedImgs();
-    this.br.prefetch(); // Preload images or reload if scaling has changed
+    if (drawNewSpread || hasNewDisplayPagesOrDimensions) {
+      console.log("2UP PREPARE before 1st Prune & Prefetch (!sameStart || (sameStart && !sameReducer)) ", (!sameStart || (sameStart && !sameReducer)));
+      this.br.pruneUnusedImgs();
+      this.br.prefetch(); // Preload images or reload if scaling has changed
```
iisa, line 195: Now we only prune & pre-fetch if it is needed.

src/js/BookReader/Mode2Up.js
```diff
@@ -226,8 +241,6 @@ export class Mode2Up {
     this.drawLeafs();
     this.br.updateToolbarZoom(this.br.reduce);
 
-    this.br.prefetch();
```
iisa, line 229: no longer needed

src/js/BookReader.js
```diff
@@ -1682,42 +1684,80 @@ BookReader.prototype._scrollAmount = function() {
...
-    this.prefetchedImgs[index] = pageContainer[0];
+    $($imgEl).load(() => {
+      console.log('** PREFETCH DONE: index, this.reduce', index, this.reduce);
+    }).error(() => {
+      console.log('** PREFETCH ERROR: index, this.reduce', index, this.reduce);
+    })
+    this.prefetchedImgs[index] = $imgShell;
+  };
+
+  if (fetchNow || (index == this.twoPage.currentIndexL) || (index == this.twoPage.currentIndexR)) {
```
nsharma123, line 1739: Let's just consistent with identical operator (===)

iisa: hi Neeraj, bookreader prefers `==` to `===` - this is the only place that `==` is allowed ^_^

src/js/BookReader.js
bfalling: BLOCKER? If `prefetchedImg` is `{}`, then `prefetchedImg.reduce` in the next line will be accessing a non-existent property, no?

src/js/BookReader.js
```diff
@@ -1682,42 +1684,80 @@ BookReader.prototype._scrollAmount = function() {
...
+    const time = 500;
+    setTimeout(() => {
+      fetchImageAndRegister();
+    }, time);
   }
 };
 
+/** used in 2up */
 BookReader.prototype.pruneUnusedImgs = function() {
+  const prefetchIsCapped = Object.keys(this.prefetchedImgs).length == 50;
```
nsharma123, line 1752: Any reason we want to store 50 perfetched images? Why we can't store more or less? what are problems if we want to store more prefetch images, it will more bookreader better for large items.

src/js/BookReader.js
```diff
@@ -1682,42 +1684,80 @@ BookReader.prototype._scrollAmount = function() {
...
+ *     - this allows for "2up to prepare a page flip"
+ */
+BookReader.prototype.prefetchImg = async function(index, fetchNow = false) {
+  const pageURI = this._getPageURI(index, this.reduce);
   const pageURISrcset = this.options.useSrcSet ? this._getPageURISrcset(index, this.reduce) : [];
 
   // Load image if not loaded or URI has changed (e.g. due to scaling)
-  var loadImage = false;
+  let loadImage = false;
+  const wasPrefetchedSmaller = this.prefetchedImgs[index]?.reduce > this.reduce;
```
bfalling, line 1703: QUESTION: Will subsequent logic path be correct if `this.prefetchedImgs[index]?.reduce` and thus `wasPrefetchedSmaller` is null?

src/js/BookReader/Mode2Up.js
```diff
@@ -99,17 +99,21 @@ export class Mode2Up {
...
 
     const currentImagesAreLarger = this.br.reduce <= idealReductionFactor;
-    const pagesInViewArePrefetched = prefetchedImgs[twoPage.currentIndexL] && prefetchedImgs[twoPage.currentIndexR];
+    const leftDisplayed = prefetchedImgs[displayedIndices[0]] || {};
+    const rightDisplayed = prefetchedImgs[displayedIndices[1]] || {};
+    const LeftImgIsPrefetchedToScale = leftDisplayed && (leftDisplayed.reduce <= idealReductionFactor);
+    const RightImgIsPrefetchedToScale = rightDisplayed && (rightDisplayed.reduce <= idealReductionFactor);
 
-    return !currentImagesAreLarger || !pagesInViewArePrefetched;
+    return !currentImagesAreLarger || !(LeftImgIsPrefetchedToScale && RightImgIsPrefetchedToScale);
```
bfalling, line 116: QUESTION: Why capitalized variables?

src/js/BookReader.js
mmcnl: @bfalling If  `prefetchedImg` is `{}` and `prefetchedImg.reduce` is undefined, the comparison evals to false (since all greater/less than comparisons with `undefined` are false) and we end up in the second branch of the ternary.

src/js/BookReader/Mode2Up.js
iisa: pardon, idk why i did PascalCase... updating.

src/js/BookReader.js
iisa: yes!
image

src/js/BookReader.js
iisa: 50 is an arbitrary number as a starting point.
During this work, without this cap, a very large book can save a ton of images.
I don't want to keep an "open" gate until we:
1) plan for cached image sharing across modes
2) test against various ceilings.

thank you for calling out.

PR #646, opened 2021-03-06
Mode1Up DPI awareness

cdrini:
Add support for books/things scanned at multiple PPIs.

TODO:
- [x] Figure out the initial zoom :/ Right now it's showing up as super small. But things are the right size relative to each other :+1:
- [x] Fix/add tests

Review App: https://ia-petabox-drini.archive.org/details/lp_broken-hearted-ill-wander_dolores-keane-john-faulkner

Known issues:
- Things still start out to small on IA, but that was a pre-existing issue, so not dealing with it as part of this, cause this change is large enough as is :)

| Before | After |
| --- | --- |
| image https://lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=lp_broken-hearted-ill-wander_dolores-keane-john-faulkner#page/n1/mode/1up | image https://deploy-preview-646--nostalgic-austin-a91c9d.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=lp_broken-hearted-ill-wander_dolores-keane-john-faulkner#page/n1/mode/1up |

Other tests:
- https://deploy-preview-646--nostalgic-austin-a91c9d.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=lp_quatre-mephisto-valses_franz-liszt-strak#page/n2/mode/1up

### DIFF
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -1168,9 +1168,11 @@ BookReader.prototype.nextReduce = function(currentReduce, direction, reductionFa
   return reductionFactors[0];
 };
 
-BookReader.prototype._reduceSort = function(a, b) {
-  return a.reduce - b.reduce;
-};
+/**
+ * @param {ReductionFactor} a
+ * @param {ReductionFactor} b
+ */
+BookReader.prototype._reduceSort = (a, b) => a.reduce - b.reduce;
 
 /**
  * Attempts to jump to page
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -1208,8 +1208,6 @@ BookReader.prototype._isIndexDisplayed = function(index) {
  * @param {boolean} [noAnimate]
  */
 BookReader.prototype.jumpToIndex = function(index, pageX, pageY, noAnimate) {
-  const prevCurrentIndex = this.currentIndex();
-
   // Don't jump into specific unviewable page
   const page = this._models.book.getPage(index);
   if (!page.isViewable && page.unviewablesStart != page.index) {
@@ -1222,13 +1220,7 @@ BookReader.prototype.jumpToIndex = function(index, pageX, pageY, noAnimate) {
   this.trigger(BookReader.eventNames.stop);
 
   if (this.constMode2up == this.mode) {
-    // By checking against min/max we do nothing if requested index
-    // is current
-    if (index < Math.min(this.twoPage.currentIndexL, this.twoPage.currentIndexR)) {
-      this.flipBackToIndex(index);
-    } else if (index > Math.max(this.twoPage.currentIndexL, this.twoPage.currentIndexR)) {
-      this.flipFwdToIndex(index);
-    }
+    this._modes.mode2Up.jumpToIndex(index);
   } else if (this.constModeThumb == this.mode) {
     const { floor } = Math;
     const { book } = this._models;
@@ -1265,39 +1257,7 @@ BookReader.prototype.jumpToIndex = function(index, pageX, pageY, noAnimate) {
         .animate({ scrollTop: leafTop }, 'fast', () => { this.animating = false });
     }
   } else { // 1up
-    const { abs, floor } = Math;
-    let offset = 0;
-    let leafTop = this.onePageGetPageTop(index);
-    let leafLeft = 0;
-
-    if (pageY) {
-      const clientHeight = this.refs.$brContainer.prop('clientHeight');
-      offset = floor(pageY / this.reduce) - floor(clientHeight / 2);
-      leafTop += offset;
-    } else {
-      // Show page just a little below the top
-      leafTop -= this.padding / 2;
-    }
-
-    if (pageX) {
-      const clientWidth = this.refs.$brContainer.prop('clientWidth');
-      offset = floor(pageX / this.reduce) - floor(clientWidth / 2);
-      leafLeft += offset;
-    } else {
-      // Preserve left position
-      leafLeft = this.refs.$brContainer.scrollLeft();
-    }
-
-    // Only animate for small distances
-    if (!noAnimate && abs(prevCurrentIndex - index) <= 4) {
-      this.animating = true;
-      this.refs.$brContainer.stop(true).animate({
-        scrollTop: leafTop,
-        scrollLeft: leafLeft,
-      }, 'fast', () => { this.animating = false });
-    } else {
-      this.refs.$brContainer.stop(true).prop('scrollTop', leafTop);
-    }
+    this._modes.mode1Up.jumpToIndex(index, pageX, pageY, noAnimate);
   }
 };
 
--- a/src/js/BookReader/BookModel.js
+++ b/src/js/BookReader/BookModel.js
@@ -23,17 +23,20 @@ export class BookModel {
 
     /** @type {{width: number, height: number}} memoize storage */
     this._medianPageSize = null;
+    /** @deprecated @type {{width: number, height: number}} memoize storage */
+    this._medianPageSizePixels = null;
     /** @type {[PageData[], number]} */
     this._getDataFlattenedCached = null;
   }
 
   /**
+   * @deprecated Use getMedianPageSizeInches
    * Memoized
    * @return {{width: number, height: number}}
    */
   getMedianPageSize() {
-    if (this._medianPageSize) {
-      return this._medianPageSize;
+    if (this._medianPageSizePixels) {
+      return this._medianPageSizePixels;
     }
 
     // A little expensive but we just do it once
@@ -42,6 +45,27 @@ export class BookModel {
     for (let i = 0; i < this.getNumLeafs(); i++) {
       widths.push(this.getPageWidth(i));
       heights.push(this.getPageHeight(i));
+    }
+
+    widths.sort();
+    heights.sort();
+    this._medianPageSizePixels = {
+      width: widths[Math.floor(widths.length / 2)],
+      height: heights[Math.floor(heights.length / 2)]
+    };
+    return this._medianPageSizePixels;
+  }
+
+  getMedianPageSizeInches() {
+    if (this._medianPageSize) {
+      return this._medianPageSize;
+    }
+
+    const widths = [];
+    const heights = [];
+    for (const page of this.pagesIterator()) {
+      widths.push(page.widthInches);
+      heights.push(page.heightInches);
     }
 
     widths.sort();
@@ -346,10 +370,14 @@ class PageModel {
    * @param {PageIndex} index
    */
   constructor(book, index) {
+    // TODO: Get default from config
+    this.ppi = book._getDataProp(index, 'ppi', 500);
     this.book = book;
     this.index = index;
     this.width = book.getPageWidth(index);
+    this.widthInches = this.width / this.ppi;
     this.height = book.getPageHeight(index);
+    this.heightInches = this.height / this.ppi;
     this.pageSide = book.getPageSide(index);
     this.leafNum = book._getDataProp(index, 'leafNum', this.index);
 
--- a/src/js/BookReader/BookModel.js
+++ b/src/js/BookReader/BookModel.js
@@ -58,6 +58,7 @@ export class BookModel {
     return this._medianPageSizePixels;
   }
 
+  /** Get median width/height of page in inches. Memoized for performance. */
   getMedianPageSizeInches() {
     if (this._medianPageSize) {
       return this._medianPageSize;
--- a/src/js/BookReader/BookModel.js
+++ b/src/js/BookReader/BookModel.js
@@ -9,6 +9,11 @@ import { clamp } from './utils.js';
 const UNVIEWABLE_PAGE_URI = '/bookreader/static/preview-default.png';
 
 /**
+ * Fallback PPI when one isn't specified in the book data
+ */
+export const FALLBACK_PPI = 500;
+
+/**
  * Contains information about the Book/Document independent of the way it is
  * being rendering. Nothing here should reference e.g. the mode, zoom, etc.
  * It's just information about the book and its pages (usually as specified
@@ -371,7 +376,7 @@ class PageModel {
    */
   constructor(book, index) {
     // TODO: Get default from config
-    this.ppi = book._getDataProp(index, 'ppi', 500);
+    this.ppi = book._getDataProp(index, 'ppi', FALLBACK_PPI);
     this.book = book;
     this.index = index;
     this.width = book.getPageWidth(index);
--- a/src/js/BookReader/BookModel.js
+++ b/src/js/BookReader/BookModel.js
@@ -1,4 +1,5 @@
 // @ts-check
+import { DEFAULT_OPTIONS } from './options.js';
 import { clamp } from './utils.js';
 /** @typedef {import('./options.js').PageData} PageData */
 /** @typedef {import('../BookReader.js').default} BookReader */
@@ -7,11 +8,6 @@ import { clamp } from './utils.js';
 // TODO Render configurable html for the user instead.
 // FIXME Don't reference files on archive.org
 const UNVIEWABLE_PAGE_URI = '/bookreader/static/preview-default.png';
-
-/**
- * Fallback PPI when one isn't specified in the book data
- */
-export const FALLBACK_PPI = 500;
 
 /**
  * Contains information about the Book/Document independent of the way it is
@@ -25,6 +21,7 @@ export class BookModel {
    */
   constructor(br) {
     this.br = br;
+    this.ppi = br.options?.ppi ?? DEFAULT_OPTIONS.ppi;
 
     /** @type {{width: number, height: number}} memoize storage */
     this._medianPageSize = null;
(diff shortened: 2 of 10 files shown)


src/js/BookReader.js
```diff
@@ -1216,13 +1214,7 @@ BookReader.prototype.jumpToIndex = function(index, pageX, pageY, noAnimate) {
...
 
   if (this.constMode2up == this.mode) {
-    // By checking against min/max we do nothing if requested index
-    // is current
-    if (index < Math.min(this.twoPage.currentIndexL, this.twoPage.currentIndexR)) {
-      this.flipBackToIndex(index);
-    } else if (index > Math.max(this.twoPage.currentIndexL, this.twoPage.currentIndexR)) {
-      this.flipFwdToIndex(index);
-    }
+    this._modes.mode2Up.jumpToIndex(index);
```
iisa, line 1217: thanks!

src/js/BookReader/Mode1Up.js
```diff
@@ -12,144 +12,219 @@ export class Mode1Up {
...
+    this.$documentContainer = $('<div class="BRpageview" />');
+    /** @private */
+    this.screenDPI = calcScreenDPI();
+    /** @private */
+    this.LEAF_SPACING_IN = 0.2;
+
+    /**
+     * How much smaller the picture on screen is than the real-world item
+     *
+     * Mode1Up doesn't use the br.reduce because it is DPI aware. The reduction factor
```
iisa, line 26: ✔️

src/js/BookReader/Mode1Up.js
```diff
@@ -12,144 +12,219 @@ export class Mode1Up {
   constructor(br, bookModel) {
     this.br = br;
     this.book = bookModel;
+
+    /** @private */
+    this.$documentContainer = $('<div class="BRpageview" />');
```
iisa, line 17: Q: does this get bound to `br.refs`? do we need to?

src/js/BookReader/Mode1Up.js
```diff
@@ -12,144 +12,219 @@ export class Mode1Up {
...
 
-  drawLeafs() {
-    const { book } = this;
-    const containerHeight = this.br.refs.$brContainer.height();
-    const containerWidth = this.br.refs.$brPageViewEl.width();
-    const scrollTop = this.br.refs.$brContainer.prop('scrollTop');
-    const scrollBottom = scrollTop + containerHeight;
+  /**
+   * Get the number of pixels required to display the given inches with the given reduce
+   * @param {number} inches
```
iisa, line 60: suggestion: update jsdoc to reflect args & return val

src/js/BookReader/Mode1Up.js
```diff
@@ -12,144 +12,219 @@ export class Mode1Up {
...
+   **/
+  physicalInchesToDisplayPixels(inches, reduce = this.realWorldReduce, screenDPI = this.screenDPI) {
```
iisa, line 62: q: does `inches` need a default value?

src/js/BookReader/Mode1Up.js
```diff
@@ -12,144 +12,219 @@ export class Mode1Up {
...
+    return inches * screenDPI / reduce;
+  }
 
-    const indicesToDisplay = [];
+  /** Iterate over pages, augmented with their top/bottom bounds */
```
iisa, line 66: suggestion: update docs to reflect args

src/js/BookReader/Mode1Up.js
```diff
@@ -12,144 +12,219 @@ export class Mode1Up {
...
-        const img = $('<img />', {
-          src: page.getURI(this.br.reduce, 0),
-          srcset: pageURISrcset,
-          alt: 'Book page image'
-        });
-        pageContainer.append(img);
-
-        BRpageViewEl.appendChild(pageContainer[0]);
+  drawLeafs() {
+    const pagesToDisplay = Array.from(this.findIntersectingPages());
```
iisa, line 117: q: why do we need to create a new array, does `this.findIntersectingPages()` not return an array?

src/js/BookReader/Mode1Up.js
```diff
@@ -12,144 +12,219 @@ export class Mode1Up {
...
     }
+  }
 
-    const BRpageViewEl = this.br.refs.$brPageViewEl.get(0);
-    leafTop = 0;
+  /**
+   * @param {{ top: number; bottom: number; }} bound1
+   * @param {{ top: number; bottom: number; }} bound2
+   */
+  static boundsIntersect(bound1, bound2) {
```
iisa, line 84: suggestion: update jsdoc to reflect return value

src/js/BookReader/Mode1Up.js
```diff
@@ -12,144 +12,219 @@ export class Mode1Up {
...
+    return (bound1.bottom >= bound2.top) && (bound1.top <= bound2.bottom);
+  }
 
-    for (const page of book.pagesIterator({ end: firstIndexToDraw, combineConsecutiveUnviewables: true })) {
-      leafTop += Math.floor(page.height / this.br.reduce) + 10;
+  /**
+   * Find the pages that intersect the current viewport, including 1 before/after
+   **/
+  * findIntersectingPages() {
```
iisa, line 91: q: is this easy enough  to test?

src/js/BookReader/Mode1Up.js
```diff
@@ -12,144 +12,219 @@ export class Mode1Up {
...
+          const width = this.physicalInchesToDisplayPixels(page.widthInches);
+          const reduce = page.width / width;
+
+          this.br._createPageContainer(page.index, {
+            width,
+            height,
+            top,
+            left: Math.floor((documentContainerWidth - width) / 2),
+          })
+            .append($('<img />', {
```
iisa, line 138: q: should we just use the image cache for this, too?

src/js/BookReader/Mode1Up.js
```diff
@@ -12,144 +12,219 @@ export class Mode1Up {
...
       return;
     }
 
-    this.br.reduce = reduceFactor.reduce; // $$$ incorporate into function
-    this.br.onePage.autofit = reduceFactor.autofit;
+    this.realWorldReduce = nextReductionFactor.reduce;
+    this.br.onePage.autofit = nextReductionFactor.autofit;
 
-    this.br.pageScale = this.br.reduce; // preserve current reduce
+    // this.br.pageScale = this.reduce; // preserve current reduce
```
iisa, line 224: suggestion: yank if not needed

src/js/BookReader/Mode1Up.js
```diff
@@ -178,13 +260,9 @@ export class Mode1Up {
    * @return {number}
    */
   getPageTop(index) {
-    const { floor } = Math;
-    const { book } = this;
-    let leafTop = 0;
-    for (const page of book.pagesIterator({ end: index, combineConsecutiveUnviewables: true })) {
-      leafTop += floor(page.height / this.br.reduce) + this.br.padding;
+    for (const {page, top} of this.pagesWithBounds()) {
+      if (page.index == index) return top;
```
iisa, line 264: q: does we expect this to always find a match & return a value?

src/js/BookReader/Mode1Up.js
```diff
@@ -205,10 +283,10 @@ export class Mode1Up {
    * Note this calls drawLeafs
    */
   resizePageView() {
-    const viewWidth  = this.br.refs.$brContainer.prop('clientWidth');
-    const oldScrollTop  = this.br.refs.$brContainer.prop('scrollTop');
-    const oldPageViewHeight = this.br.refs.$brPageViewEl.height();
-    const oldPageViewWidth = this.br.refs.$brPageViewEl.width();
+    const viewWidth  = this.$scrollContainer.prop('clientWidth');
```
iisa, line 286: Q: are we/should we be moving away from stashing Dom in `this.br.refs`?

src/js/BookReader/Mode1Up.js
```diff
@@ -226,38 +304,39 @@ export class Mode1Up {
       // current index in drawLeafsOnePage after we create the new view container
 
       // Make sure this will count as current page after resize
-      const fudgeFactor = (this.book.getPageHeight(this.br.currentIndex()) / this.br.reduce) * 0.6;
+      const fudgeFactor = 0.6 * this.physicalInchesToDisplayPixels(this.book.getPage(this.br.currentIndex()).heightInches);
       const oldLeafTop = this.getPageTop(this.br.currentIndex()) + fudgeFactor;
-      const oldViewDimensions = this.calculateViewDimensions(this.br.reduce, this.br.padding);
+      const oldViewDimensions = this.calculateViewDimensions();
```
iisa, line 309: so happy to see the step to move away from `br.reduce` as a main dimension to calculate book/image size against.

src/js/BookReader/Mode1Up.js
```diff
@@ -226,38 +304,39 @@ export class Mode1Up {
...
+      this.realWorldReduce = reductionFactor.reduce;
     }
 
-    const viewDimensions = this.calculateViewDimensions(this.br.reduce, this.br.padding);
-
-    this.br.refs.$brPageViewEl.height(viewDimensions.height);
-    this.br.refs.$brPageViewEl.width(viewDimensions.width);
+    const viewDimensions = this.calculateViewDimensions();
+    this.$documentContainer.height(viewDimensions.height);
+    this.$documentContainer.width(viewDimensions.width);
```
iisa, line 323: the old property `$brPageViewEl` is also used in thumbnails, have u found any other shared properties across modes, like css or anything else? (does it matter?)

src/js/BookReader/Mode1Up.js
```diff
@@ -226,38 +304,39 @@ export class Mode1Up {
...
 
+    // Remove all pages
+    this.$documentContainer.empty();
+    this.br.displayedIndices = [];
```
iisa, line 327: this signals to me a fresh draw on resize, how expensive is this when we view BR in context of book navigator & the menu slides open/close? (does it matter?)

src/js/BookReader/Mode1Up.js
```diff
@@ -226,38 +304,39 @@ export class Mode1Up {
...
-    const newTop = Math.max(0, Math.floor( newCenterY - this.br.refs.$brContainer.height() / 2 ));
-    this.br.refs.$brContainer.prop('scrollTop', newTop);
-
-    // We use clientWidth here to avoid miscalculating due to scroll bar
+    // Scroll to the right spot
     const newCenterX = oldCenterX * (viewWidth / oldPageViewWidth);
-    let newLeft = newCenterX - this.br.refs.$brContainer.prop('clientWidth') / 2;
-    newLeft = Math.max(newLeft, 0);
-    this.br.refs.$brContainer.prop('scrollLeft', newLeft);
+    const newCenterY = scrollRatio * viewDimensions.height;
```
iisa, line 331: I tend to prefer object destructure, but I have seen the reference not reflect the "fresh scope" some of these function instances hold inside the core code...I am wondering if you have seen that as well?
if not, what should be our baseline style if scope isn't an option - destructure or traversal like `viewDimensions.height`?

src/js/BookReader/utils.js
```diff
@@ -159,3 +159,19 @@ export function PolyfilledCustomEvent(eventName, {bubbles = false, cancelable =
   event.initCustomEvent(eventName, bubbles, cancelable, detail);
   return event;
 }
+
+/**
+ * Returns the number pixels something should be rendered at to be ~1n on the users
+ * screen when measured with a ruler.
+ */
+export function calcScreenDPI() {
```
iisa, line 167: Q: should we unit test this also?

iisa (changes requested):
looks good. 🙌  I did have a few questions which sort of segue to how we should be deprecating properties, patterns like pinning DOM to this.refs, etc.

+1  for isolating reduction factor

src/js/BookReader/BookModel.js
```diff
@@ -8,6 +8,11 @@ import { clamp } from './utils.js';
 // FIXME Don't reference files on archive.org
 const UNVIEWABLE_PAGE_URI = '/bookreader/static/preview-default.png';
 
+/**
+ * Fallback PPI when one isn't specified in the book data
+ */
+export const FALLBACK_PPI = 500;
```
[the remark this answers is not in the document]
cdrini: I found books with both? I added a new option, options.ppi, as the fallback (adding to petabox to specify it when BR is setup)

src/js/BookReader/Mode2Up.js
```diff
@@ -22,6 +22,19 @@ export class Mode2Up {
     this.leafEdgeR = null;
   }
 
+  /**
+   * @param {PageIndex} index
```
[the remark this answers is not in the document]
cdrini: I did some work a while back; basically PageIndex is usually what's inside BR; it's 0-based, and it's litterly the `i`th displayed page. LeafIndex is what's in scandata. It's documented somewhere here, but I forget 😅 And I think I got something wrong when I wrote those down 😅 (Here: https://github.com/internetarchive/bookreader/blob/39323cc8bbf06f9dd005057d922fb84ff457a8e7/src/js/BookReader/BookModel.js#L447-L462 )

src/js/BookReader/Mode1Up.js
cdrini: Nope! Only used by Mode1Up, so no need to expose it globally :)

src/js/BookReader/Mode1Up.js
cdrini: Not here; that's page specific, so is always going to be variable. reduce/screenDPI _should_ be the same for everything, but occasionally a special value is specified.

src/js/BookReader/Mode1Up.js
cdrini: It's a generator ! (#es6 #future). Array.from let's us convert a generator to an array.

src/js/BookReader/Mode1Up.js
cdrini: 👍 This is a good one to add a test for...

src/js/BookReader/Mode1Up.js
cdrini: Absolutely! In a future PR 😅

src/js/BookReader/Mode1Up.js
cdrini: Yep! Should always be given a valid index.

src/js/BookReader/Mode1Up.js
cdrini: Anything that's not necessary to be globally accessible, should be class-specific. $brPageViewEl is Mode1Up specific, so hiding it here from the rest of the world.

src/js/BookReader/Mode1Up.js
cdrini: Bye bye bye!

src/js/BookReader/Mode1Up.js
cdrini: I think that's a (odd and confusing :P ) coincidence; haven't noticed any issues; we can probably be less careful about calling `.empty()` now maybe?

src/js/BookReader/Mode1Up.js
cdrini: Hmm, not sure I follow the scope thing. Can you give an example?

For readability here, I think `viewDimensions.height` is more readable, cause if I destructure to just `height`, there are like 6 different heights at play in this block of code :D

src/js/BookReader/utils.js
cdrini: I actually couldn't think of a way to unit test this one. Setting `width:1in` in jest seems to return an object with `width: 0px` 😬 .

src/js/BookReader/Mode1Up.js
cdrini: Added description, but specifically not adding the types here. The inference is smart enough to figure these types out, so specifying them just means they're more likely to fall out of sync.

src/js/BookReader/Mode1Up.js
cdrini: Add descriptions; ditto to above.

src/js/BookReader/Mode1Up.js
cdrini: Fixed

src/js/BookReader/Mode2Up.js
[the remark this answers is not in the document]
iisa: thank you, i will keep this in mind and use them to bring clarity as well

src/js/BookReader/Mode1Up.js
cdrini: Punting on it for now; it would be a good one to test, but I'm not sure it would be super easy 😅

src/js/BookReader/utils.js
iisa: oh yes, i totally have encountered weird "stub dom" issues ^_^

src/js/BookReader/Mode1Up.js
iisa: i am here for it, i think the image caching work has allowed for a "fresher" dom redraw at the moment since we don't have to account for an expensive request

src/js/BookReader/Mode1Up.js
iisa: > Hmm, not sure I follow the scope thing. Can you give an example?
this was an overall observation, esp. working with lit element, i've found that `const { fooFn } = this` can break fooFn scope. so anytime something gets/doesn't get destructured, my feelers go up.

readability-wise, the object traversal is for sure more helpful ^_^

iisa (approved):
🚢

PR #651, opened 2021-03-11
Thumbnail draw with prefetched images ++

iisa:
### Thumbnail draw fixes
- uses image cache dictionary to use best fetched image url
  - value can be seen during the following actions:
    - switching modes from 2up -> thumbnail/gallery
    - browser width resize while in thumbnail/gallery mode
    - scrolling quickly up & down through thumbnail/gallery mode
- reduces "row buffer" on draw
- image reducer is factor of 2, to reduce calls with odd number factors that aren't respected by image service
thumbnail image reuse)

### Image Cache dictionary updates
- no longer saving image element
- serves image element on the fly
- `imageLoaded` checks against given index & reducer
- unit tests

### Util.sleep
- new promise based sleep function
- unit test

### Bonus:  2up mode resizes book on flip
- this accounts for books of irregular widths, ensuring that:
  - a spread can be scrolled horizontally if needed
  - a spread that doesn't need to be scrolled doesn't have a scrollable container
  
2up mode horizontal scroll

### DIFF
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -981,9 +981,7 @@ BookReader.prototype.lazyLoadImage = function ($imgPlaceholder) {
   const $img = this.imageCache.image(leaf, reduce);
 
   // replace with the new img
-  const $parentContainer = $($imgPlaceholder).parent();
-  $parentContainer.append($img);
-  $($imgPlaceholder).remove();
+  $(imgPlaceholder).replaceWith($img);
 };
 
 /**
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -856,7 +856,7 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
   for (let i = 1; i < this.thumbRowBuffer + 1; i++) {
     if (lastRow + i < leafMap.length) { rowsToDisplay.push(lastRow + i); }
   }
-  for (let i = 1; i < this.thumbRowBuffer + 1; i++) {
+  for (let i = 1; i < this.thumbRowBuffer; i++) {
     if (firstRow - i >= 0) { rowsToDisplay.push(firstRow - i); }
   }
   rowsToDisplay.sort();
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -982,14 +982,14 @@ BookReader.prototype.lazyLoadThumbnails = function() {
  *
  * @param {$lazyLoadImgPlaceholder} - $imgPlaceholder
  */
-BookReader.prototype.lazyLoadImage = function ($imgPlaceholder) {
-  const leaf =  $($imgPlaceholder).data('leaf');
-  const reduce =  $($imgPlaceholder).data('reduce');
+BookReader.prototype.lazyLoadImage = function (imgPlaceholder) {
+  const leaf =  $(imgPlaceholder).data('leaf');
+  const reduce =  $(imgPlaceholder).data('reduce');
   const $img = this.imageCache.image(leaf, reduce);
-  const $parent = $($imgPlaceholder).parent();
+  const $parent = $(imgPlaceholder).parent();
   /* March 16, 2021 (isa) - manually append & remove, `replaceWith` currently loses closure scope */
   $($parent).append($img);
-  $($imgPlaceholder).remove();
+  $(imgPlaceholder).remove();
 };
 
 /**
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -959,14 +959,9 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
 };
 
 BookReader.prototype.lazyLoadThumbnails = function() {
-  const self = this;
-
   const batchImageRequestByRow = ($images) => {
     utils.wait().then(() => {
-      $images.each(function loadEachImg() {
-        const $imgPlaceholder = this;
-        self.lazyLoadImage($imgPlaceholder)
-      });
+      $images.each((index, $imgPlaceholder) => this.lazyLoadImage($imgPlaceholder));
     })
   };
 
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -711,8 +711,7 @@ BookReader.prototype._createPageContainer = function(index, styles) {
   const container = $('<div />', {
     'class': `BRpagecontainer BRmode${modeClasses[this.mode]} pagediv${index}`,
     css,
-  }).attr('data-side', pageSide).append($('<div />', { 'class': 'BRscreen' }))
-    .append($(`<div />INDEX: ${index}</div>`, { 'style': 'width: 100%; background-color: red' }));
+  }).attr('data-side', pageSide).append($('<div />', { 'class': 'BRscreen' }));
   container.toggleClass('protected', this.protected);
 
   return container;
@@ -824,6 +823,7 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
   let leafTop = 0;
   let leafBottom = 0;
   const rowsToDisplay = [];
+  const imagesToDisplay = [];
 
   // Visible leafs with least/greatest index
   let leastVisible = book.getNumLeafs() - 1;
@@ -903,13 +903,13 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
         });
 
         this.refs.$brPageViewEl.append($pageContainer);
+        imagesToDisplay.push(leaf);
 
         const thumbReduce = floor(book.getPageWidth(leaf) / this.thumbWidth);
         // use prefetched img src first if previously requested & available img is good enough
         // dip into cache so we don't set off a request
 
         const baseCSS = { width: `${leafWidth}px`, height: `${leafHeight}px` };
-
         if (this.imageCache.imageLoaded(leaf)) {
           // send to page
           console.log("SEND TO PAGE")
@@ -926,8 +926,7 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
             .attr('data-reduce', thumbReduce)
             .attr('data-row', row)
             .attr('alt', 'Loading book image');
-          $pageContainer.append($lazyLoadImgPlaceholder)
-            .append(`<div src="width: 100%; background-color: red;">${leaf} -- ${thumbReduce} -- row: ${row}</div>`);
+          $pageContainer.append($lazyLoadImgPlaceholder);
         }
       }
     }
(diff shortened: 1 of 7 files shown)


src/js/BookReader.js
```diff
@@ -940,58 +959,37 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
...
-  var toLoad = this.thumbMaxLoading - loading;
+  const self = this;
 
-  var self = this;
+  const batchImageRequestByRow = ($images) => {
+    utils.wait().then(() => {
+      $images.each(function loadEachImg() {
+        const $imgPlaceholder = this;
+        self.lazyLoadImage($imgPlaceholder)
+      });
```
cdrini, line 969:
```suggestion
      $images.each((i, imgPlaceholder) => this.lazyLoadImage(imgPlaceholder));
```

src/js/BookReader.js
```diff
@@ -940,58 +959,37 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
...
 BookReader.prototype.lazyLoadThumbnails = function() {
-  // We check the complete property since load may not be fired if loading from the cache
-  this.$('.BRlazyloading').filter('[complete=true]').removeClass('BRlazyloading');
-
-  var loading = this.$('.BRlazyloading').length;
-  var toLoad = this.thumbMaxLoading - loading;
+  const self = this;
 
-  var self = this;
+  const batchImageRequestByRow = ($images) => {
```
cdrini, line 964: We can make this function async

```suggestion
  const batchImageRequestByRow = async ($images) => {
```

src/js/BookReader.js
```diff
@@ -940,58 +959,37 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
...
+    utils.wait().then(() => {
```
cdrini, line 965: Can you add a comment about why we're waiting here?

```suggestion
    await utils.wait(500);
```

src/js/BookReader/utils.js
```diff
@@ -159,3 +159,13 @@ export function PolyfilledCustomEvent(eventName, {bubbles = false, cancelable =
   event.initCustomEvent(eventName, bubbles, cancelable, detail);
   return event;
 }
+
+/**
+ * Promise based wait - resolves at default 500ms
+ * @param {Number} wait time in milliseconds
+ */
+export function wait(ms = 500) {
```
cdrini, line 167: This should probably not have a default; also this function is usually called `sleep`

```suggestion
export function sleep(ms) {
```

src/js/BookReader/utils.js
```diff
@@ -159,3 +159,13 @@ export function PolyfilledCustomEvent(eventName, {bubbles = false, cancelable =
...
+  return new Promise((res) => {
+    setTimeout(() => res(true), ms);
+  });
```
cdrini, line 170: SUGGESTION: I love promises :)

```suggestion
  return new Promise(res => setTimeout(res, ms));
```

src/js/BookReader.js
```diff
@@ -940,58 +959,37 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
...
-      'alt': 'Loading book image'
-    });
+/**
+ * Replaces placeholder image with real one
+ *
+ * @param {$lazyLoadImgPlaceholder} - $imgPlaceholder
+ */
+BookReader.prototype.lazyLoadImage = function ($imgPlaceholder) {
+  const leaf =  $($imgPlaceholder).data('leaf');
+  const reduce =  $($imgPlaceholder).data('reduce');
```
cdrini, line 986: No `$` since this doesn't look like a jquery object.

```suggestion
 * @param {HTMLImageElement} - imgPlaceholder
 */
BookReader.prototype.lazyLoadImage = function (imgPlaceholder) {
  const leaf =  $(imgPlaceholder).data('leaf');
  const reduce =  $(imgPlaceholder).data('reduce');
```

src/js/BookReader.js
```diff
@@ -940,58 +959,37 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
...
+  const $img = this.imageCache.image(leaf, reduce);
 
   // replace with the new img
-  $(dummyImage).before(img).remove();
-
-  img = null; // tidy up closure
+  const $parentContainer = $($imgPlaceholder).parent();
+  $parentContainer.append($img);
+  $($imgPlaceholder).remove();
```
cdrini, line 992: Can you do
```suggestion
  $(imgPlaceholder).replaceWith($img);
```

src/js/BookReader.js
```diff
@@ -1503,12 +1502,12 @@ BookReader.prototype.prepareThumbnailView = function() {
   // $$$ keep select enabled for now since disabling it breaks keyboard
   //     nav in FF 3.6 (https://bugs.edge.launchpad.net/bookreader/+bug/544666)
   // utils.disableSelect(this.$('#BRpageview'));
+  const defaultThumbnailWidth = 100;
```
cdrini, line 1505:
```suggestion
  const DEFAULT_THUMBNAIL_WIDTH = 100;
```

src/js/BookReader.js
```diff
@@ -1503,12 +1502,12 @@ BookReader.prototype.prepareThumbnailView = function() {
...
 
-  this.thumbWidth = this.getThumbnailWidth(this.thumbColumns);
+  const idealReduce = this.getThumbnailWidth(this.thumbColumns);
```
cdrini, line 1507: Actually, can we put `DEFAULT_THUMBNAIL_WIDTH` inside this function? So it returns the default if it gets a number <= 0?

src/js/BookReader/ImageCache.js
```diff
@@ -32,6 +32,37 @@ export class ImageCache {
...
+   * @returns {Number} reduce
+   */
+  imageReduce(index) {
+    return this.cache[index]?.reduce;
+  }
+
+  /**
+   * Checks if image's cache status
+   * @param {String|Number} index - page index
+   * @returns {{ index: Number, inCache: (Object|undefined), reduce: Number, loaded: Boolean }}
```
cdrini, line 56: You can remove this; it'll be able to deduce. (Less likely to fall out of sync with the code if we don't duplicate it).
```suggestion
```

cdrini (changes requested):
Looking good. I think we have two _potential_ issues I'll test:

1) Does the ImageCache have a maximum? It might result in a memory leak if a person is reading a long book. Not a blocker though
2) Does zooming in work? The `imageLoaded` check doesn't take the reduce into account, so wouldn't it always return "true", even if the zoom has changed?

I think we can add some quick unit tests here :) Namely for:

1) The sleep function; that should be easy to test!
2) The ImageCache functions.

src/js/BookReader.js
iisa: thanks! updated as such + removed unused `self` var

src/js/BookReader.js
iisa: yep

iisa:
hi @cdrini , thanks so much for the care in your CR.

resolving your outstanding questions:
> Does the ImageCache have a maximum? It might result in a memory leak if a person is reading a long book. Not a blocker though

image element is no longer stored, just a dictionary of the best scaled image that has been fetched. image elements are now served on the fly.

> Does zooming in work? The imageLoaded check doesn't take the reduce into account, so wouldn't it always return "true", even if the zoom has changed?

thanks for catching - i've updated check to account for a given reducer + unit tests on expected values

As @pezvi noted, the edge of the cache dictionary drifting from an image's cache expiration is still to watch out for. But, outside of implementing our own service worker, or calculating an entry refresh, this is a good place to start. 👍  + this is easily reset by a page refresh.

src/js/BookReader.js
```diff
@@ -940,58 +966,30 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
...
 BookReader.prototype.lazyLoadThumbnails = function() {
-  // We check the complete property since load may not be fired if loading from the cache
-  this.$('.BRlazyloading').filter('[complete=true]').removeClass('BRlazyloading');
-
-  var loading = this.$('.BRlazyloading').length;
-  var toLoad = this.thumbMaxLoading - loading;
-
-  var self = this;
+  const batchImageRequestByRow = async ($images) => {
+    await utils.sleep(300);
```
jbuckner, line 970: _Question_ What is the `sleep` for?

src/js/BookReader.js
```diff
@@ -891,29 +899,46 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
...
+            .attr('data-row', row)
+            .attr('alt', 'Loading book image');
+          $pageContainer.append($lazyLoadImgPlaceholder);
+        }
       }
     }
   }
+  console.log("IMGS TO DISPLAY", imagesToDisplay);
+  console.log('rowsInView', rowsInView);
+  console.log('this.displayedRows', this.displayedRows);
```
jbuckner, line 933: Some logging here

src/js/BookReader.js
```diff
@@ -768,6 +771,7 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
   let seekTop;
 
   // Calculate the position of every thumbnail.  $$$ cache instead of calculating on every draw
+  // make `leafMap`
```
jbuckner, line 774: Is this next loop making the `leafMap`?

src/js/BookReader.js
```diff
@@ -891,29 +899,46 @@ BookReader.prototype.drawLeafsThumbnail = function(seekIndex) {
...
+        const idealReduce = floor(book.getPageWidth(leaf) / this.thumbWidth);
+        const nearestFactor2 = 2 * Math.round(idealReduce / 2);
+        const thumbReduce = nearestFactor2;
+
+        const baseCSS = { width: `${leafWidth}px`, height: `${leafHeight}px` };
+        if (this.imageCache.imageLoaded(leaf, thumbReduce)) {
+          // send to page
+          $pageContainer.append($(this.imageCache.image(leaf, thumbReduce)).css(baseCSS));
+        } else {
+          // lazy load
```
jbuckner, line 915: _Question_ Does the image get added to the `imageCache` if it gets lazy-loaded?

jbuckner (approved):
A few questions/comments, but looks good overall 👍

src/js/BookReader.js
iisa: the sleep is to help give image requests some space between drawing rows instead of the current behavior - burst request of 20-40 images at a time that's putting strain on the service.

src/js/BookReader.js
iisa: thank you for catching. :pokeball:

PR #667, opened 2021-03-18
Update Node + update testing dependencies

cdrini:
No changes to generated files.


cdrini:
Auto-merging; testing-dependency only change.

PR #673, opened 2021-03-20
fix: prevent keyboard shortcut when target is `input`

### LINKED ISSUE #371: Unable to type "-" or "=" in input fields outside BR
To reproduce:
1. Go to https://archive.org/details/completebookofpi00hass
2. Try to type "-" or "=" into the search box at the top of the page

* Expected: Can type them
* Actual: They zoom in/out the BR

## Proposal & Constraints

- Keyboard shortcuts should only work if the event target is not an input box

## Stakeholders

@textfiles

CodeDotJS:
fixes #371

Tested it using @shaneriley's [method](https://github.com/internetarchive/bookreader/issues/371#issuecomment-664587467)

### DIFF
--- a/src/js/BookReader.js
+++ b/src/js/BookReader.js
@@ -1920,8 +1920,9 @@ BookReader.prototype.stopFlipAnimations = function() {
  * @param {Event}
  * @return {boolean}
  */
-BookReader.prototype.keyboardNavigationIsDisabled = function(event) {
-  return event.target.tagName == "INPUT";
+
+BookReader.prototype.keyboardNavigationIsDisabled = function() {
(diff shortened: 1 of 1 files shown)


CodeDotJS:
@cdrini can you take a look?

cdrini:
Hi @CodeDotJS , thanks for PR!

It looks like the issue on IA is because the search input is inside a web components. Run this in the console to reproduce:

```js
  constructor() {
    // Always call super first in constructor
    super();

    // Element functionality written in here

    this.shadowRoot.append(document.createElement('input'));
  }
});
$(document.body).append($(`<test-element/>`));
```

PR #685, opened 2021-03-25
Upgrade from dependabot v1 to v2

cclauss:


cdrini:
Closing; we no longer use dependabot

Issue #509, opened 2020-10-26
TTS Forward button doesn't work on Chrome+Touchscreen


cdrini:
The tooltips are getting in the way somehow. And clicking the tooltip, because they're contained in the button, cause the audio button to be clicked.

To reproduce:

1. Go to https://lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive.html?ocaid=adventuresofoli00dick#page/n14/mode/1up
2. Click the listen button
3. try to press the forward button

- Expected: It goes forward
- Actual: It stops playback

**Evidence / Screenshot (if possible)**

This is what it looks like on Chrome desktop with mobile mode:

![forward tts bug](https://user-images.githubusercontent.com/6251786/97234122-92d74400-17b6-11eb-97b6-f174d85fadd6.gif)

NOTE: In real mobile, the tooltip is invisible, but still clickable.

**Context**

Chrome 86 @ And6

**Stakeholders**

> @-tag any stakeholders of this bug

participator:
How do you get the tooltips to appear?  I don't see them when using the live example links.

See live examples:

On details page: https://archive.org/details/birdbookillustra00reedrich
Full Screen: https://archive.org/stream/birdbookillustra00reedrich (mobile-friendly)

himanshujaidka:
yes @cdrini, it is not seen in the live examples?

cdrini:
The tooltip might be invisible; but it still prevents clicking the forward button

cdrini:
Ohhhh, I see what you mean, they're fully missing on archive.org :( You should be able to see them here: https://lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive.html?ocaid=adventuresofoli00dick#page/n14/mode/1up

ArunTeltia:
we can fix this issue by instead of showing the message above the button we can show that below the button what you think 
Can I work on this

Adityakushwaha142:
I can work on this issue

cdrini:
Fixed by #1213

PR #689, opened 2021-03-27
fix forward button issue

Dnouv:
As mentioned in issue #509, the button title was hiding the forward button, I have tried to fix that; the forward button is working now. Here are the screenshots after the changes:
br_509

Feedbacks are most welcomed. Thank you!


Dnouv:
@cdrini, I understand you may be busy, but I would truly appreciate it if you could review this project.

SwapnilSoni1999 (approved):

PR #691, opened 2021-03-29
Changed the var declarations to modern ES6 const or let

Priyaraj17:
Updated the javascript files in which "VAR" declaration was used to modern ES6 declarations such as "CONST" or "LET".


Priyaraj17:
Can someone please review this pull request?

Priyaraj17:
Hey @openlibrary is it okay?

PR #696, opened 2021-04-06
Feature/pow2 reduce

cdrini:
Before/After of ImageCache.cache when zooming in 3 times in 2up mode:

image

The cache now knows to only care about pow2 values.


iisa (approved):
:shipit: 

I diff'd against this https://github.com/cdrini/bookreader/compare/cdrini:feature/page-container...cdrini:feature/pow2-reduce

looks good!

PR #701, opened 2021-04-09
Fix: Load order + rm tiny image requests during loading

cdrini:
Remove loading indicator after JSIA init

The slow part of loading is the loading of the BR JSIA ; once that's done we're good. This also fixes an issue with requests being made for very tiny images during load.

Note there is a petabox-side change which delays the `br.init()` to remove the small image requests; it seems like otherwise it's happening too soon.


iisa (approved):
LGTM

PR #706, opened 2021-04-12
Br fullscreen on scroll

iisa:
Book Nav's fullscreen update - FS after scroll
Testing:
1. go to: https://www-isa.archive.org/details/lakesuperior1850agas
2. when BR loads, click fullscreen button -> BR opens in full window
3. close fullscreen -> scroll down towards details page but keep fullscreen button in view
4. click fullscreen -> BR opens in full window
5. close fullscreen -> user is back to their original scroll position

Dev notes:
- demo updates to propagate rem size & added tall div for scrollability
- 

ezgif com-gif-maker (8)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules/
 coverage/
 
 .vscode
+BookReader/
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ node_modules/
 coverage/
 
 .vscode
-BookReader/
--- a/BookReaderDemo/BookReaderDemo.css
+++ b/BookReaderDemo/BookReaderDemo.css
@@ -1,6 +1,10 @@
 /*
  * Custom overrides for BookReader Demo.
  */
+html {
+    font-size: 10px;
+}
+
 body {
     background-color: #939598;
     margin: 0px;
--- a/src/BookNavigator/BookNavigator.js
+++ b/src/BookNavigator/BookNavigator.js
@@ -303,29 +303,22 @@ export class BookNavigator extends LitElement {
    * Manages Fullscreen behavior
    * This makes sure that controls are _always_ in view
    * We need this to accommodate LOAN BAR during fullscreen
-   * @param { Event } event
    */
-  manageFullScreenBehavior(event) {
-    this.emitFullScreenState(event);
+  manageFullScreenBehavior() {
+    this.emitFullScreenState();
 
-    const { isFullscreenActive, fullscreenScrollPos = {} } = this.bookreader;
-    const { scrollX = 0, scrollY = 0 } = fullscreenScrollPos;
-
-    if (!isFullscreenActive) {
+    if (!this.bookreader.isFullscreen()) {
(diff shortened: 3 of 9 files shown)


src/BookNavigator/BookNavigator.js
```diff
@@ -308,12 +308,13 @@ export class BookNavigator extends LitElement {
...
 
-    const { isFullscreenActive } = this.bookreader;
+    const { isFullscreenActive, fullscreenScrollPos = {} } = this.bookreader;
+    const { scrollX = 0, scrollY = 0 } = fullscreenScrollPos;
 
     if (!isFullscreenActive) {
       this.fullscreenMgr.teardown();
     } else {
-      this.fullscreenMgr.setup();
+      this.fullscreenMgr.setup(scrollX, scrollY);
```
cdrini, line 317: We can get `$('.BookReader').offset()` (Or however we read the root BookNavigator element) to give us these values; that way we don't have to keep track of them in `fullscreenScrollPos` at all!

cdrini (changes requested):
One potential cleanup; otherwise lgtm! Still need to test.

.gitignore
```diff
@@ -4,3 +4,4 @@ node_modules/
 coverage/
 
 .vscode
+BookReader/
```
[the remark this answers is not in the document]
iisa: hahah whoops didn't mean to do this!

src/BookNavigator/BookNavigator.js
iisa: I have an interesting dilemma here, when the `fullscreenActive` class gets appended, bc of position: fixed, the bookreader blinks out of view until the height is dynamically set.  this is causing side effects of scroll positions & offsets to effectively return `0` - so as a reaction, I passed it through an instance param (bc passing it through the custom event broke all the things 😞 )
I will retry with your suggestion 🤞

src/css/_BRmain.scss
```diff
@@ -148,7 +148,7 @@
     position: fixed;
     width: 100%;
     height: 100%;
-    max-height: 100%;
+    max-height: 100vh;
```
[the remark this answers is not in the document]
iisa: reverted...for safety 😅

src/BookNavigator/BookNavigator.js
iisa: really good catch. i've updated to reflect our discussion: on fullscreen start, FS Manager will set class.

dualcnhq:
tested demo... lgtm

dualcnhq (approved):

jbuckner (approved):

BookReaderDemo/BookReaderDemo.css
```diff
@@ -1,6 +1,10 @@
 /*
  * Custom overrides for BookReader Demo.
  */
+html {
+    font-size: 10px;
```
cdrini, line 5: Note to self, we need to find a better solution to this :P

cdrini (approved):
Nice! This looks much neater :) 🚀

PR #707, opened 2021-04-12
Remove reference to deprecated pruneUnused

cdrini:
🤦‍♂️ Was getting lots of console.warns, which I think will flood sentry.

Tested that the `pageContainers` object gets cleared when jumping large ranges.


iisa (approved):
Lgtm

PR #712, opened 2021-04-14
page container: add `data-side` attr

iisa:


cdrini (approved):
Tested page-side set in all 3 modes, and that the "fake" pagecontainer on front/end in 2up does not get a page-side.

PR #719, opened 2021-04-19
build should start from clean directory

cdrini:
Noticed a lot of the files that were claimed to be deleted were still there!


iisa (approved):
ok

PR #720, opened 2021-04-19
Split CI into 3 jobs

cdrini:
E2E tests always take the longest, so let's run them concurrently with the other tests and on a separate GitHub Actions Job.


cdrini:
Apparently each one happens separately, so would need to npm i and checkout separately -\_- Not worth it

PR #725, opened 2021-04-22
Add bookmark feature to check if user is signedIn

dualcnhq:
**Problem:**
- add access and integrate bookmarks provider inside menu panel
- check if a user is signed in to load the bookmarks list saved 

**Solution:**
- add/link up BookmarkProvider in BookNavigator
- if a user is signed in, load bookmarks saved on a specific page of the book

**Testing Steps:**
- load a book
- access the side menu panel 
- check if Bookmarks option is displayed
- click Bookmarks option
- login layout will be displayed if user is not signed in
- otherwise, it will display the bookmarks list layout

----
_gif layout to follow_

### DIFF
--- a/src/BookNavigator/bookmarks/ia-bookmarks.js
+++ b/src/BookNavigator/bookmarks/ia-bookmarks.js
@@ -57,6 +57,7 @@ class IABookmarks extends LitElement {
       bookmarks: { type: Array },
       bookreader: { type: Object },
       bookmarkOptions: { type: Object },
+      displayMode: { type: String },
       editedBookmark: { type: Object },
     };
   }
@@ -91,6 +92,7 @@ class IABookmarks extends LitElement {
     this.bookreader = {};
     this.editedBookmark = {};
     this.bookmarkOptions = {};
+    this.displayMode = ''
 
     this.bookmarkColors = [{
       id: 0,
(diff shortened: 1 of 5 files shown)


iisa (changes requested):
hi @dualcnhq this is a solid first pass through the code. it really looks like you have a solid handle as to what is going on.

let's push this abstraction one level up and allow BookNavigator to determine what `<ia-bookmarks>` will display.

primary suggestion: use `this.bookmarksOptions` to determine whether or not <ia-bookmarks> should display given a new field: `displayMode=login|bookmarks` or something like that.

will also look for the unit test[s] in the next round + testing  steps

src/BookNavigator/bookmarks/ia-bookmarks.js
```diff
@@ -444,27 +450,34 @@ class IABookmarks extends LitElement {
...
-          .editedBookmark=${this.editedBookmark}
-          .bookmarks=${{ ...this.bookmarks }}
-          .activeBookmarkID=${this.activeBookmarkID}
-
-          .bookmarkColors=${this.bookmarkColors}
-          .defaultBookmarkColor=${this.defaultColor}
-        ></ia-bookmarks-list>
-        ${this.allowAddingBookmark ? addBookmarkButton : nothing}
-      </section
+        ${this.displayMode === 'login' ? bookmarkLoginLayout : bookmarkSignedInLayout }
```
nsharma123, line 479: Can we say these const like `bookmarkLoginLayout` and `bookmarkLoggedInLayout`. login and signedIn seems not correct?

iisa:
Tested on a.o staging - logged out & logged in views are displaying properly.

src/BookNavigator/bookmarks/ia-bookmarks.js
dualcnhq: refactored to be more semantic and readable

iisa (approved):
worked through side-by-side code review & updated the following:
- using semantic html to display login link
- using getters in <ia-bookmarks> to isolate parts of views
- new component to house login cta `<bookmarks-loginCTA>`
- talked through various best practices & code style
- variable naming - internal component relationships may be different than task scope

PR #731, opened 2021-04-22
Fix BookReader erroring on older iOS

cdrini:
Was getting a "e.ui.mouse.prototype" is undefined error (or something)
on iOS (tested 8-12, errored on all), preventing me from testing there.
Turns out we need some more pieces from jquery-ui for touch-punch.

Oooh, this also fixed it on FF Mobile!
Build file went up ~5kb.

Testing: Confirm https://deploy-preview-731--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala working on iOS 11 .


iisa (approved):
thanks for catching.

PR #732, opened 2021-04-22
Update testing dependencies

cdrini:
These dependencies only affect the tests, so if CI passes, it's good to go!


cdrini:
CI passes! Auto-merging as this is very low risk

PR #733, opened 2021-04-22
Update dev-flow dependencies

cdrini:
npm i --save-dev concurrently@latest \
http-server@latest \
live-server@latest \
node-fetch@latest

concurrently was the only one with updates.


cdrini:
Auto-merging since only a point version update.

PR #734, opened 2021-04-22
Reorg core BR files

cdrini:
Most things were just renaming. Deleted some now redundant files in the root directory.


iisa (approved):
nice and neat https://github.com/cdrini/bookreader/tree/refactor/br-reorg/src

PR #741, opened 2021-04-24
Updates to allow liner notes view to use web components

iisa:
Problem:
- [liner notes](https://archive.org/details/cd_abba-gold-greatest-hits_abba) cannot load in fullscreen because we are updating fullscreen classes in `BookNavigator`

Solution:
- get liner notes to use BookNavigator & its new web components

Dev notes:
- BookNav:
  - add new global event listener `BrBookNav:PostInit` when book nav gets initialized with bookreader
  - initialize fullscreen manager at `BookReader:PostInit` when it receives bookreader to instantiate with correct bookreader selector
  - `sideMenuOpen` toggle now adjusts book by resizing and jumping to current index
- Fullscreen manager:
  - remove static reference to BookReader's selector so we may pass in other selector names
- item nav refresh to match petabox: new `item-nav-header` slot (for lending bar)

Testing:
- local/netlify dev server: https://deploy-preview-741--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala
- archive.org test steps are in: https://webarchive.jira.com/browse/WEBDEV-4375


dualcnhq:
Screen Shot 2021-04-27 at 2 36 18 PM

src/BookNavigator/BookNavigator.js
```diff
@@ -53,14 +52,33 @@ export class BookNavigator extends LitElement {
     this.signedIn = false;
 
     // Untracked properties
-    this.fullscreenMgr = new BRFullscreenMgr();
+    this.fullscreenMgr = new BRFullscreenMgr(this.bookreader?.el);
```
dualcnhq, line 55: why do we need `?` in `this.bookreader` here while in some parts of the code we don't add `?`?

iisa: hi @dualcnhq good catch, there is a timing issue as to when bookreader is available.  i will move this instantiation out of constructor and initialize when the event `BookReader:PostInit` is being handled.

bfalling (approved):
No obvious blockers.

PR #765, opened 2021-05-11
Cancel search

iisa:
### Cancel Search

- Primary Changes:
  - new: `BookReader.cancelSearchRequest` to cancel search from side menu
  - new: `BookReader.cancelSearch` to house all particulars of cancelling search
  - BookNavigator - update so that 
  - UI - add 2 areas where user can cancel search:
  - progress overlay
    - image
  - side panel
  - image

- Extraneous:
  - Error messaging cleanup
  - Extend progress popup maker to receive a close callback
  - light refactoring


src/plugins/search/view.js
```diff
@@ -408,10 +426,10 @@ class SearchView {
   bindEvents() {
     const namespace = 'BookReader:';
 
+    window.addEventListener(`${namespace}SearchCallbackError`, this.handleSearchCallbackError.bind(this));
```
iisa, line 429: do not use jQuery's event, receive CustomEvent instead

src/BookNavigator/BookNavigator.js
```diff
@@ -103,14 +103,21 @@ export class BookNavigator extends LitElement {
...
+         */
+        (brInstance = null, searchUpdates = {}) => {
           if (brInstance) {
             /* refresh br instance reference */
             this.bookreader = brInstance;
           }
           this.updateMenuContents();
-          if (this.brWidth >= 640) { /* open side search menu */
-            this.openSideSearchMenu();
+          const wideEnoughToOpenMenu = this.brWidth >= 640;
```
jbuckner, line 117: Nice

src/BookNavigator/search/search-results.js
```diff
@@ -90,10 +91,7 @@ export class IABookSearchResults extends LitElement {
   }
 
   dispatchSearchCanceled() {
-    this.dispatchEvent(new CustomEvent('bookSearchCanceled', {
-      bubbles: true,
-      composed: true,
-    }));
+    this.dispatchEvent(new CustomEvent('bookSearchCanceled'));
```
jbuckner, line 94: Not a blocker, but for events that don't have any `details`, you can just use `new Event('foo')` instead of `CustomEvent`

src/BookReader.js
```diff
@@ -2227,21 +2227,37 @@ BookReader.prototype._getPageURI = function(index, reduce, rotate) {
...
  */
-BookReader.prototype.showProgressPopup = function(msg) {
+BookReader.prototype.showProgressPopup = function(msg, onCloseCallback) {
   if (this.popup) return;
 
   this.popup = document.createElement("div");
   $(this.popup).prop('className', 'BRprogresspopup');
-  var bar = document.createElement("div");
+
+  if (typeof(onCloseCallback) === 'function') {
```
jbuckner, line 2239: Is the intention to check for `undefined` or specifically if it's a `function`? If it's the former, I might change the check to that since checking for a function makes me think other things can be passed in.

jbuckner: _Suggestion_ I might pull this conditional out to its own method since it's kinda big, ie:
```
if (typeof(onCloseCallback) === 'function') {
  this.addCloseButton(onCloseCallback);
}
```

src/ItemNavigator/ItemNavigator.js
```diff
@@ -84,18 +84,24 @@ export default class ItemNavigator extends LitElement {
...
-    if (menuId) {
-      if (action === 'open') {
-        this.openShortcut(menuId);
-      } else if  (action === 'toggle') {
-        this.openMenu = menuId;
-        this.toggleMenu();
-      }
+    if (!menuId) {
+      return;
+    }
```
jbuckner, line 94: Nice. You could even one-line this since it's so small.

src/BookNavigator/search/search-provider.js
```diff
@@ -51,6 +53,25 @@ export default class {
...
     window.addEventListener('BookReader:SearchResultsCleared', () => { this.onSearchResultsCleared() });
+    window.addEventListener('BookReader:SearchCanceled', (e) => { this.onSearchCanceled(e) });
+  }
+
+  /**
+   * Cancel search handler
+   * resets `searchState`
+   */
+  onSearchCanceled() {
+    searchState = {
```
jbuckner, line 64: _Question_ Where is `searchState` defined?

src/plugins/search/plugin.search.js
```diff
@@ -158,15 +176,48 @@ BookReader.prototype.search = function(term = '', overrides = {}) {
...
+ * cancels AJAX Call
+ * emits custom event
+ */
+BookReader.prototype._cancelSearch = function () {
+  this.searchXHR?.abort();
+  this.searchView.clearSearchFieldAndResults(false);
+  this.searchTerm = '';
+  this.searchXHR = null;
+  this.searchResults = [];
+  window.BRSearchInProgress = () => {};
```
jbuckner, line 206: Does `BRSearchInProgress` get used outside of this plugin? If not, can it be a class property instead of a `window` property?

src/plugins/search/view.js
```diff
@@ -2,11 +2,12 @@ class SearchView {
   /**
    * @param {object} params
    * @param {object} params.br The BookReader instance
+   * @param {function} params.cancelSearch callback when a user wants to cancel search
    *
    * @event BookReader:SearchResultsCleared - when the search results nav gets cleared
    * @event BookReader:ToggleSearchMenu - when search results menu should toggle
    */
-  constructor({ br }) {
+  constructor({ br, cancelSearch = () => {} }) {
```
jbuckner, line 10: _Minor_ I might call this `searchCancelledCallback` since that seems like when it'll be executed

src/BookNavigator/search/search-results.js
iisa: nice! TIL. will update

src/BookNavigator/search/search-provider.js
iisa: it's at the top of the file:
image

src/BookReader.js
iisa: it is chunky, yes. but this file is unique, as it just binds everything to the bookreader object.. i didn't want to muddle those "publicly accessible functions" with this as it is just for that modal...

src/plugins/search/plugin.search.js
iisa: i tried to put it in the class, but jQuery's JSONP callback naming (tho it says in [docs](https://api.jquery.com/jquery.ajax/) it can take a function, it doesn't for our version of jQuery 😢 )
image

src/BookReader.js
iisa: i will hold bc that file is quite brittle 😬  i had to stop myself from making that whole thing a web component 😅

jbuckner (approved):

PR #766, opened 2021-05-12
Fullscreen toggle - wrap in animation flags so BookNav knows that core is animating

iisa:
Problem:
During 1up mode, when toggling fullscreen, booknav's resize caller is breaking fullscreen animation mid-way through.  Result is Black page - the actual page is there, just unseen

Solution:
set animation flags during fullscreen animation

Tech note: - animation sets `opacity` directly to `BRContainer` and gradually changes it from 0 - 1. In order to ensure that there isn't a race condition, we need to set `this.animating` flag at the beginning & unset at the end of the animation.


mmcnl (approved):
LGTM

PR #768, opened 2021-05-12
Remove some dead CSS + temp. Search mmenu rendering

cdrini:
This is no longer needed now search is rendered BookNavigator.


iisa (approved):
Looks good

PR #769, opened 2021-05-13
Add node_modules caching to GitHub Actions

cdrini:
Shaves off ~30 seconds off CI.

First time: npm i took 37s (4m29 total)
Second time: npm i took 10s (3m6 total)

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -18,6 +18,15 @@ jobs:
     - uses: actions/setup-node@v1
       with:
         node-version: 14.x
+    - uses: actions/cache@v2
+      with:
+        # Caching node_modules isn't recommended because it can break across
+        # Node versions and won't work with npm ci (See https://github.com/actions/cache/blob/main/examples.md#node---npm )
+        # But we pin the node version, and we don't update it that often anyways. And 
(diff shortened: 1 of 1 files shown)


.github/workflows/node.js.yml
```diff
@@ -18,6 +18,15 @@ jobs:
...
         node-version: 14.x
+    - uses: actions/cache@v2
+      with:
+        # Caching node_modules isn't recommended because it can break across
+        # Node versions and won't work with npm ci (See https://github.com/actions/cache/blob/main/examples.md#node---npm )
+        # But we pin the node version, and we don't update it that often anyways. And 
+        # we don't use `npm ci` specifically to try to get faster CI flows. So caching
+        # `node_modules` directly.
+        path: 'node_modules'
+        key: ${{ runner.os }}-node-${{ hashFiles('package*.json') }}
```
iisa, line 29: Q: when does it cache bust? Does the cache apply per branch/PR?

cdrini: It caches based on the `package*.json` files; so I believe it should persist between branches/PRs if the package*.json files are the same. A while back I noticed that in some contexts it seems to clear in some odd scenarios, but we've been using this on OL for a while and it seems to be working well.

iisa (approved):
Woohoo!

PR #770, opened 2021-05-13
Make 1up default to fullwidth upto real world size

cdrini:


src/BookReader/utils.js
```diff
@@ -179,7 +179,11 @@ export function calcScreenDPI() {
   const dpi = el.offsetWidth;
   document.body.removeChild(el);
 
-  const screenDPI = dpi * devicePixelRatio;
+  // Do you believe in magic... numbers? We tested on some devices, and the displayed
```
iisa, line 182: :magic:

iisa (approved):
looks good

PR #775, opened 2021-05-24
Full-Text Search results map to wrong page in item

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-4187


src/plugins/search/plugin.search.js
```diff
@@ -199,11 +199,22 @@ BookReader.prototype.search = function(term = '', overrides = {}) {
...
   this.searchResults = results;
+  console.log('11options.goToFirstResult:- ', options.goToFirstResult);
 
   this.updateSearchHilites();
   this.removeProgressPopup();
   if (options.goToFirstResult) {
-    this._searchPluginGoToResult(results.matches[0].par[0].page);
+    console.log('results.matches[0].par[0].page', results.matches[0].par[0].page)
+    let index = results.matches[0].par[0].page;
+    this._searchPluginGoToResult((index % 2 === 0) ? index : --index);
```
iisa, line 209: so this one is intriguing, 

apparently `results.matches[0].par[0].page` maps to => `leafNum` not index.

the image1 below has results on left & page data by indices (152, 153) on the right.

a few notes:
- the change above results in the correct index of `152` = leafNum `153` ([search results return](https://ia802806.us.archive.org/fulltext/inside.php?item_id=threehundredyear00cors_0&doc=threehundredyear00cors_0&path=/33/items/threehundredyear00cors_0&q=mary+mathis) )
- the fn that takes a user to the first result is [here](https://github.com/internetarchive/bookreader/blob/master/src/plugins/search/plugin.search.js#L349). => it uses `br._models.book.getPage(pageIndex)` (uses index, not leafNumber)
- **QA Fails** with this search, 1up mode: `https://ia-petabox-neeraj.archive.org/details/threehundredyear00cors_0/mode/1up?q=sommers+carson`

A possible solution is just to reduce the index by 1. 
_IF_ we go with `--page` to get index from leafNum, we [may] need to confirm that leafNum is always +1 from index? 

I'd prefer to use leafNum to turn the page so that we can just rely on the data...but that is more work.

@cdrini, @nsharma123 what do you think - should we lean on reducing results index by 1 OR get page info by leaf num OR or be able to turn page by LeafNum OR something else...?

img 1:
image

iisa (commented):
got some questions

src/plugins/search/plugin.search.js
iisa: @nsharma123 from our slack discussion thread with @cdrini, let's update `_searchPluginGoToResult` to use `book.getPage(book.leafNumToIndex(leafNum))` somewhere around line 349

src/BookReader.js
```diff
@@ -377,7 +377,7 @@ BookReader.prototype.initParams = function() {
   // Check for Search plugin
   if (this.options.enableSearch) {
     // Go to first result only if no default or URL page
-    this.goToFirstResult = !params.pageFound
+    this.options.goToFirstResult = !params.pageFound
```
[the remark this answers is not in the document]
nsharma123: `this.goToFirstResult` was incorrect because we don't have any property like this. This change will carry the `this.options.goToFirstResult` value to determine if we need to jump on first search result in search.plugin.js file

src/plugins/plugin.resume.js
```diff
@@ -64,5 +64,5 @@ BookReader.prototype.updateResumeValue = function(index, cookieName) {
   // It's likely we can remove resumeCookiePath using getCookiePath()
   const path = this.options.resumeCookiePath
     || this.getCookiePath(window.location.pathname);
-  BookReader.docCookies.setItem(cookieName || 'br-resume', index, ttl, path, null, false);
+  // BookReader.docCookies.setItem(cookieName || 'br-resume', index, ttl, path, null, false);
```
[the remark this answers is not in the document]
nsharma123: Yeah. Just WIP PR.

src/plugins/search/plugin.search.js
```diff
@@ -199,11 +199,22 @@ BookReader.prototype.search = function(term = '', overrides = {}) {
...
 BookReader.prototype.BRSearchCallback = function(results, options) {
   this.searchResults = results;
+  console.log('11options.goToFirstResult:- ', options.goToFirstResult);
 
   this.updateSearchHilites();
   this.removeProgressPopup();
   if (options.goToFirstResult) {
-    this._searchPluginGoToResult(results.matches[0].par[0].page);
+    console.log('results.matches[0].par[0].page', results.matches[0].par[0].page)
+    let index = results.matches[0].par[0].page;
```
iisa, line 208:
```suggestion
    const leafNum = results.matches[0].par[0].page;
    const index = this._models.book.leafNumToIndex(leafNum);
    console.log({ leafNum, index });
```

src/plugins/search/plugin.search.js
iisa, line 209:
```suggestion
    this._searchPluginGoToResult(index);
```

iisa (commented):
I _think_ this will give us the safety net between index & leafNum that we need in order to pass "the best index we can find".

try to run this code, you can see the console.log I temporarily put in for development

iisa (approved):
looks good!

PR #800, opened 2021-06-17
Add 3rd state for sorting volumes and change sorting behavior

dualcnhq:
## Changes:

- add an initial load of files by filename
- update sorting volumes behavior states from:
    - sort ascending by filename
   - sort ascending by title
   - sort descending by title
   - return back to sending by filename
- add index count per volume item
- add neutral/initial state actionButton

br-sorting-states


src/BookNavigator/volumes/volumes-provider.js
```diff
@@ -24,40 +24,52 @@ export default class VolumesProvider {
...
-    const volumesOrderBy = this.isSortAscending ? 'asc' : 'desc';
+  /**
+   * @param {string} sortType (initial, asc, desc)
+   */
+  sortVolumes(sortType) {
+    let sortedFiles = [];
 
-    const sortedFiles = this.viewableFiles.sort((a, b) => {
-      if (this.isSortAscending) return a.title.localeCompare(b.title);
+    sortedFiles = this.viewableFiles.sort((a, b) => {
```
cdrini, line 56: Oh, `sort` doesn't create a copy; it sorts them in place (and returns the same array again)! You'll want to create a copy first, and then call sort.

src/BookNavigator/volumes/volumes-provider.js
```diff
@@ -24,40 +24,52 @@ export default class VolumesProvider {
...
     this.optionChange(this.bookreader);
 
-    if (!initialSort) {
-      this.multipleFilesClicked(volumesOrderBy);
-    }
+    this.multipleFilesClicked(sortType);
   }
 
+  /**
+   * @param {string} orderBy (initial, asc, desc)
```
cdrini, line 71: It's just like TypeScript :)
```suggestion
   * @param {'initial' | 'asc' | 'desc'} orderBy
```

src/BookNavigator/volumes/volumes-provider.js
```diff
@@ -24,40 +24,52 @@ export default class VolumesProvider {
     this.id = 'volumes';
     this.label = `Viewable files (${this.volumeCount})`;
     this.icon = html`${volumesIcon}`;
-    this.actionButton = this.headerIcon;
-    this.sortVolumes(true);
+
+    this.sortOrderBy = "";
```
cdrini, line 28: Add a type here and let's explicitly specify the default.

```suggestion
    /** @type {'initial' | 'asc' | 'desc'} */
    this.sortOrderBy = "";
```

src/BookNavigator/volumes/volumes-provider.js
```diff
@@ -24,40 +24,52 @@ export default class VolumesProvider {
...
+    if (this.sortOrderBy === 'initial') return this.sortInitialIcon;
+    else if (this.sortOrderBy === 'asc') return this.sortAscendingIcon;
+    else return this.sortDescendingIcon;
   }
 
-  sortVolumes(initialSort = false) {
-    this.isSortAscending = !this.isSortAscending;
-    const volumesOrderBy = this.isSortAscending ? 'asc' : 'desc';
+  /**
+   * @param {string} sortType (initial, asc, desc)
```
cdrini, line 51: Type here as well; You can also add:

```js
/** @typedef {'initial' | 'asc' | 'desc'} SortByType */
```

To the top of the doc, and then reference it by name instead of copying it everywhere. Ie.

```suggestion
   * @param {SortByType} sortType
```

cdrini: Actually, we might want to name these something else; like `'orig_sort' | 'title_asc' | 'title_desc'` to make it a little clearer; took me a long time to realize what each one corresponded to.

tests/karma/BookNavigator/volumes/volumes-provider.test.js
```diff
@@ -10,18 +10,21 @@ const brOptions = {
         "/details/SubBookTest": {
           "url_path": "/details/SubBookTest",
           "file_subprefix": "book1/GPORFP",
+          "orig_sort": 1,
```
cdrini, line 13: Is this name set in stone? `orig` is kind of odd, since we don't know when "orig" was set, or what it means. Can this be `"index"`?

src/BookNavigator/volumes/volumes-provider.js
```diff
@@ -24,40 +24,52 @@ export default class VolumesProvider {
...
   get sortDescendingIcon() {
-    return html`<button class="sort-desc icon" aria-label="Sort volumes in descending order" @click=${() => this.sortVolumes()}>${sortDescendingIcon}</button>`;
+    return html`<button class="sort-by desc-icon" aria-label="Sort volumes in descending order" @click=${() => this.sortVolumes('initial')}>${sortDescendingIcon}</button>`;
   }
 
   get headerIcon() {
-    return this.isSortAscending ? this.sortAscendingIcon : this.sortDescendingIcon;
+    if (this.sortOrderBy === 'initial') return this.sortInitialIcon;
+    else if (this.sortOrderBy === 'asc') return this.sortAscendingIcon;
+    else return this.sortDescendingIcon;
```
cdrini, line 47: This is getting bigger now, so we might want to convert to an object:

```suggestion
    // Put this in the constructor, perhaps?
    const sortIcons = {
        index: html`<button ... >`,
        title_asc: html`<button ...>`,
        title_desc: html`<button ...>`,
    };
    
    return sortIcons[this.sortOrderBy];
```

This will let us remove data from the getter names ( <code>sort<b>Initial</b>Icon</code> ).

src/BookNavigator/volumes/volumes-provider.js
```diff
@@ -24,40 +24,52 @@ export default class VolumesProvider {
...
-    return html`<button class="sort-asc icon" aria-label="Sort volumes in ascending order" @click=${() => this.sortVolumes()}>${sortAscendingIcon}</button>`;
+    return html`<button class="sort-by asc-icon" aria-label="Sort volumes in ascending order" @click=${() => this.sortVolumes('desc')}>${sortAscendingIcon}</button>`;
   }
 
   get sortDescendingIcon() {
-    return html`<button class="sort-desc icon" aria-label="Sort volumes in descending order" @click=${() => this.sortVolumes()}>${sortDescendingIcon}</button>`;
+    return html`<button class="sort-by desc-icon" aria-label="Sort volumes in descending order" @click=${() => this.sortVolumes('initial')}>${sortDescendingIcon}</button>`;
   }
 
   get headerIcon() {
```
cdrini, line 44: Should call this `headerButton`, since it's returning a button, not an icon.

```suggestion
  get headerButton() {
```

cdrini: Even better, let's call it `sortButton`.

tests/karma/BookNavigator/volumes/volumes-provider.test.js
```diff
@@ -56,6 +59,22 @@ describe('Volumes Provider', () => {
     expect(provider.component).to.exist;
   });
 
+  describe('sorting cycles - render headerIcon actionButton', async () => {
```
cdrini, line 62: Hmmm, I'm not seeing this test in the karma output 😬 Could it be this isn't even getting run? Actually, I'm not seing `VolumesProvider` in the karam output at all.

Uh oh. In jest, `describe` can't be `async`. Could it be the same in karma? I'm noticing https://github.com/internetarchive/bookreader/blob/master/tests/karma/BookNavigator/book-navigator.test.js uses `it` when it has `async`. It might be the case none of these tests are running 😬

src/BookNavigator/volumes/volumes-provider.js
```diff
@@ -24,40 +24,52 @@ export default class VolumesProvider {
...
+   * @param {string} sortType (initial, asc, desc)
+   */
+  sortVolumes(sortType) {
+    let sortedFiles = [];
 
-    const sortedFiles = this.viewableFiles.sort((a, b) => {
-      if (this.isSortAscending) return a.title.localeCompare(b.title);
+    sortedFiles = this.viewableFiles.sort((a, b) => {
+      if (sortType === 'initial') return a.orig_sort - b.orig_sort;
+      else if (sortType === 'asc') return a.title.localeCompare(b.title);
```
cdrini, line 58: Hmm, not sure why codecov is not realizing this line was run :/ It must've been by your test --- update: I think the test isn't running!

cdrini (changes requested):
Uh oh, I think none of the tests in that file are running 😬

tests/karma/BookNavigator/volumes/volumes-provider.test.js
```diff
@@ -56,6 +59,22 @@ describe('Volumes Provider', () => {
...
+    const provider = new volumesProvider(onSortClick, baseHost, brOptions);
+
+    provider.sortVolumes("initial");
+    expect(provider.headerIcon.getIcon.getHTML()).includes("sort-by neutral-icon");
+
+    provider.sortVolumes("asc");
+    expect(provider.headerIcon.getIcon.getHTML()).includes("sort-by asc-icon");
+
+    provider.sortVolumes("desc");
+    expect(provider.headerIcon.getIcon.getHTML()).includes("sort-by desc-icon");
```
cdrini, line 74: Oh, and question: What is `getIcon` here? I though `headerIcon` was a ``html`<button ... />`;``, so where is the `getIcon` method coming from?

tests/karma/BookNavigator/volumes/volumes-provider.test.js
dualcnhq: we can change it to index in PB side i think

tests/karma/BookNavigator/volumes/volumes-provider.test.js
dualcnhq: `getIcon` was a method from the `TemplateResult` since `get headerIcon` returns a `TemplateResult`

tests/karma/BookNavigator/volumes/volumes-provider.test.js
dualcnhq: got it working now, changed `describe` to `it`

tests/karma/BookNavigator/volumes/volumes-provider.test.js
dualcnhq: keeping `orig_sort` for now

iisa (approved):
looks good! nice upgrade to sorting unit tests, and icon getter.

iisa:
> Uh oh, I think none of the tests in that file are running 😬

hi drini, what tests aren't running? I just looked at runner and it seems to have caught the volumes unit tests https://github.com/internetarchive/bookreader/runs/2864674042?check_suite_focus=true#step:8:294

Screen Shot 2021-06-19 at 6 10 38 AM

tests/karma/BookNavigator/volumes/volumes-provider.test.js
iisa: @cdrini it looks like an index, but if we change algo from under, it will no longer be an index.  I think it's safe to keep as `orig_sort` for now to decouple from actual `index

dualcnhq:
> Uh oh, I think none of the tests in that file are running 😬

all good now, provider tests are running now. thanks @cdrini

PR #812, opened 2021-07-06
Fix keyboard shortcuts interferring with input fields [BC]

### LINKED ISSUE #371: Unable to type "-" or "=" in input fields outside BR

cdrini:
Closes #371 ; Closes #673 ; Closes WEBDEV-4107

Breaking change: Remove public method `BookReader.prototype.keyboardNavigationIsDisabled`

To test:

Go to one of the demos.

Run this JS:

```js
  constructor() {
    super();
    this.shadowRoot.append(document.createElement('input'));
  }
});
$(document.body).prepend($(`<test-element/>`));
```

Type in the added input.

- [x] Confirm keyboard shortcuts work when the input does not have focus
- [x] Confirm keyboard shortcuts do not work when the input does have focus
- [x] Confirm keyboard shortcuts do not work when typing in searchbox in side panel
- [x] Confirm keyboard shortcuts do not work when typing in searchbox in header

### DIFF
--- a/src/BookReader/utils.js
+++ b/src/BookReader/utils.js
@@ -41,6 +41,23 @@ export function notInArray(value, array) {
 }
 
 /**
+ * Determines the active element, going into shadow doms.
+ * @return {Element}
+ */
+export function getActiveElement(doc = document, recurseShadowDom = true) {
+  const activeElement = doc.activeElement;
+  if (recurseShadowDom && activeElement?.shadowRoot) {
+    return getActiveElement(activeElement.shadowRoot, true);
+  }
+  return activeElement;
+}
+
+/** Check if an input field is active. Also checks shadow DOMs. */
+export function isInputActive(doc = document) {
+  return getActiveElement(doc)?.tagName == "INPUT";
+}
+
+/**
  * @param {HTMLIFrameElement} iframe
  * @return {Document}
  */
--- a/src/BookReader.js
+++ b/src/BookReader.js
@@ -659,7 +659,7 @@ BookReader.prototype.setupKeyListeners = function() {
   $(document).keydown(function(e) {
 
     // Keyboard navigation
-    if (!self.keyboardNavigationIsDisabled(e)) {
+    if (!utils.isInputActive()) {
       switch (e.keyCode) {
       case KEY_PGUP:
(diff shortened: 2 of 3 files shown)


src/BookReader/utils.js
```diff
@@ -40,6 +40,23 @@ export function notInArray(value, array) {
...
 }
 
+/**
+ * Determines the active element, going into shadow doms.
+ * @return {Element}
+ */
+export function getActiveElement(doc = document, recurseShadowDom = true) {
+  const activeElement = doc.activeElement;
+  if (recurseShadowDom && activeElement?.shadowRoot) {
+    return getActiveElement(activeElement.shadowRoot, true);
```
iisa, line 50: smol q: this shouldn't be too expensive, ya?

iisa (approved):
looks good! thanks for jumping on this.
had 1 q

src/BookReader/utils.js
cdrini: Hopefully? :P It should only have to go through I'd guess like max 10? We might want to optimize to only run this if the key pressed is on of interest. Right now it runs for any kepress, before determining if it's of interest. Let me see how hard that would be...

iisa: maybe if we debounce/throttle that keypress event handler, it _may_ be enough..

jbuckner: Is there a way to be more explicit about this than traversing the DOM? ie add a method to the components that can be active and call `component.isActive()`?

cdrini: @iisa Switched order of calls :+1:

We'd need something like `isKeyboardSink()` or something. This isn't the _best_ solution as is; I think there are ways one can fully block shadow DOM access to your web component. But it solves the current problem for our components, and we can always change it later :) This also has the benefit of working for other people who might embed BR in their sites, which might not be using our web component methods.

PR #813, opened 2021-07-07
Add script for updating test deps, and update test deps!

cdrini:
These dependencies only affect testing, so they can be updating together, and if CI passes, it's good to merge!


cdrini:
Auto-merging; largely just dependency updates :+1:

PR #815, opened 2021-07-08
Convert search/tts hilites into an SVG layer

cdrini:
Now we don't have to manually update their positions, since they just scale like images! Pre-requisite refactor for Pinch zoom.

Notes:
- ~This does increase the memory footprint a bit :/ Since page containers could be kept around for a potentially long time, and precached/cached, I needed a way to make sure all layers were up-to-date.~
- ~Actually, it introduces a memory leak :/~
    - Both fixed!

### DIFF
--- a/src/BookReader.js
+++ b/src/BookReader.js
@@ -597,11 +597,9 @@ BookReader.prototype.resize = function() {
     if (this.onePage.autofit != 'none') {
       this.resizePageView1up();
       this.centerPageView();
-      if (this.enableSearch) this.updateSearchHilites(); //deletes highlights but does not call remove()
     } else {
       this.centerPageView();
       this.displayedIndices = [];
-      if (this.enableSearch) this.updateSearchHilites(); //deletes highlights but does not call remove()
       this.drawLeafsThrottled();
     }
   } else if (this.constModeThumb == this.mode) {
@@ -1030,7 +1028,6 @@ BookReader.prototype.switchMode = function(
   }
 
   this.trigger(BookReader.eventNames.stop);
-  if (this.enableSearch) this.removeSearchHilites();
 
   this.prevReadMode = this.getPrevReadMode(this.mode);
 
--- a/src/BookReader.js
+++ b/src/BookReader.js
@@ -267,6 +267,32 @@ BookReader.prototype.setup = function(options) {
   });
 };
 
+/**
+ * Get all the HTML Elements that are being/can be rendered.
+ * Includes cached elements which might be rendered again.
+ */
+BookReader.prototype.getActivePageContainerElements = function() {
+  let containerEls = Object.values(this._modes.mode2Up.pageContainers).map(pc => pc.$container[0]);
+  if (this.mode != this.constMode2up) {
+    containerEls = containerEls.concat(this.$('.BRpagecontainer').toArray());
+  }
+  return containerEls;
+};
+
+/**
+ * Get the HTML Elements for the rendered page. Note there can be more than one, since
+ * (at least as of writing) different modes can maintain different caches.
+ * @param {PageIndex} pageIndex
+ */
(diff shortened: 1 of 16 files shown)


iisa:
### QA:

✅  on-page result highlights scales correctly when zooming in/out
✅  i can select text when zoomed in/out

q: where's the jira ticket for this? reviewing code, now...

src/BookReader.js
```diff
@@ -267,6 +267,32 @@ BookReader.prototype.setup = function(options) {
...
+/**
+ * Get the HTML Elements for the rendered page. Note there can be more than one, since
+ * (at least as of writing) different modes can maintain different caches.
+ * @param {PageIndex} pageIndex
+ */
+BookReader.prototype.getActivePageContainerElementsForIndex = function(pageIndex) {
+  const mode2UpContainer = this._modes.mode2Up.pageContainers[pageIndex]?.$container?.[0];
+  let containerEls = mode2UpContainer ? [mode2UpContainer] : [];
+  if (this.mode != this.constMode2up) {
+    containerEls = containerEls.concat(this.$(`.pagediv${pageIndex}`).toArray());
```
iisa, line 291: q: what happens when 2up container exists && 1up/thumb containers exist for the same index.. will there be a possible drift between views or will it not matter? if the latter, where in the code does it get refreshed?

src/plugins/tts/plugin.tts.js
```diff
@@ -233,12 +236,12 @@ BookReader.prototype.ttsStop = function () {
...
-    .then(() => {
-      this.ttsHighlightChunk(chunk);
-      this.ttsScrollToChunk(chunk);
-    });
+BookReader.prototype.ttsBeforeChunkPlay = async function(chunk) {
+  await this.ttsMaybeFlipToIndex(chunk.leafIndex);
+  this.ttsHighlightChunk(chunk);
+  // This appears not to work; ttsMaybeFlipToIndex causes a scroll to the top of
+  // the active page :/ Disabling cause the extra scroll just adds an odd jitter.
+  // this.ttsScrollToChunk(chunk);
```
iisa, line 244: do we want to keep this as memo? will it turn back on?

src/plugins/tts/plugin.tts.js
```diff
@@ -281,12 +284,22 @@ BookReader.prototype.ttsMaybeFlipToIndex = function (leafIndex) {
  * @param {PageChunk} chunk
  */
 BookReader.prototype.ttsHighlightChunk = function(chunk) {
+  // The poorly-named variable leafIndex
+  const pageIndex = chunk.leafIndex;
```
iisa, line 288: 😭 🍂 📖 #️⃣

src/plugins/tts/plugin.tts.js
cdrini: I think so :/ We probably do want to re-enable this at some point. Otherwise as the TTS highlights the next chunk, it doesn't scroll to make sure it's in view :/

src/plugins/tts/plugin.tts.js
cdrini: :P

src/BookReader.js
cdrini: Currently, Mode2Up is the only mode that caches PageContainers. So even when it's not enabled, it could still have PageContainers that need to be updated. Mode1Up/ModeThumb only have PageContainers for what's currently rendered into the HTML; so when they're not enabled, they don't having anything (I believe). Does that answer your question?

Longer term we might want all the modes sharing the same PageContainer cache, but for now each has its own.

cdrini:
Voila the JIRA: https://webarchive.jira.com/browse/WEBDEV-4580

cdrini:
@iisa also tested on iOS; I tested on Windows/Android, so merge time!

PR #817, opened 2021-07-27
loading spinner: sometimes in safari, ring does not spin around

iisa:
**Problem**: sometimes in safari, the loading spinner does not spin properly.

**Solution**: set `display: block` on inline-by-default element that is spinning

**Testing**:
1. use safari, go to [review app](https://www-isa3.archive.org/details/matthewabc0000cata) - ring in loader spins around book properly
2. use firefox, go to [review app](https://www-isa3.archive.org/details/matthewabc0000cata) - ring in loader spins around book properly
3. use chrome, go to[review app](https://www-isa3.archive.org/details/matthewabc0000cata) - ring in loader spins around book properly

**Dev notes**: 

Very intriguing issue, CSS animation and transform for Safari 13+ say that it does not need -web-* prefix for those properties but sometimes, Safari will improperly display loading spinner - spinner’s square.
ref - caniuse: [animation](https://caniuse.com/?search=animation), [transform](https://caniuse.com/?search=transform)

Come to find out, transform on safari needs to explicitly be a non-inline element.
See stack overflow issue [here](https://stackoverflow.com/questions/33745221/rotate-animation-not-working-on-safari-9) with [jsfiddle](http://jsfiddle.net/4hwMm/1/) showing inline element transforming when `display: inline-block|block;` is set.


dualcnhq:
✅  tested in Brave, Firefox and Safari

dualcnhq (approved):

PR #824, opened 2021-08-05
Textareas should also be keyboard shortcut deadzones

cdrini:
To test:

Go to one of the demos.

Run this JS:

```js
  constructor() {
    super();
    this.shadowRoot.append(document.createElement('textarea'));
  }
});
$(document.body).prepend($(`<test-element/>`));
```

Type in the added input.

- [x] Confirm keyboard shortcuts work when the input does not have focus
- [x] Confirm keyboard shortcuts do not work when the input does have focus
- [x] Confirm keyboard shortcuts do not work when typing in searchbox in side panel
- [x] Confirm keyboard shortcuts do not work when typing in searchbox in header

### DIFF
--- a/src/BookReader/utils.js
+++ b/src/BookReader/utils.js
@@ -52,9 +52,10 @@ export function getActiveElement(doc = document, recurseShadowDom = true) {
   return activeElement;
 }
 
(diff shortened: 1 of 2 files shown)


iisa (approved):
:lgtm: TY!

PR #828, opened 2021-08-23
Update test deps

cdrini:


cdrini:
Auto-merging; testing dep updates, and CI passes :+1:

PR #830, opened 2021-08-23
Update build deps

cdrini:


cdrini:
Auto-merging; dep updates. Tested on FF+Win10, iOS11 via BS.

PR #854, opened 2021-09-21
Move jest tests into separate directory

cdrini:
@mc2 :)


dualcnhq (commented):
quick question, is there a way for us to support `'src/...'` instead of `'../../src/...'` in webpack config?

cdrini:
@dualcnhq great question! I haven't been able to find a way :/ There are a few jest options, but they break the autocomplete in vs code :/ It's probably possible though.

PR #860, opened 2021-09-29
Replace deprecated babel-eslint with successor

cdrini:


cdrini:
Auto-merging; small test dep update

PR #867, opened 2021-10-07
Added voice selection menu to tts

mekarpeles:
- Checks whether any voices are available (e.g. doesn't show for Festival), if so, shows voice selection menu to right of other tts options
- If you select a voice mid-stream, it should stop utterances and recompile with the new selected voice

This still needs iconography from @pezvi + CR from @cdrini or @iisa 

Requires testing across many systems:
* [x] tested on Linux + Chrome

127 0 0 1_8080_BookReaderDemo_demo-internetarchive html_ocaid=theworksofplato01platiala

image

### DIFF
--- a/src/css/_controls.scss
+++ b/src/css/_controls.scss
@@ -52,7 +52,7 @@
   select.playback-voices {
       width: 30px;
       background: transparent url("icons/voice.svg") 50% 0 no-repeat;
-      background-size: 24px 18px;
+      background-size: 18px;
   }
 
   .active {


mekarpeles:
Now includes @pezvi's graphics + we use `voice.lang` - `voice.name` as a composite select `option` so when the menu is collapsed the patron gets usable feedback / labeling.

127 0 0 1_44073_BookReaderDemo_demo-internetarchive html_ocaid=theworksofplato01platiala

PR #871, opened 2021-10-15
Make jest tests support root-level importing

cdrini:
This avoids the headache of nested ../../../src, and we still get auto-complete in vs code!

CC @dualcnhq


cdrini:
Note I couldn't find a way to make karma do this; we can do that in a future PR

PR #893, opened 2021-11-30
Sync volumes sorting state in URL + new UrlPlugin

dualcnhq:


src/plugins/plugin.url.js
```diff
@@ -283,19 +286,23 @@ export class UrlPlugin {
...
+          urlState[schema.deprecated_for] = urlStrSplitSlashObj[schema.name];
+          return;
+        }
 
-        if (hasPropertyKey)
-          return urlState[schema.name] = urlStrSplitSlashObj[schema.name];
+        if (hasPropertyKey) {
+          urlState[schema.name] = urlStrSplitSlashObj[schema.name];
+          return;
+        }
```
cdrini, line 297: Put these changes in your other PR, and then rebase this ; that way this PR becomes just about the volume sorting stuff!

cdrini (changes requested):
So close! 🎉

cdrini (approved):
Ok, lgtm! In a future PR, we should start deleting all the old refs to the old URL plugin 🎉 


Tested in FF95:

- Flipping pages updates URL
- Changing url and reloading opens to right page

- Changing volume order updates url
- Loading page with volume order param shows correct order

- Clicking volume reloads page with that volume
- Trailing slash works https://www-clau.archive.org/details/SubBookTest/

- Doing search shows up in url
- url with search auto-triggers search

iisa (approved):

PR #897, opened 2021-12-06
Restrict keyboard shortcuts to when BookReader has focus

mc2:
* Flag if BookReader has "focus" for keyboard shortcuts
   * Initially true, set to false when:
       * BookReader scrolled out of view
       * User clicked outside of BookReader or NavBar
       * User tabbed to element outside of BookReader or NavBar
       
For testing steps, see: https://webarchive.jira.com/browse/WEBDEV-4610


dualcnhq (approved):

src/BookReader.js
```diff
@@ -98,7 +98,7 @@ BookReader.optionOverrides = {};
  * It is separate from the constructor, so plugins can extend.
  * @param {BookReaderOptions} options
  */
-BookReader.prototype.setup = function(options) {
+BookReader.prototype.setup = function (options) {
```
cdrini, line 101: Let's revert these formatting changes ; clutters the history/diff.

src/BookReader.js
```diff
@@ -662,93 +673,146 @@ BookReader.prototype.resize = function() {
...
+    this.hasKeyFocus = this.isKeyFocus(event.target);
+  });
+
+  // Keyboard focus by tabbed element location
+  document.addEventListener('focusin', (event) => {
+    this.hasKeyFocus = this.isKeyFocus(event.target);
+  });
+
+  // Keyboard focus by BookReader in viewport
+  const brArea = document.getElementById('BookReader');
```
cdrini, line 719: We don't need to assume this id ; this should be stored here.

```suggestion
  const brArea = this.refs.$br;
```

src/BookReader.js
```diff
@@ -662,93 +673,146 @@ BookReader.prototype.resize = function() {
...
-    case KEY_PGUP:
-    case KEY_UP:
-      // In 1up mode page scrolling is handled by browser
-      if (!utils.isInputActive() && self.constMode2up == self.mode) {
-        e.preventDefault();
-        self.prev();
-      }
+/**
+ * Returns true if element in area to activate BookReader keyboard focus
+ * @param {object} element
```
cdrini, line 691:
```suggestion
 * @param {HTMLElement} element
```

src/BookReader.js
```diff
@@ -662,93 +673,146 @@ BookReader.prototype.resize = function() {
...
 /**
- * Binds keyboard event listeners
+ * Intersection observer callback sets BookReader keyboard focus
+ * off when the BookReader is not in the viewport.
+ * @param {object} entries
+ * @param {function} observer
  */
-BookReader.prototype.setupKeyListeners = function() {
-  var self = this;
+BookReader.prototype.handleIntersection = function (entries, observer) {
```
cdrini, line 681: This doesn't seem significant enough to get it's own function? Inlining it would be clearer for me. It also doesn't make too much sense out of context, since I don't know what element this intersection observer is currently being applied to.

src/BookReader.js
```diff
@@ -662,93 +673,146 @@ BookReader.prototype.resize = function() {
...
+ * @param {object} element
+ * @returns {boolean}
+ */
+BookReader.prototype.isKeyFocus = function (element) {
+  if (element.closest('#theatre-ia')) {
+    return true;
+  } else if (element.closest('#topnav')) {
+    return true;
+  }
+  return false;
```
cdrini, line 700: Ah, this is a little non-ideal ; is this why you were asking about how to test standalone BR? This seems like it might be the simplest solution:

```suggestion
  return Boolean(element.closest('#theatre-ia, #topnav, ia-bookreader, #BookReader'));
```

Longer term, this should only need `ia-bookreader`, but we're not there yet.

src/BookReader.js
```diff
@@ -662,93 +673,146 @@ BookReader.prototype.resize = function() {
...
+
+  // Keyboard focus by BookReader in viewport
+  const brArea = document.getElementById('BookReader');
+  const options = {
+    root: null,
+    rootMargin: '0px',
+    threshold: [0, 1],
+  };
+  const observer = new IntersectionObserver(this.handleIntersection, options);
+  observer.observe(brArea);
```
cdrini, line 726: Combine these into a single statement ; the extra variables are making it more confusing.

src/BookReader.js
```diff
@@ -264,13 +264,22 @@ BookReader.prototype.setup = function(options) {
...
     reduceSet: this.reduceSet,
   });
+
+  /**
+   * Flag if BookReader has "focus" for keyboard shortcuts
+   * Initially true, set to false when:
+   * - BookReader scrolled out of view
+   * - User clicked outside of BookReader or NavBar
+   * - User tabbed to element outside of BookReader or NavBar
+   */
```
cdrini, line 274: Thank you for the wonderful docs!

src/BookReader.js
cdrini: I'm confused about `#topnav` ; I think you mentioned it earlier, but I didn't fully understand. Can you expand as to why that's needed?

cdrini (changes requested):
Some code q changes. Otherwise looks good!

Big things:
- [x] Add some unit tests!
- [x] Revert the formatting changes

Check my understanding:

1) Scrolling away to the point that bookreader is no longer visible results in keyboard shortcuts being disabled (sg)
2) Clicking outside of BR results in keyboard shortcuts being disabled _unless_ they click on topnav (why is topnav special?)
3) Tabbing outside of BR results in keyboard shortcuts being disabled _unless_ they tab into topnav.

Is that correct?

src/BookReader.js
mc2: Thanks, re-read https://developer.mozilla.org/en-US/docs/Web/API/Element/closest about selector list.

> `#topnav`
If:
- clicking in
- tabbing to controls 
in TopNav area, then so long as one doesn't activate them, i.e., leave the page, one is ... sensibly ... in BookReader, rather than "page" focus.

That is, simply, a narrative use case that goes like this: 
"I've navigated to a book, and I see some stuff at the top. I'll click something, for example, the user icon, and see a drop down menu. When I don't choose anything in that menu, I expect BookReader to have keyboard shortcut focus."

mc2:
> Check my understanding:

1. Scrolling away to the point that bookreader is no longer visible results in keyboard shortcuts being disabled (sg)
1. Clicking outside of BR results in keyboard shortcuts being disabled unless they click on topnav (why is topnav special?)
1. Tabbing outside of BR results in keyboard shortcuts being disabled unless they tab into topnav.
Is that correct?

Yes. 

> why is topnav special?
If:
- clicking in
- tabbing to controls
in TopNav area, or banner
then so long as one doesn't activate them, i.e., leave the page, one is ... sensibly ... in BookReader, rather than "page" focus.
That is, simply, a narrative use case that goes like this:
"I've navigated to a book, and I see some stuff at the top. I'll click something, for example, the user icon, and see a drop down menu. When I don't choose anything in that menu, I expect BookReader to have keyboard shortcut focus."

src/BookReader.js
```diff
@@ -662,87 +671,140 @@ BookReader.prototype.resize = function() {
...
+
+  // Keyboard focus by tabbed element location
+  document.addEventListener('focusin', (event) => {
+    this.hasKeyFocus = this.isKeyFocus(event.target);
+  });
+
+  // Keyboard focus by BookReader in viewport
+  //
+  // Intersection observer and callback sets BookReader keyboard
+  // "focus" flag off when the BookReader is not in the viewport.
```
jbuckner, line 709: Nice use of the `IntersectionObserver`!

src/BookReader.js
```diff
@@ -662,87 +671,140 @@ BookReader.prototype.resize = function() {
 };
 
 /**
- * Binds keyboard event listeners
+ * Returns true if passed element activates or maintains BookReader keyboard shortcut focus
```
jbuckner, line 674: Does this mean when users interact with these elements like the `item-navigator`, `#topnav` and `#donate_banner`, the BookReader keyboard shortcuts continue to work? (and conversely do other elements cause it them to stop?)

src/BookReader.js
```diff
@@ -662,87 +671,140 @@ BookReader.prototype.resize = function() {
...
+ * @param {HTMLElement} element
+ * @returns {boolean}
  */
-BookReader.prototype.setupKeyListeners = function() {
-  var self = this;
+BookReader.prototype.isKeyFocus = function (element) {
+  // BookReader and ItemNavigator
+  let selectors = this.el;
+  selectors += ',' + 'item-navigator';
```
jbuckner, line 681: Are these all elements from petabox? Just wondering if it's possible to not hard code these since it kind of binds this to the petabox structure, but it may not be.

jbuckner (commented):
A couple of questions, but overall looks good!

src/BookReader.js
```diff
@@ -662,87 +671,140 @@ BookReader.prototype.resize = function() {
...
+      if (entry.intersectionRatio === 0) {
+        this.hasKeyFocus = false;
       }
+    });
+  }, {
+    root: null,
+    rootMargin: '0px',
+    threshold: [0, 1],
+  });
+  observer.observe(this.refs.$br);
```
cdrini, line 721: Looking at the CI output: https://github.com/internetarchive/bookreader/runs/4465817244?check_suite_focus=true#step:10:193

It looks like this is what's causing it to error; `$br` is a jquery el! Need to convert it to an HTMLElement

```suggestion
  observer.observe(this.refs.$br[0]);
```

mc2: Thanks @cdrini -- already caught and fixed this

src/BookReader.js
mc2: Yes

src/BookReader.js
mc2: The BookReader container structure is in flux. I expect this to change -- and tried to isolate all the changes here, but we're going to have to monitor it.

It still works outside of petabox.

src/BookReader.js
mc2: Thank you, @jbuckner

src/BookReader.js
jbuckner: Gotcha. So when users interact with those elements, do they have the ability to use keyboard navigation on them?

mc2: Yes, 
- different keys -- or so I thought -- retesting, the donation banner radio buttons use up/down/right/left keys. These need to _not_ be take over as keyboard shortcuts.

src/BookReader.js
```diff
@@ -662,87 +671,114 @@ BookReader.prototype.resize = function() {
...
-    case KEY_UP:
-      // In 1up mode page scrolling is handled by browser
-      if (!utils.isInputActive() && self.constMode2up == self.mode) {
-        e.preventDefault();
-        self.prev();
+  // Keyboard focus by BookReader in viewport
+  //
+  // Intersection observer and callback sets BookReader keyboard
+  // "focus" flag off when the BookReader is not in the viewport.
+  const observer = new IntersectionObserver((entries) => {
```
cdrini, line 682: Wrap in an `if (window.IntersectionObserver)` ; Breaking keyboard shortcuts on iOS11

tests/jest/BookReader.keyboard.test.js
```diff
@@ -0,0 +1,190 @@
...
+  test('Focus flag disables', () => {
+    br.next = jest.fn();
+    br.hasKeyFocus = false;
+    const keyEvent = new KeyboardEvent('keydown', {'key': 'ArrowDown'});
+    keyEvent.preventDefault = jest.fn();
+    document.dispatchEvent(keyEvent);
+    expect(br.next).toHaveBeenCalledTimes(0);
+    expect(keyEvent.preventDefault).toHaveBeenCalledTimes(0);
+    // Must reset for following tests
+    br.hasKeyFocus = true;
```
cdrini, line 38: Oh jeez, the state going through these tests 😅 We should see if we can find a way to fix this in the future.

tests/jest/BookReader.keyboard.test.js
```diff
@@ -0,0 +1,190 @@
...
+afterEach(() => {
+  jest.clearAllMocks();
+});
+
+/**
+ * Only run init() once. Otherwise multiple EventListeners will be added.
+ */
+test('Initialzation enables IntersectionObserver and defaults', () => {
+  const observe = jest.fn();
+  window.IntersectionObserver = jest.fn(() => ({
```
cdrini, line 19: Why do we need this? Shouldn't the jest setup file be doing this?

cdrini (approved):
Looks good to me! 1 browser compatability fix. I tested on Safari 11, Win10 + FF95, win10 + chrome96

- [x] f key works! F key works
- [x] Numpad -/+ work
- [x] Scrolling away stops kebyoard shortcuts
- [x] Scrolling back returns them
- [x] Can type `f` in input boxes

src/BookReader.js
mc2: done

tests/jest/BookReader.keyboard.test.js
mc2: Yes.

tests/jest/BookReader.keyboard.test.js
mc2: I could not find a clear way to do it.
The jest setup file is to allow all other tests to run without crashing, although not needed now what we're wrapping it so I am removing it.

iisa:
@mc2 are we ready??

do let us know when we cut the next tag for BR, there are some tickets on sprint board that rely on the next tag

Issue #899, opened 2021-12-14
Make eslint not complain about $.browser on this line


cdrini:
We keep getting things like this in our PRs:

![image](https://user-images.githubusercontent.com/6251786/145913651-0292f9a4-0343-41aa-92e7-f615ca5f5556.png)

Let's add an eslint comment to this line to disable the `$.browser` rule. Note:
- We don't want to disable the `$.browser` rule globally!
- We don't want to disable all linting rules for this line! Just $.browser.

PR #900, opened 2021-12-14
Resolved eslint on $.browser

homewardgamer:
Resolves Issue #899 
Added the `//eslint-disable-line` comment to line 26 so that it disables all warnings for that particular line only.
Kindly review @cdrini


cdrini:
Thanks @homewardgamer ! Can we make it so that only the $.browser rule is disabled on this line, instead of all rules?

homewardgamer:
I have changed the comment to `//eslint-disable-line no-jquery/no-browser` to disable the $.browser warning. Kindly check  @cdrini

cdrini (approved):
Awesome, thank you @homewardgamer !

PR #905, opened 2021-12-15
search inside request: turn on cache for request (removes random string in url)

iisa:
Problem: jQuery AJAX JSONP calls are no cache by default.

Even if search endpoint returns cache header, there is a random string in URL that prevents browser to cache request.

Task: set `cache: true` in AJAX config

WEBDEV-4972

### DIFF
--- a/src/plugins/search/plugin.search.js
+++ b/src/plugins/search/plugin.search.js
@@ -213,6 +213,7 @@ BookReader.prototype.search = function(term = '', overrides = {}) {
   return $.ajax({
     url: url,
     dataType: 'jsonp',
+    cache: true,
     beforeSend,
     jsonpCallback: 'BRSearchInProgress'
   }).then(processSearchResults);


jbuckner (approved):

src/plugins/search/plugin.search.js
```diff
@@ -213,6 +213,7 @@ BookReader.prototype.search = function(term = '', overrides = {}) {
   return $.ajax({
     url: url,
     dataType: 'jsonp',
+    cache: true,
```
jbuckner, line 216: That's interesting.. is the default behavior to break the cache each time?

bfalling (approved):
Good catch! LGTM.

src/plugins/search/plugin.search.js
iisa: yes, when `dataType: 'jsonp' https://api.jquery.com/jquery.ajax/

PR #913, opened 2022-01-18
book-nav gets imported by <ia-bookreader>

iisa:
🎉 `<ia-bookreader>` now directly imports `<book-navigator>` into `<ia-item-nav>`'s main slot

### Testing
Go to: https://deploy-preview-913--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala
✅ demo works! 🎉


src/BookReader/Mode1UpLit.js
```diff
@@ -368,7 +368,7 @@ export class Mode1UpLit extends LitElement {
   computeDefaultScale(page) {
     // Default to real size if it fits, otherwise default to full width
     const containerWidthIn = this.visiblePixelsToWorldUnits(this.htmlDimensionsCacher.clientWidth);
-    return Math.min(1, containerWidthIn / (page.widthInches + 2 * this.SPACING_IN));
+    return Math.min(1, containerWidthIn / (page.widthInches + 2 * this.SPACING_IN)) || 1;
```
iisa, line 371: @cdrini - I found that this can return `0` and make `this.scale = 0` -> this is when I found th theater to be empty.  I added a default value of 1 here (as it Is the starting default) - and everything after that seems to play nicely. lmk what you think, hope this is ok.

src/ia-bookreader/ia-bookreader.js
```diff
@@ -26,45 +32,101 @@ export class BookReader extends LitElement {
...
+    this.menuShortcuts = [];
+    this.menuContents = [];
   }
 
-  firstUpdated() {
-    this.createModal();
-  }
+  updated() {
+    if (!this.modal) {
+      this.modal = new ModalManager();
```
jbuckner, line 47: Are you able to create a modal manager like this? I thought you had to add the DOM element.

jbuckner (approved):
One question / possible blocker for creating a modal manager, but looks good otherwise 👍

BookReaderDemo/IADemoBr.js
```diff
@@ -18,6 +18,9 @@ if (openFullImmersionTheater) {
   iaBookReader.fullscreen = openFullImmersionTheater;
 }
 
+const modal = document.querySelector('modal-manager');
```
mc2, line 21: **question**: Why use var here?
why not
```
iaBookReader.modal = document.querySelector('modal-manager');
```
?

BookReaderDemo/demo-internetarchive.html
```diff
@@ -88,19 +88,17 @@ <h3 id="placeholder">please wait as we are fetching the following: </h3>
     <h3>Placeholder div to allow scrolling</h3>
   </div>
 
+  <modal-manager></modal-manager>
```
mc2, line 91: question: if this is a demo, and we're changing things, can there be an  explanation here?

src/ia-bookreader/ia-bookreader.js
```diff
@@ -26,45 +32,101 @@ export class BookReader extends LitElement {
...
+    }
   }
-  /* End Modal management */
+
 
   manageFullscreen(e) {
     const { detail } = e;
     const fullscreen = !!detail.isFullScreen;
     this.fullscreen = fullscreen;
+    this.dispatchEvent(new CustomEvent('fullscreenStatusUpdated', { detail: { fullscreen }}));
```
mc2, line 60: suggestion: remove `...status...` => `fullscreenUpdated` or use `...state...` as in `loadingStateUpdated`

src/ia-bookreader/ia-bookreader.js
```diff
@@ -26,45 +32,101 @@ export class BookReader extends LitElement {
...
+
+  }
+
+  loadingStateUpdated(e) {
```
mc2, line 64: suggestion: Nice to have a comment for _which_ loading state. The whole bookreader, book-navigator? I don't know without having to look through the code to find out.

BookReaderDemo/IADemoBr.js
iisa: bc don't really need to pin to class instance

BookReaderDemo/demo-internetarchive.html
iisa: thanks, added

BookReaderDemo/demo-internetarchive.html
```diff
@@ -88,19 +88,18 @@ <h3 id="placeholder">please wait as we are fetching the following: </h3>
     <h3>Placeholder div to allow scrolling</h3>
   </div>
 
+  <!-- Certain features, like Bookmarks, uses modal-manager to draw their interface -->
```
iisa, line 91:
```suggestion
  <!-- Certain features, like Bookmarks, use modal-manager to draw their interface-->
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@internetarchive/bookreader",
-  "version": "5.0.0-29",
+  "version": "5.0.0-30-a",
```
iisa, line 3:
```suggestion
  "version": "5.0.0-29",
```

src/ia-bookreader/ia-bookreader.js
iisa: good catch, why did i think that this automagically unfolds itself... 🤔

iisa: wow! just tried it and it looks good - as long as the `<modal-manager>` is in the DOM. so i will keep ^_^

jbuckner: Do you mean the `new ModalManager()` only works if there's already a `<modal-manager>` element? Wouldn't this conditional only be entered if there's _not_ already a `<modal-manager>`?

If the case is that you already need a `<modal-manager>` element, you should be able to just query select it, ie: `this.modal = document.querySelector('modal-manager');`

src/ia-bookreader/ia-bookreader.js
iisa: it's local to the web component chain, so book-nav, not core br...yet.

src/ia-bookreader/ia-bookreader.js
iisa: it's local to the web component chain, so book-nav, not core br...yet.

src/ia-bookreader/ia-bookreader.js
iisa: i am unsure if i want to reach up to the DOM like that..

>Wouldn't this conditional only be entered if there's not already a <modal-manager>?

hmm, ok, this is a stab at getting ready to receive a `ModalManager` thru dependency injection while working towards a bookreader that can just spin itself up (and the things that it needs) by itself.
(which i am clearly punting on until we have a plan around what the `<ia-bookreader>` looks like)

I guess i am not getting the difference between the two ways to instantiate the modal.

iisa: thanks for checking in with me, currently resolving to - check for dom, if not, load.

PR #919, opened 2022-01-26
Shady dom css shifts

iisa:
updates to support flattened, shadydom 

testing: go to https://deploy-preview-919--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala -> works ✅ 

using https://github.com/internetarchive/iaux-item-navigator/pull/5

### DIFF
--- a/src/ia-bookreader/ia-bookreader.js
+++ b/src/ia-bookreader/ia-bookreader.js
@@ -105,7 +105,7 @@ export class IaBookReader extends LitElement {
 
   render() {
     return html`
-      <div class="ia-bookreader">
+      <div>
         <ia-item-navigator
           ?viewportInFullscreen=${this.fullscreen}
           .basehost=${this.baseHost}
@@ -148,8 +148,6 @@ export class IaBookReader extends LitElement {
     return css`
       :host {
         display: block;
-        height: inherit;
-        min-height: inherit;
         --primaryBGColor: var(--black, #000);
         --secondaryBGColor: #222;
         --tertiaryBGColor: #333;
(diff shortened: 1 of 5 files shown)


cdrini:
[#477](https://github.com/webcomponents/polyfills/issues/477) Strikes again!

src/ia-bookreader/ia-bookreader.js
```diff
@@ -105,7 +105,7 @@ export class IaBookReader extends LitElement {
 
   render() {
     return html`
-      <div class="ia-bookreader">
+      <div class="main-component">
```
cdrini, line 108: Gaah, https://github.com/webcomponents/polyfills/issues/477 ! Let's give this a tighter class name to avoid hitting other things; I started doing `*_root`; e.g. `ia-bookreader__root` so that it can't possibly in any way break 😅

cdrini (commented):
- Fullscreen/normal working in FF, Chrome @ Win10
- Still getting black screen of doom on Safari 9 :/ 
- Safari 11 is showing the book, but missing the vertical toolbar; also on master branch
    - The error is `ReferenceError: Can't find variable: globalThis` ; what's that polyfill file we need to add for lit 2? We should add it to the demo html
- The sidebar no overlaps the view; the book isn't resizing 😬 
image

iisa:
thanks for the quick look @cdrini - i've set as draft as i tinker further.  and thanks for the polyfill check. will add.

src/ia-bookreader/ia-bookreader.js
dualcnhq: TIL about polyfills class naming

PR #920, opened 2022-01-26
Add polyfill for lit 2

cdrini:
Some of the packages we import are already on lit 2, and need this special polyfill file.

CC @iisa


iisa (approved):
LGTM!

i was also thinking maybe we should put a polyfill.io for ios9 issues.

iisa (approved):
lgtm

PR #921, opened 2022-01-26
Update hyphen stitching regex to include dangling "¬" (not signs)

cdrini:
Some books mis-OCR a dangling hyphen as a ¬ (mathematical not sign) . Since in math the not sign should not appear followed by a space, we think we can safely assume this should be replaced.

Reported on twitter: https://twitter.com/neurdy/status/1484567701666156546

Tested with the word "Between" on the right page here, in Chrome: https://archive.org/details/contestedknowled0000seid_x2i5/page/230/mode/2up


iisa (approved):
LGTM

PR #923, opened 2022-01-27
update test deps

cdrini:


cdrini:
Auto-merging since this only touches tests, and CI passes!

Issue #932, opened 2022-02-07
Impossible to use Bookreader with WinXP browsers


AlanHK:
**Description**
Until a week ago I was able to read books on archive,org. Very recently, if I open any book, the page loads then continually reloads about once a second, producing a strobe effect. 

Can't do a screencap of a flashing screen....

This affects the last version of Firefox to run on XP: 52.9.
Also a fork of Firefox: MyPal 29.3.0, that was updated more recently (though not for a few months, so it is not the cause.)

The only version of Chrome that works on XP, 28.0, has been useless for years.

> What should be happening? What will it look like / how will it behave when fixed?

Work as it did a week ago, please!

**Context**

> What operating system? What browser type and version?
WinXP 32 bit
MyPal 29.3.0,
Firefox 52.9.

**Stakeholders**
https://github.com/internetarchive/bookreader
"Target Devices
Note that BookReader is a core part of Archive.org's mission of Universal Access to All Knowledge. Therefore, care must be taken to support legacy browsers. It should still work and be useable on old devices."

PR #935, opened 2022-02-07
specify colorbox-dependent shift styling

iisa:
Closes https://github.com/internetarchive/bookreader/issues/932

Problem: when flatting out web components for older browsers -> `.shift` global class triggered infinite loop for ResizeObserver.  This caused bookreader's theater on archive.org to flicker for desktop/tablet browsers

Solution: Rename global `.shift` class to be very specific to its colorbox case.


src/css/_colorbox.scss
```diff
@@ -46,7 +46,7 @@
         }
         #cboxLoadingOverlay{background:transparent;}
 
-.shift{
+.br-colorbox-shift {
     position: absolute !important;
     left: -10000px !important;
 }
```
iisa, line 52:
```suggestion
}

```

jbuckner (approved):
👏  :shipit:

PR #936, opened 2022-02-07
Shared RO late load & Url Plugin double slash fix

iisa:
Problem: The Shared Resize observer loads after the side menu providers have, thus not conveying proper browser dimensions to those listening for it.

Solution: re-hydrate menus.

Problem: when primary url ends with trailing slash, it appends it, thus adding `//` to url. cannot properly reload page.

Solution: check for `//`

### DIFF
--- a/src/plugins/url/UrlPlugin.js
+++ b/src/plugins/url/UrlPlugin.js
@@ -140,7 +140,14 @@ export class UrlPlugin {
     const concatenatedPath = urlStrPath !== '/' ? urlStrPath : '';
     if (this.urlMode == 'history') {
       if (window.history && window.history.replaceState) {
-        const newUrlPath = `${this.urlHistoryBasePath}${concatenatedPath}`;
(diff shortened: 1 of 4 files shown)


dualcnhq (approved):
LGTM

src/plugins/url/UrlPlugin.js
```diff
@@ -140,7 +140,14 @@ export class UrlPlugin {
     const concatenatedPath = urlStrPath !== '/' ? urlStrPath : '';
     if (this.urlMode == 'history') {
       if (window.history && window.history.replaceState) {
-        const newUrlPath = `${this.urlHistoryBasePath}${concatenatedPath}`;
```
cdrini, line 143:
```
const newUrlPath = `${this.urlHistoryBasePath}/${concatenatedPath}`.replace(/\/+/, '/');
```

cdrini (approved):

PR #941, opened 2022-02-11
Update concurrently + Small speedup to build & test

cdrini:
Update to latest concurrently, and use it to run our build steps in parallel and our test steps in parallel. The new --group option lets the output appear human-readable despite the parallelization

Shaved ~20s on average on my machine (sample size of 5)
image

Oh, test: 101s -> 45s (-55%) !


iisa (approved):
🚢 🍉

PR #943, opened 2022-02-11
Var replaced with let/const

Aadilhassan:


cdrini:
Thanks @Aadilhassan ! This is missing some of the requirements of the initial issue; we should be using const instead of let where possible, and we also want to update the eslintrc.js . Closing because it would be better to start fresh using the command in the issue :)

Issue #945, opened 2022-02-11
Testcafe not running for all test files


cdrini:
Since at least 5.0.0-24 (Nov 23, 2021), testcafe has not been running all our e2e tests 😭 

Sample output:

```
Run npx testcafe --concurrency 2 --quarantine-mode
 Running tests in:
 - Chrome 95.0.4638.69 / Linux 0.0
 - Firefox 94.0 / Linux 0.0

 Autoplay plugin
 ✓ page auto-advances after allotted flip speed and delay

 Viewmode carousel
 ✓ Clicking `view mode` cycles through view modes


 2 passed (7s)
```

I though this might be to a recent testcafe update, but I think it might be the result of https://github.com/internetarchive/bookreader/pull/870

PR #946, opened 2022-02-11
remove overrides block and changed var to let and const

### LINKED ISSUE #937: Replace vars with const/let in src/BookReader.js
In working towards modernizing the BookReader JavaScript, we want to switch to using `const`/`let` (see [here](https://github.com/internetarchive/bookreader/wiki/JS-Modernization-Plan-Notes#uses-constlet-instead-of-var) for why).

Summary of Requirements:

* [ ] Remove the `BookReader.js` override in `.eslintrc.js` that allows `var`
* [ ] There are no occurrences of `var` in `src/BookReader.js`
    * You might try running `npx eslint --fix src/BookReader.js` after removing the override. That might fix everything automagically!
* [ ] `const` is preferred over `let` where possible

sancodes:
Closes #937 

- remove overrides block from .eslintrc.js
- ran `npx eslint --fix src/BookReader.js` after removing the overrides block
- changed `var` to `let` or `const` where needed

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -47,15 +47,5 @@ module.exports = {
     "prefer-const": ["error"],
     "space-infix-ops": ["error", { "int32Hint": false }],
     "eol-last": ["error", "always"]
-  },
-  "overrides": [
-    { // TODO; these should be cleaned up as well
-      "files": [
-        "src/BookReader.js"
-      ],
-      "rules": {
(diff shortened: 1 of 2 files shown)


cdrini (approved):
Lgtm! Did a number of tests manually since our testcafe isn't working correctly right now (#945). Works like a charm, thanks @sancodes !

Issue #940, opened 2022-02-10
Use async/await instead of .then where appropriate


cdrini:
Find all occurrences of `.then` in the codebase (there are 9 matching files) and replace them with `async`/`await` where it makes sense to do so. Read the notes here about why this improves code quality:  https://github.com/internetarchive/bookreader/wiki/JS-Modernization-Plan-Notes#uses-asyncawait-instead-of-then-where-appropriate

Note for simple cases like `.then(r => r.json())` , do _not_ convert those. Those are more legible like that then if we introduced a temporary variable.

Files containing `.then`; note not all of these might need changes!

- [x] `src/BookNavigator/bookmarks/ia-bookmarks.js`
    - [x] #952
    - [x] #961 One more spot remaining in `fetchUserBookmarks`
- [x] `src/plugins/tts/AbstractTTSEngine.js`#952
- [x]  `src/plugins/tts/FestivalTTSEngine.js` #958
- [x] `src/plugins/tts/WebTTSEngine.js`
    - [x] #952
    - [x] #961 `reload`, `finish`, `pause`, `_chromePausingBugFix`, 
- [x] `src/plugins/tts/PageChunk.js` #957
- [x] `src/plugins/tts/PageChunkIterator.js`#952
- [x] `src/plugins/plugin.chapters.js` #957
- [x] `src/plugins/plugin.text_selection.js` #958
- [x] `src/plugins/search/plugin.search.js` #957
- [x] `tests/jest/plugins/search/plugin.search.test.js` #958
- [x] `tests/jest/plugins/tts/AbstractTTSEngine.test.js` #958

sancodes:
if no one is working on this, I'll help

cdrini:
Go for it @sancodes !

sancodes:
@cdrini Thank you, I will be glad to work on other tasks as well. I will get started on this task soon.

padv:
Hi!

I went ahead and did some of the refactoring #951, but I will stop since @sancodes will be handling the rest. he may use some of my code if he wants to. Cheers!

sancodes:
@cdrini not going to lie. Looking at the old code base is very frightening, given the limits I have in my coding experience. Especially, the promise chain, seems to be very complicated for me.
 I still tried my best, 
Please help me make sure if the code is working correctly. I am willing to work on any bug, for the changes I have made.

cdrini:
No worries @padv ! Do you want to fix up the plugin.chapters.js file?

[...]

PR #951, opened 2022-02-14
some async/await instead of .then refactoring

padv:
partially resolving issue #940

### DIFF
--- a/src/BookNavigator/bookmarks/ia-bookmarks.js
+++ b/src/BookNavigator/bookmarks/ia-bookmarks.js
@@ -229,17 +229,14 @@ class IABookmarks extends LitElement {
     return bookmark;
   }
 
-  fetchBookmarks() {
-    return this.api.getAll().then(res => res.text()).then((res) => {
-      let response;
-      try {
-        response = JSON.parse(res);
-      } catch (e) {
-        response = { error: e.message };
-      }
-      return response;
-    }).then((response) => {
-      const {
+  async fetchBookmarks() {
+    let response = await this.api.getAll().then(res => res.text());
+    try {
+      response = JSON.parse(res);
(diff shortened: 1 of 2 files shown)


src/BookNavigator/bookmarks/ia-bookmarks.js
```diff
@@ -229,17 +229,14 @@ class IABookmarks extends LitElement {
...
-      let response;
-      try {
-        response = JSON.parse(res);
-      } catch (e) {
-        response = { error: e.message };
-      }
-      return response;
-    }).then((response) => {
-      const {
+  async fetchBookmarks() {
```
cdrini, line 232: This one looks like it's tackled by @sancodes ' PR, so leaving it there :) Can you revert this file?

src/plugins/plugin.chapters.js
```diff
@@ -190,20 +190,14 @@ BookReader.prototype.getOpenLibraryRecord = function () {
...
-      // try sourceid
-        return $.ajax({ url: `${baseURL}&source_records=ia:${this.bookId}`, dataType: 'jsonp' });
-      }
-    })
-    .then(data => {
-      if (data && data.length > 0) {
-        setUpChapterMarkers(data[0]);
-      }
-    });
+  const data = await $.ajax({ url: fetchUrlByBookId, dataType: 'jsonp' });
```
cdrini, line 193: The logic here isn't quite right. It needs to request the first url to get `data`; and if that doesn't return `data`, it fetches the second request to get `data`. So both `$.ajax` calls should be before the `setUpChapterMarkers` call.

cdrini (changes requested):

cdrini:
This was addressed in #957

PR #952, opened 2022-02-14
update to async await

sancodes:
Work toward #940

Code Review: 
- Need Heavy Code Review, if the logic has been transferred correctly

Style: 
- Ran `npx eslint .`  to make sure the styling is correct.

### DIFF
--- a/src/BookNavigator/bookmarks/ia-bookmarks.js
+++ b/src/BookNavigator/bookmarks/ia-bookmarks.js
@@ -230,39 +230,34 @@ class IABookmarks extends LitElement {
   }
 
   async fetchBookmarks() {
+    const resText = await this.api.getAll().then(r=> r.text());
+    let parsedResponse;
     try {
-      const res = await this.api.getAll();
-      const resText = await res.text();
-      let parsedResponse;
-      try {
-        parsedResponse = await JSON.parse(resText);
-      } catch (e) {
-        parsedResponse = {error : e.message};
-      }
-
-      const {
-        success,
-        error = 'Something happened while fetching bookmarks.',
-        value: bkmrks = [],
-      } = await parsedResponse;
-
-      if (!success) {
-        console?.warn('Error fetching bookmarks', error);
-      }
-
-      const bookmarks = {};
-      Object.keys(bkmrks).forEach((leafNum) => {
-        const bookmark = bkmrks[leafNum];
-        const formattedLeafNum = parseInt(leafNum, 10);
-        const formattedBookmark = this.formatBookmark({ ...bookmark, leafNum: formattedLeafNum });
-        bookmarks[leafNum] = formattedBookmark;
-      });
-
-      this.bookmarks = bookmarks;
-      return bookmarks;
+      parsedResponse = JSON.parse(resText);
     } catch (e) {
-      return e;
+      parsedResponse = {error : e.message};
     }
+
+    const {
+      success,
+      error = 'Something happened while fetching bookmarks.',
+      value: bkmrks = [],
+    } = parsedResponse;
+
+    if (!success) {
+      console?.warn('Error fetching bookmarks', error);
+    }
+
+    const bookmarks = {};
+    Object.keys(bkmrks).forEach((leafNum) => {
+      const bookmark = bkmrks[leafNum];
+      const formattedLeafNum = parseInt(leafNum, 10);
+      const formattedBookmark = this.formatBookmark({ ...bookmark, leafNum: formattedLeafNum });
+      bookmarks[leafNum] = formattedBookmark;
+    });
+
+    this.bookmarks = bookmarks;
+    return bookmarks;
   }
 
   emitBookmarksChanged() {
--- a/src/BookNavigator/bookmarks/ia-bookmarks.js
+++ b/src/BookNavigator/bookmarks/ia-bookmarks.js
@@ -229,21 +229,23 @@ class IABookmarks extends LitElement {
     return bookmark;
   }
 
-  fetchBookmarks() {
-    return this.api.getAll().then(res => res.text()).then((res) => {
-      let response;
+  async fetchBookmarks() {
+    try {
+      const res = await this.api.getAll();
+      const resText = await res.text();
+      let parsedResponse;
       try {
-        response = JSON.parse(res);
+        parsedResponse = await JSON.parse(resText);
       } catch (e) {
-        response = { error: e.message };
+        parsedResponse = {error : e.message};
       }
-      return response;
-    }).then((response) => {
+
       const {
         success,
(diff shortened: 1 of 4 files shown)


src/BookNavigator/bookmarks/ia-bookmarks.js
```diff
@@ -229,21 +229,23 @@ class IABookmarks extends LitElement {
     return bookmark;
   }
 
-  fetchBookmarks() {
-    return this.api.getAll().then(res => res.text()).then((res) => {
-      let response;
+  async fetchBookmarks() {
+    try {
+      const res = await this.api.getAll();
+      const resText = await res.text();
```
cdrini, line 235: These are easier to read if kept together.

```suggestion
      const resText = await this.api.getAll().then(r => r.text());
```

src/BookNavigator/bookmarks/ia-bookmarks.js
```diff
@@ -229,21 +229,23 @@ class IABookmarks extends LitElement {
...
+      let parsedResponse;
       try {
-        response = JSON.parse(res);
+        parsedResponse = await JSON.parse(resText);
```
cdrini, line 238: `JSON.parse` doesn't return a promise, so no need for `await` here.

```suggestion
        parsedResponse = JSON.parse(resText);
```

src/BookNavigator/bookmarks/ia-bookmarks.js
```diff
@@ -229,21 +229,23 @@ class IABookmarks extends LitElement {
...
       }
-      return response;
-    }).then((response) => {
+
       const {
         success,
         error = 'Something happened while fetching bookmarks.',
         value: bkmrks = [],
-      } = response;
+      } = await parsedResponse;
```
cdrini, line 247: The parsedResponse isn't a Promise, so no need for await here.
```suggestion
      } = parsedResponse;
```

src/plugins/tts/AbstractTTSEngine.js
```diff
@@ -147,36 +147,32 @@ export default class AbstractTTSEngine {
...
+    this.opts.onLoadingStart();
+    const sound = this.createSound(chunk);
+    sound.chunk = chunk;
+    sound.rate = this.playbackRate;
+    sound.voice = this.voice;
+    sound.load(() => this.opts.onLoadingComplete());
+
+    this.opts.onLoadingComplete();
+
+    const soundFn = await this.opts.beforeChunkPlay(chunk).then(() => sound);
```
cdrini, line 166: We actually don't need this anymore! We previously needed this because sound was a function parameter, and we wanted it passed forward to the next `.then`. Now `sound` is in the same scope :)

```suggestion
    await this.opts.beforeChunkPlay(chunk);
```

src/plugins/tts/AbstractTTSEngine.js
```diff
@@ -147,36 +147,32 @@ export default class AbstractTTSEngine {
...
+
+    if (!this.playing) return;
+
+    await this.playSound(soundFn);
+    this.opts.afterChunkPlay(soundFn.chunk);
+
+    if (this.paused) this.pause();
```
cdrini, line 173: Hmm, I can't remember exactly why this was the way it was, but I think we might need to keep it the same

```suggestion
    const playPromise = await this.playSound(sound).then(() => this.opts.afterChunkPlay(sound.chunk));
    if (this.paused) this.pause();
    await playPromise;
```

src/plugins/tts/PageChunkIterator.js
```diff
@@ -53,22 +53,22 @@ export default class PageChunkIterator {
...
-    return this._fetchPageChunks(this._cursor.page)
-      .then(chunks => {
-        if (this._cursor.chunk == chunks.length) {
-          this._cursor.page++;
-          this._cursor.chunk = 0;
-          return this._nextUncontrolled();
-        }
-        return chunks[this._cursor.chunk++];
-      });
+    try {
```
cdrini, line 61: I don't believe we need this `try`/`catch` here

src/plugins/tts/WebTTSEngine.js
```diff
@@ -278,16 +278,15 @@ export class WebTTSSound {
...
-            const reloadPromise = this.reload();
-            reloadPromise.then(() => this.play());
-          }
-        });
+      if (result != 'timeout') return;
+
+      this.utterance.dispatchEvent(new CustomEvent('resume', {}));
+      if (resumeMightNotWork) {
+        const reloadPromise = this.reload();
+        reloadPromise.then(() => this.play());
```
cdrini, line 288:
```suggestion
        await this.reload();
        this.play();
```

src/BookNavigator/bookmarks/ia-bookmarks.js
```diff
@@ -229,21 +229,23 @@ class IABookmarks extends LitElement {
     return bookmark;
   }
 
-  fetchBookmarks() {
-    return this.api.getAll().then(res => res.text()).then((res) => {
-      let response;
+  async fetchBookmarks() {
+    try {
```
cdrini, line 233: Don't need this outer try/catch

cdrini (changes requested):
Nice! This looks like a great first stab :) A few small fixes, but otherwise looks good!

sancodes:
> Nice! This looks like a great first stab :) A few small fixes, but otherwise looks good!

@cdrini Will fix this right now. Thank you for the feedback. I'm on it

sancodes:
@cdrini I have fixed the issues that you've addressed above and pushed the code. Please let me know, if I may have missed something.

cdrini:
Nice! The code looks good; let me do some testing! The text to speech code is one of the more complicated async pieces in the codebases, because the browsers have some very broken APIs :P

sancodes:
> Nice! The code looks good; let me do some testing! The text to speech code is one of the more complicated async pieces in the codebases, because the browsers have some very broken APIs :P

Oh sure. for me it was a huge learning process, even when it came to refactoring. I had never seen this many varieties of promises chaining that is being used in the code base.

cdrini (approved):
Tested pausing, resuming, skipping ahead, skipping back, and changing voices on https://deploy-preview-952--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala#page/n3/mode/2up with:

- Chrome 98 + Win 11
- FF 97 + Win 11
- FF 97 + Nexus 5 (Janky, but so is master :/)
- Chrome 98 + Android 6 (Broken, but also broken on master)
- Safari 15 on iPhone (Skipping causes problems; known issue)
- Chrome on some other android device

Ooof, I forgot how hard this code is to test :P

Edit: Not testing the bookmarks changes; they look unlikely to cause issues.

PR #953, opened 2022-02-16
search: remove event loop when clearing results + pass thru which menu is open

iisa:


cdrini (approved):
Lgtm! Confirmed button no longer causes crashes :)

Note the url isn't removing the `q` when the "X" is clicked, but that feels like a lesser bug for another day!

PR #957, opened 2022-02-17
update .then() to async/await

sancodes:
issue #940 

Need review on the changes. Please let me know the issues in the changes and I'll fix them.

### DIFF
--- a/src/plugins/plugin.chapters.js
+++ b/src/plugins/plugin.chapters.js
@@ -192,9 +192,9 @@ BookReader.prototype.getOpenLibraryRecord = async function () {
 
   let data = await $.ajax({ url: fetchUrlByBookId, dataType: 'jsonp' });
 
-  if (!data && data.length < 0) {
+  if (!data || !data.length) {
     // try sourceid
-    data = $.ajax({ url: `${baseURL}&source_records=ia:${this.bookId}`, dataType: 'jsonp' });
+    data = await $.ajax({ url: `${baseURL}&source_records=ia:${this.bookId}`, dataType: 'jsonp' });
   }
 
   if (data && data.length > 0) {
--- a/src/plugins/plugin.chapters.js
+++ b/src/plugins/plugin.chapters.js
@@ -174,7 +174,7 @@ BookReader.prototype.addChapterFromEntry = function(tocEntryObject) {
  * This makes a call to OL API and calls the given callback function with the
  * response from the API.
  */
-BookReader.prototype.getOpenLibraryRecord = function () {
+BookReader.prototype.getOpenLibraryRecord = async function () {
   // Try looking up by ocaid first, then by source_record
   const baseURL = `${this.olHost}/query.json?type=/type/edition&*=`;
   const fetchUrlByBookId = `${baseURL}&ocaid=${this.bookId}`;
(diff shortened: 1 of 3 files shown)


src/plugins/plugin.chapters.js
```diff
@@ -190,20 +190,16 @@ BookReader.prototype.getOpenLibraryRecord = function () {
...
-    .then(data => {
-      if (data && data.length > 0) {
-        setUpChapterMarkers(data[0]);
-      }
-    });
+  let data = await $.ajax({ url: fetchUrlByBookId, dataType: 'jsonp' });
+
+  if (!data && data.length < 0) {
+    // try sourceid
+    data = $.ajax({ url: `${baseURL}&source_records=ia:${this.bookId}`, dataType: 'jsonp' });
```
cdrini, line 197: $.ajax returns a promise

```suggestion
    data = await $.ajax({ url: `${baseURL}&source_records=ia:${this.bookId}`, dataType: 'jsonp' });
```

src/plugins/plugin.chapters.js
```diff
@@ -190,20 +190,16 @@ BookReader.prototype.getOpenLibraryRecord = function () {
...
-      }
-    })
-    .then(data => {
-      if (data && data.length > 0) {
-        setUpChapterMarkers(data[0]);
-      }
-    });
+  let data = await $.ajax({ url: fetchUrlByBookId, dataType: 'jsonp' });
+
+  if (!data && data.length < 0) {
```
cdrini, line 195: Oh the logic for this isn't quite right; it should be this.

(If you're interested, negating an `&&` expression is called applying De Morgan's Law: https://en.wikipedia.org/wiki/De_Morgan%27s_laws#:~:text=not%20(A%20and%20B)%20%3D%20(not%20A)%20or%20(not%20B)%2C )

```suggestion
  if (!data || !data.length) {
```

src/plugins/search/plugin.search.js
```diff
@@ -210,13 +210,14 @@ BookReader.prototype.search = function(term = '', overrides = {}) {
   };
 
   this.trigger('SearchStarted', { term: this.searchTerm, instance: this });
-  return $.ajax({
+  await $.ajax({
```
cdrini, line 213: we'll need something like `return processSearchResults(await $.ajax(...))`, because

- Previously, the result of the `ajax` call was passed into `processSearchResults` by `.then` .
- Previously, result of the whole function was a Promise, that itself contained the result of `processSearchResults`, so we want to keep the `return` just in case someone that called `search` expected that!

src/plugins/tts/PageChunk.js
```diff
@@ -21,27 +21,27 @@ export default class PageChunk {
    * @param {number} leafIndex
    * @return {Promise<PageChunk[]>}
    */
-  static fetch(server, bookPath, leafIndex) {
+  static async fetch(server, bookPath, leafIndex) {
     // jquery's ajax "PromiseLike" implementation is inconsistent with
     // modern Promises, so convert it to a full promise (it doesn't forward
     // a returned promise to the next handler in the chain, which kind of
     // defeats the entire point of using promises to avoid "callback hell")
```
cdrini, line 28: This should no longer necessary, no we're using async/await
```suggestion
```

src/plugins/tts/PageChunk.js
```diff
@@ -21,27 +21,27 @@ export default class PageChunk {
...
-    return new Promise((res, rej) => {
-      $.ajax({
+    try {
```
cdrini, line 29: Nice work! This was a tricky one.

Let's remove the try/catch. We want the error to go forward! Previously we forwarded it by calling the `reject` function. When constructing a promise like this, the `res` (short for `resolve`) function controls the returned value, and the `rej` (short for `reject`) function controls any errors thrown. In this case, we can just let `$.ajax` throw its error like normal :)

cdrini (changes requested):
Awesome, thanks @sancodes ! A few fixes :) Let me know if I can answer any questions!

sancodes:
@cdrini Thank you for the correction. I really appreciate you being patient in even giving me those links and for those corrections. I'm correcting it right away.

sancodes:
@cdrini I have implemented what you have mentioned. Please let me know, if something is not correct.

cdrini (approved):
Tested:
- Chapters demo shows chapter marks
- searching works
- confirmed TTS works

PR #958, opened 2022-02-18
update async await remaining

sancodes:
work on #940 

need heavy review.

### DIFF
--- a/src/plugins/tts/FestivalTTSEngine.js
+++ b/src/plugins/tts/FestivalTTSEngine.js
@@ -91,10 +91,9 @@ export default class FestivalTTSEngine extends AbstractTTSEngine {
    * See https://stackoverflow.com/questions/12206631/html5-audio-cant-play-through-javascript-unless-triggered-manually-once
    * @return {PromiseLike}
    */
-  iOSCaptureUserIntentHack() {
+  async iOSCaptureUserIntentHack() {
     const sound = soundManager.createSound({ url: SILENCE_1MS[this.audioFormat] });
-    return new Promise(res => sound.play({onfinish: res}))
-      .then(() => sound.destruct());
+    await sound.play({onfinish: await sound.destruct()});
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/plugins/tts/FestivalTTSEngine.js
```diff
@@ -91,10 +91,9 @@ export default class FestivalTTSEngine extends AbstractTTSEngine {
    * See https://stackoverflow.com/questions/12206631/html5-audio-cant-play-through-javascript-unless-triggered-manually-once
    * @return {PromiseLike}
    */
-  iOSCaptureUserIntentHack() {
+  async iOSCaptureUserIntentHack() {
     const sound = soundManager.createSound({ url: SILENCE_1MS[this.audioFormat] });
-    return new Promise(res => sound.play({onfinish: res}))
-      .then(() => sound.destruct());
+    await sound.play({onfinish: await sound.destruct()});
```
cdrini, line 96: The order we want this to run is:

1) createSound
2) Play sound;
3) Once sound is finished playing (! this is the async thing), destruct sound.

So we'll still need the `await new Promise(res => sound.play({onfinish: res}))`; That converts the callback `onfinish` to a promise we can `await` !

src/plugins/tts/FestivalTTSEngine.js
```diff
@@ -122,21 +121,18 @@ class FestivalTTSSound {
...
-  play() {
-    return new Promise(res => {
-      this._finishResolver = res;
-      this.sound.play({ onfinish: res });
-    })
-      .then(() => this.sound.destruct());
+  async play() {
+    const destruct = await this.sound.destruct();
+    this._finishResolver = destruct;
+    this.sound.play({ onfinish: destruct});
```
cdrini, line 135: Similarly here; we'll steel need the `new Promise` to promisify the `onfinish` event.

tests/jest/plugins/search/plugin.search.test.js
```diff
@@ -123,28 +123,26 @@ describe('Plugin: Search', () => {
...
   });
 
-  test('SearchCallback event fires when AJAX search returns results', () => {
+  test('SearchCallback event fires when AJAX search returns results', async () => {
     br.init();
     const dfd = br.search('foo');
-    return dfd.then(() => {
-      expect(triggeredEvents()).toContain(`${namespace}SearchCallback`);
-    });
+    await expect(triggeredEvents()).toContain(`${namespace}SearchCallback`);
```
cdrini, line 129: `br.search` returns a promise (that's why we call `.then` on its return); so that's the thing we want to wait for. Actually we can probably remove `dfd` entirely and just call `await br.search('foo')`.

Similarly for the other cases in the file.

cdrini (changes requested):
Awesome, thanks @sancodes ! A few fixes :)

sancodes:
@cdrini fix the bad code. Please let me know, if i missed on the correct logic.

cdrini:
Awesome! Code looks good; I'll do some testing when I'm back at work on Tuesday :) Have a good weekend!

cdrini (approved):
Awesome! Lgtm :)

Tested:
- Text selection works and is loaded in
- TTS works when festival engine is forced (eg https://deploy-preview-958--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala&_forceTTSEngine=festival#page/118/mode/2up )
- Tested passed in CI

Note didn't test iOS + festival; that should no longer be a concern, since only iOS < 7 would use festival.

PR #959, opened 2022-02-19
Remove unused demo bundle

cdrini:
This was moved to BookReader/ a while back


cdrini:
No clue what codecov is doing. Auto-merging since this doesn't affect production.

PR #960, opened 2022-02-20
README: Fix broken link for about plugin directory

duggiefresh:


cdrini (approved):
Good catch, thanks @duggiefresh !

PR #961, opened 2022-02-24
update remaining async/await

sancodes:
work on #940 

- need review

### DIFF
--- a/src/BookNavigator/bookmarks/ia-bookmarks.js
+++ b/src/BookNavigator/bookmarks/ia-bookmarks.js
@@ -151,14 +151,12 @@ class IABookmarks extends LitElement {
     }
   }
 
-  fetchUserBookmarks() {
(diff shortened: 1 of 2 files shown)


cdrini:
Awesome! Code looks good; I'm running through Read Aloud testing :P

cdrini (approved):
Everything lgtm! I tested on
- MS edge + Win 10
    - Found/fixed unrelated bug #962 
- Chrome + Win 10 
- FF + Android 6
    - Weirdly slow to start. And changing voices doesn't work :/ But also happening on prod
- Ch + And 6
    - Doesn't start at first, but changing language kicked it off. Same on prod
- Latest iOS on BrowserStack

Testing steps:
- Listen for at least 20s
- skip forward/back
- pausing/resuming
- Change voice -- confirm sticks
- Change rate -- confirm sticks

PR #962, opened 2022-02-24
Fix MS Edge voice selection not sticking

cdrini:
MS Edge seems to fire the `voiceschanged` event a lot; causing it to reset the voice. So only switch to "Best" voice if user hasn't explicitly selected a voice.


iisa (approved):
looks good

cdrini:
Tested setting voice and moving forward/back in: FF97@Win10, Chr new-ish@Win10, Edge new@win10

PR #964, opened 2022-02-28
Remove package publishing to github

cdrini:
It's been broken for a really long time, and no one uses it, so let's not waste computer cycles building it, eh?


cdrini:
Auto-merging ; small change not affecting our workflows/production code

PR #976, opened 2022-03-09
Try splitting node workflow into different jobs

cdrini:
Saves about ~45s on CI (~5.2min → ~4.5min, sample size ~2). A big chunk is probably from fully skipping `npm install` when we have a `node_modules` cache hit. Then we also parallelize e2e tests with unit test running.

Sample run graph: https://github.com/internetarchive/bookreader/actions/runs/1959466247

image

Before:

image

After:

image


iisa (approved):
looks good!

build pipelines have split and are using cache 👍 

♻️

PR #979, opened 2022-03-09
Try for more testcafe concurrency

cdrini:


cdrini:
Not faster and more unstable :) No thanks!

PR #992, opened 2022-03-10
Group @internetarchive icon updates

cdrini:
This will place all icon updates in one PR, but not auto-merge. Since there are no tests that really cover whether the icons are working correctly, I think it might be best to have these require manual verification/merge.


cdrini:
Auto-merging this one to stop the incoming internet archive icon PRs. Also small change; no trailing commas -_-.

PR #997, opened 2022-03-10
Use range strategy bump for @internetarchive

cdrini:
These repos often contain web components which get registered globally, so we would rather have ranges with these than pinned dependencies


cdrini:
Auto-merging config change; no trailing commas. Want to get the "pin dependencies" PRs fixed.

PR #1006, opened 2022-03-15
Performance improvements to scroll/zooming when text layer is larger

cdrini:
### Perf Improvment 1: Don't re-mount image during scroll/zoom

This didn't have a big impact on experience anecdotally, but I imagine it's probably good. Note the elements are no longer flashing in the devtools.

| Before | After |
| -- | -- |
| <video src="https://user-images.githubusercontent.com/6251786/158461482-e53e70ce-a325-4271-9266-cf1841f0956f.mp4"> | <video src="https://user-images.githubusercontent.com/6251786/158461668-1f7a0fce-4c99-494d-aab4-68c1a47c1e08.mp4"> |

### Perf Improvement 2: Hide text layer during scroll/zoom

This had a big impact on books with large text layers!

https://user-images.githubusercontent.com/6251786/158468209-a580f8d0-c379-4387-a7d8-5c23da8079b9.mp4

### Perf Improvement 3: hide off-screen text layers

Reduces "sticky" feeling when starting scroll/zoom. Note before/after here is relative to Perf improvement 2.

https://user-images.githubusercontent.com/6251786/158490805-cefd5993-1b52-41e5-bec1-078fcb694de5.mp4

- Before: https://lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=sim_weekly-recorder-a-news-paper_1819-12-09_6_17#mode/1up
- After: https://deploy-preview-1006--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=sim_weekly-recorder-a-news-paper_1819-12-09_6_17#mode/1up

Tested on Chrome/FF98 on Windows

Test:

- pinching/zooming, scrolling on mode 1up and 2up
- Selecting text

### DIFF
--- a/src/BookReader/utils/ScrollClassAdder.js
+++ b/src/BookReader/utils/ScrollClassAdder.js
@@ -1,0 +1,31 @@
+/** Adds a class while the given element is experiencing scrolling */
+export class ScrollClassAdder {
+  /**
+   * @param {HTMLElement} element
+   * @param {string} className
+   */
+  constructor(element, className) {
+    /** @type {HTMLElement} */
+    this.element = element;
+    /** @type {string} */
+    this.className = className;
+    this.timeout = null;
+  }
+
+  attach() {
+    this.element.addEventListener('scroll', this.onScroll);
+  }
+
+  detach() {
+    this.element.removeEventListener('scroll', this.onScroll);
+  }
+
+  onScroll = () => {
+    this.element.classList.add(this.className);
+    clearTimeout(this.timeout);
+    // TODO: Also remove class on mousemove, touch, click, etc.
+    this.timeout = setTimeout(() => {
+      this.element.classList.remove(this.className);
+    }, 1000);
+  }
+}
--- a/src/BookReader/utils/ScrollClassAdder.js
+++ b/src/BookReader/utils/ScrollClassAdder.js
@@ -26,6 +26,6 @@ export class ScrollClassAdder {
     // TODO: Also remove class on mousemove, touch, click, etc.
     this.timeout = setTimeout(() => {
       this.element.classList.remove(this.className);
-    }, 1000);
+    }, 600);
   }
 }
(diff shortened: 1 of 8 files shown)


src/BookReader/utils/ScrollClassAdder.js
```diff
@@ -0,0 +1,31 @@
...
+
+  attach() {
+    this.element.addEventListener('scroll', this.onScroll);
+  }
+
+  detach() {
+    this.element.removeEventListener('scroll', this.onScroll);
+  }
+
+  onScroll = () => {
```
jbuckner, line 23: Q: Do you need to bind the `onScroll` event to this class when you use this syntax? I usually have to bind methods when I add them to event listeners, ie:
```
constructor() {
  this.boundScrollHandler = this.onScroll.bind(this);
}

this.element.addEventListener('scroll', this.boundScrollHandler);
```
Otherwise they reference different instances when you remove the listener.

But I use this syntax so not sure if there's a difference.
```
onScroll() {}
```

src/BookReader/utils/ScrollClassAdder.js
```diff
@@ -0,0 +1,31 @@
+/** Adds a class while the given element is experiencing scrolling */
+export class ScrollClassAdder {
```
jbuckner, line 2: Pretty cool solution! 👍

jbuckner (approved):
One question, but looks good otherwise! 👍

src/BookReader/utils/ScrollClassAdder.js
cdrini: There is! Using `onScroll = () => { }` is a class property! Equivalent to `this.onScroll = () { ... }` in the ctor. And all `() => {}` fns have "bound" `this` vars. `onScroll() { }` is a prototype method, and follows the same rules as `Foo.prototype.onScroll = function() { ... }`, and hence needs binding. I'm not crazy over it cause it's a little subtle, but I've been using `() => {}` for anything that's going to be an event handler to avoid calling `.bind()`.

PR #1011, opened 2022-03-31
Fix page number wrong in sidepanel

cdrini:
The numbers were incorrect for some reason. Change them to be consistent with the numbers on the search pins. Still not fully consistent with e.g. slider, but we can solve that another day. They are at least not wildly incorrect now.

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/src/BookNavigator/search/a-search-result.js
+++ b/src/BookNavigator/search/a-search-result.js
@@ -38,15 +38,12 @@ export class BookSearchResult extends LitElement {
   render() {
     /** @type {SearchInsideMatch} */
     const match = this.match;
-    // TODO: Make this not use the global br instance
-    const pageIndex = window.br.leafNumToIndex(match.par[0].page);
-    const pageNumber = window.br.getPageNum(pageIndex);
     const coverImage = html`<img src="${match.cover}" />`;
     return html`
       <li @click=${this.resultSelected}>
         ${match.cover ? coverImage : nothing}
         <h4>${match.title || nothing}</h4>
-        <p class="page-num">Page ${pageNumber}</p>
+        <p class="page-num">Page ${match.displayPageNumber}</p>
         ${this.highlightedHit(match.text)}
       </li>
     `;
--- a/src/BookNavigator/search/a-search-result.js
+++ b/src/BookNavigator/search/a-search-result.js
@@ -1,5 +1,6 @@
 import { html, LitElement, nothing } from 'lit';
 import { unsafeHTML } from 'lit/directives/unsafe-html.js';
+/** @typedef {import('@/src/plugins/search/plugin.search.js').SearchInsideMatch} SearchInsideMatch */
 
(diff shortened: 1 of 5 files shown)


src/BookNavigator/search/a-search-result.js
```diff
@@ -35,17 +36,17 @@ export class BookSearchResult extends LitElement {
...
   render() {
-    const { match } = this;
-    const { par = [] } = match;
-    const [resultDetails = {}] = par;
-    const pageNumber = Number.isInteger(resultDetails.page)
-      ? html`<p class="page-num">Page -${resultDetails.page}-</p>` : nothing;
+    /** @type {SearchInsideMatch} */
+    const match = this.match;
+    // TODO: Make this not use the global br instance
+    const pageIndex = window.br.leafNumToIndex(match.par[0].page);
```
iisa, line 42: hi @cdrini thanks for this! can you pipe it thru via the `searchProvider`? that layer receives `br` instance, and we won't have to call global var..

iisa, line 42: hi @cdrini thanks for this! can you pipe it thru via the `searchProvider`? that layer receives `br` instance, and we won't have to call global var..

cdrini: I think I found a good fix! Let me know what you think :)

src/plugins/search/plugin.search.js
```diff
@@ -269,22 +272,32 @@ BookReader.prototype.cancelSearchRequest = function () {
...
- * @param {boolean} options.goToFirstResult
+ * @param {(pageNum: LeafNum) => PageNumString} displayPageNumberFn
  */
-BookReader.prototype.BRSearchCallback = function(results, options) {
+export function marshallSearchResults(results, displayPageNumberFn) {
   // Attach matchIndex to a few things to make it easier to identify
   // an active/selected match
   for (const [index, match] of results.matches.entries()) {
     match.matchIndex = index;
+    match.displayPageNumber = displayPageNumberFn(match.par[0].page);
```
iisa, line 284: ok. I was hoping that the service would return it with the result sooner than later 😅

iisa (approved):
code looks good!
tested here: https://deploy-preview-1011--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=goody#page/10/mode/2up/search/goody

I tried searching 'the' and it took a reasonable amount of time to draw 👍 
 https://deploy-preview-1011--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala#page/430/mode/2up/search/the

weird that the Plato "the" search did not display results highlights properly. perhaps it's data bc `goody` search above displayed the highlights well.

cdrini:
Yeah :/ I think there's a JIRA ticket for that issue: https://webarchive.jira.com/browse/WEBDEV-5190

PR #1012, opened 2022-04-19
Lit 2 updates for <modal-manager>, <ia-activity-indicator>

iisa:


cdrini (approved):
Tested the spinning indicator works when making a search and that the modal manager creates a modal when trying to delete a bookmark with notes attached to it. lgtm!

PR #1043, opened 2022-04-28
Fetch is now ESM-only and won't work with require

cdrini:


scripts/preversion.js
```diff
@@ -1,8 +1,10 @@
 const { version: OLD_VERSION } = require('../package.json');
-const fetch = require('node-fetch');
 const OLD_RELEASE_URL = `https://api.github.com/repos/internetarchive/bookreader/releases/tags/v${OLD_VERSION}`;
 
 async function main() {
+    const {default: fetch} = await import('node-fetch');
+
+    console.log(fetch);
```
cdrini, line 7:
```suggestion
```

scripts/preversion.js
```diff
@@ -1,8 +1,9 @@
 const { version: OLD_VERSION } = require('../package.json');
-const fetch = require('node-fetch');
 const OLD_RELEASE_URL = `https://api.github.com/repos/internetarchive/bookreader/releases/tags/v${OLD_VERSION}`;
 
 async function main() {
+    const {default: fetch} = await import('node-fetch');
```
cdrini, line 5:
```suggestion
    // Need this because fetch is ESM-only, and we're on Node 16. Someday we should
    // be able to move this up to the top without renaming this file to a .mjs or whatever
    const {default: fetch} = await import('node-fetch');
```

cdrini:
Merging before CI ; none of our CI cover this file!

PR #1052, opened 2022-05-10
Run less babel for jest

cdrini:


cdrini:
Auto-merging; small change shouldn't affect prod code

PR #1054, opened 2022-05-12
Limit non-ia dep updates to monthly

cdrini:
This is getting a little noisy, and causing netlify to struggle with all the PRs. Limit to monthly for non-IA package updates.


cdrini:
Auto-merging; small devtools config change

iisa (commented):
LGTM!

PR #1056, opened 2022-05-20
add helper function, <optgroup> and sort function

### LINKED ISSUE #1055: Group ReadAloud languages by language
Currently the voice options is very long and seemingly ordered at random:
image

Use `<optgroup>` to display languages in the books language first and at the top. Then have another group for other languages, sorted by language code.

So for example the above would display:

image

**Proposal & Constraints**

**Additional Resources**
- Learn more about optgroup: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup

**Success Metrics**

> How will you know if this change is successful? What specific metrics are relevant and how do you expect them to change?

**Stakeholders**

> @-tag stakeholders of this feature

sancodes:
**Changes:** 
- add `renderVoiceOption()` which takes in `language` parameter, where two groups of language is passed, _english(en)_ and _non-english_
- `bookLanguages` and `otherLanguages` stores filtered out grouping of languages in a sorted ordered
- `voicesMenu` appends grouping of `english` and `non-english` languages where `<optgroup` is introduced

**Test and Checks** 
- Ran eslint and nothing seemed to break style wise

**Screenshot:** 
Screen Shot 2022-05-20 at 12 34 30 AM

**Fixes:** 
Fixes #1055

### DIFF
--- a/src/plugins/tts/plugin.tts.js
+++ b/src/plugins/tts/plugin.tts.js
@@ -172,13 +172,8 @@ BookReader.prototype.initNavbar = (function (super_) {
         const otherLanguages = this.ttsEngine.getVoices().filter(v => !v.lang.startsWith(bookLanguage)).sort(voiceSortOrder);
 
         if (this.ttsEngine.getVoices().length > 1) {
-          if (bookLanguages.length) {
-            voicesMenu.append($(`<optgroup label="Book Language (${bookLanguage})"> ${renderVoiceOption(bookLanguages)} </optgroup>`));
-          }
-
-          if (otherLanguages.length) {
-            voicesMenu.append($(`<optgroup label="Other Languages"> ${renderVoiceOption(otherLanguages)} </optgroup>`));
-          }
+          voicesMenu.append($(`<optgroup label="Book Language (${bookLanguage})"> ${renderVoiceOption(bookLanguages)} </optgroup>`));
+          voicesMenu.append($(`<optgroup label="Other Languages"> ${renderVoiceOption(otherLanguages)} </optgroup>`));
 
           voicesMenu.val(this.ttsEngine.voice.voiceURI);
           voicesMenu.show();
--- a/src/plugins/tts/plugin.tts.js
+++ b/src/plugins/tts/plugin.tts.js
@@ -163,19 +163,23 @@ BookReader.prototype.initNavbar = (function (super_) {
           `<option value="${voice.voiceURI}">${voice.lang} - ${voice.name}</option>`).join('');
       };
 
-      const sortedOrder = () => {
-        return (a,b) => `${a.lang} - ${a.name}`.localeCompare(`${b.lang} - ${b.name}`);
-      };
+      const voiceSortOrder = (a,b) => `${a.lang} - ${a.name}`.localeCompare(`${b.lang} - ${b.name}`);
 
       const renderVoicesMenu = (voicesMenu) => {
         voicesMenu.empty();
         const bookLanguage = this.ttsEngine.opts.bookLanguage;
-        const bookLanguages = this.ttsEngine.getVoices().filter(v => v.lang.startsWith(bookLanguage)).sort(sortedOrder());
-        const otherLanguages = this.ttsEngine.getVoices().filter(v => !v.lang.startsWith(bookLanguage)).sort(sortedOrder());
+        const bookLanguages = this.ttsEngine.getVoices().filter(v => v.lang.startsWith(bookLanguage)).sort(voiceSortOrder);
+        const otherLanguages = this.ttsEngine.getVoices().filter(v => !v.lang.startsWith(bookLanguage)).sort(voiceSortOrder);
 
         if (this.ttsEngine.getVoices().length > 1) {
-          voicesMenu.append($(`<optgroup label="Book Language (en)"> ${renderVoiceOption(bookLanguages)} </optgroup>`));
-          voicesMenu.append($(`<optgroup label="Other Languages"> ${renderVoiceOption(otherLanguages)} </optgroup>`));
+          if (bookLanguages.length > 1) {
+            voicesMenu.append($(`<optgroup label="Book Language (${bookLanguage})"> ${renderVoiceOption(bookLanguages)} </optgroup>`));
+          }
+
+          if (otherLanguages.length > 1) {
+            voicesMenu.append($(`<optgroup label="Other Languages"> ${renderVoiceOption(otherLanguages)} </optgroup>`));
+          }
+
           voicesMenu.val(this.ttsEngine.voice.voiceURI);
           voicesMenu.show();
         } else {
--- a/src/plugins/tts/plugin.tts.js
+++ b/src/plugins/tts/plugin.tts.js
@@ -158,27 +158,23 @@ BookReader.prototype.initNavbar = (function (super_) {
 
       $el.find('.BRcontrols').prepend(this.refs.$BRReadAloudToolbar);
 
-      const renderVoiceOption = (language) => {
-        return language.map(voice =>
-          `<option value="${voice.voiceURI}">${voice.lang} - ${voice.name}</option>`);
+      const renderVoiceOption = (voices) => {
+        return voices.map(voice =>
+          `<option value="${voice.voiceURI}">${voice.lang} - ${voice.name}</option>`).join('');
+      };
+
+      const sortedOrder = () => {
+        return (a,b) => a.lang.localeCompare(b.name);
       };
 
       const renderVoicesMenu = (voicesMenu) => {
         voicesMenu.empty();
-
-        const bookLanguages = this.ttsEngine.getVoices().filter(v => v.lang.startsWith('en')).sort((a,b) => {
-          if (a.lang > b.lang) return 1;
-          if (a.lang < b.lang) return -1;
-          return 0;
-        });
-        const otherLanguages = this.ttsEngine.getVoices().filter(v => !v.lang.startsWith('en')).sort((a, b) => {
-          if (a.lang > b.lang) return 1;
-          if (a.lang < b.lang) return -1;
-          return 0;
-        });
+        const bookLanguage = this.ttsEngine.opts.bookLanguage;
+        const bookLanguages = this.ttsEngine.getVoices().filter(v => v.lang.startsWith(bookLanguage)).sort(sortedOrder());
+        const otherLanguages = this.ttsEngine.getVoices().filter(v => !v.lang.startsWith(bookLanguage)).sort(sortedOrder());
 
         if (this.ttsEngine.getVoices().length > 1) {
-          voicesMenu.append($(`<optgroup label="English (Book Language)"> ${renderVoiceOption(bookLanguages)} </optgroup>`));
+          voicesMenu.append($(`<optgroup label="Book Language (en)"> ${renderVoiceOption(bookLanguages)} </optgroup>`));
           voicesMenu.append($(`<optgroup label="Other Languages"> ${renderVoiceOption(otherLanguages)} </optgroup>`));
           voicesMenu.val(this.ttsEngine.voice.voiceURI);
           voicesMenu.show();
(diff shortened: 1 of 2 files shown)


sancodes:
@cdrini Can you help me review my changes. I am willing to fix any mistakes. Thank you. 

Wouldn't it be great to sort the English languages as well? 
In the issues task, it just mentioned to order the non-english languages. 
For now, I sorted both grouping, UI wise, I think it would look clean, (just my thinking).

src/plugins/tts/plugin.tts.js
```diff
@@ -158,18 +158,35 @@ BookReader.prototype.initNavbar = (function (super_) {
...
+      };
+
       const renderVoicesMenu = (voicesMenu) => {
         voicesMenu.empty();
+
+        const bookLanguages = this.ttsEngine.getVoices().filter(v => v.lang.startsWith('en')).sort((a,b) => {
+          if (a.lang > b.lang) return 1;
+          if (a.lang < b.lang) return -1;
+          return 0;
+        });
```
cdrini, line 173: Put the sorting function in a helper, so that we can avoid the duplicate code. You can also simplify this by using `localeCompare` to compare the strings: `(a, b) => a.lang.localeCompare(b.lang)`. And +1 sorting both looks much better!

Let's also sort by the voice name! We can do that by sorting over `${voice.lang} - ${voice.name}` instead of just `voice.lang`.

src/plugins/tts/plugin.tts.js
```diff
@@ -158,18 +158,35 @@ BookReader.prototype.initNavbar = (function (super_) {
...
+
         if (this.ttsEngine.getVoices().length > 1) {
-          voicesMenu.append(this.ttsEngine.getVoices().map(
-            voice =>
-              $(`<option value="${voice.voiceURI}">${voice.lang} - ${voice.name}</option>`)));
+          voicesMenu.append($(`<optgroup label="English (Book Language)"> ${renderVoiceOption(bookLanguages)} </optgroup>`));
```
cdrini, line 181: Let's switch this to be like so ; that way we don't have to convert `en -> English`, which is actually kind of hard in the general case!
```suggestion
          voicesMenu.append($(`<optgroup label="Book Language (en)"> ${renderVoiceOption(bookLanguages)} </optgroup>`));
```

src/plugins/tts/plugin.tts.js
```diff
@@ -158,18 +158,35 @@ BookReader.prototype.initNavbar = (function (super_) {
 
       $el.find('.BRcontrols').prepend(this.refs.$BRReadAloudToolbar);
 
+      const renderVoiceOption = (language) => {
```
cdrini, line 161: Since this takes multiple voices as input, lets make this plural, and change the argument name here.
```suggestion
      const renderVoicesOptions = (voices) => {
```

src/plugins/tts/plugin.tts.js
```diff
@@ -158,18 +158,35 @@ BookReader.prototype.initNavbar = (function (super_) {
...
+        return language.map(voice =>
+          `<option value="${voice.voiceURI}">${voice.lang} - ${voice.name}</option>`);
```
cdrini, line 163: We want this to return a plain string; using just `map` returns an array of strings!

```suggestion
          `<option value="${voice.voiceURI}">${voice.lang} - ${voice.name}</option>`)
          .join('');
```

cdrini (changes requested):
This looks great @sancodes ! Some feedback for you :+1:

Now that you've got everything working for `'en'`, replace everywhere you explicitly write `'en'` with `this.ttsEngine.opts.bookLanguage` (or maybe store that in a variable since the name is so long :P ). That'll handle the general case!

sancodes:
@cdrini will do. Thank you

src/plugins/tts/plugin.tts.js
sancodes: about the sorting part: 
- I made a sorting helper, which sorts normally
- However, little confused as to how to sort here, I did sort like so, `voice.lang` - `voice.name`


```
const sortedOrder = () => {
        return (a,b) => a.lang.localeCompare(b.name);
      };
```
Screen Shot 2022-05-20 at 2 35 49 PM

Result comes out as so.

cdrini: Ah, I meant sort the string:

```
`${a.lang} - ${a.name}`.localeCompare(`${b.lang} - ${b.name}`)
```

cdrini: This will let the sort take into account both the language code, locale, and name :+1:

sancodes: @cdrini  

doing so didn't really change the result. The voice option looks similar to the first PR picture I did. 
I'll do a PR, so you can see, what I may have done wrong.  :)

cdrini: Hmm, it's working correctly for me now after the changes; can you send a screenshot?

sancodes: Screen Shot 2022-05-20 at 3 56 35 PM

This is same ordering as the first PR.

cdrini: No that looks correct! Note how `en-US - Google US English` is now before `Samantha` :)

sancodes: @ ooops, really missed that part of the order. 
Guess I was staring at those options for long. I couldnt catch it haha.

sancodes: @cdrini so looks good to merge?

cdrini: Oh you'll need to make the changes in the other comments :) Those are just code suggestions; I haven't actually modified the code! Then I'll do some cross browser testing, and we should be good!

cdrini: Oh whoops you already have! My apologies I think I was looking at an old diff. Let me do another pass!

src/plugins/tts/plugin.tts.js
```diff
@@ -158,18 +158,31 @@ BookReader.prototype.initNavbar = (function (super_) {
...
       $el.find('.BRcontrols').prepend(this.refs.$BRReadAloudToolbar);
 
+      const renderVoiceOption = (voices) => {
+        return voices.map(voice =>
+          `<option value="${voice.voiceURI}">${voice.lang} - ${voice.name}</option>`).join('');
+      };
+
+      const sortedOrder = () => {
+        return (a,b) => `${a.lang} - ${a.name}`.localeCompare(`${b.lang} - ${b.name}`);
+      };
```
cdrini, line 168: We can remove the outer function here, and then do `.sort(voiceSortOrder)`

```suggestion
      const voiceSortOrder = (a,b) => `${a.lang} - ${a.name}`.localeCompare(`${b.lang} - ${b.name}`);
```

src/plugins/tts/plugin.tts.js
```diff
@@ -158,18 +158,31 @@ BookReader.prototype.initNavbar = (function (super_) {
...
         voicesMenu.empty();
+        const bookLanguage = this.ttsEngine.opts.bookLanguage;
+        const bookLanguages = this.ttsEngine.getVoices().filter(v => v.lang.startsWith(bookLanguage)).sort(sortedOrder());
+        const otherLanguages = this.ttsEngine.getVoices().filter(v => !v.lang.startsWith(bookLanguage)).sort(sortedOrder());
+
         if (this.ttsEngine.getVoices().length > 1) {
-          voicesMenu.append(this.ttsEngine.getVoices().map(
-            voice =>
-              $(`<option value="${voice.voiceURI}">${voice.lang} - ${voice.name}</option>`)));
+          voicesMenu.append($(`<optgroup label="Book Language (en)"> ${renderVoiceOption(bookLanguages)} </optgroup>`));
```
cdrini, line 177: Replace `en` here

src/plugins/tts/plugin.tts.js
sancodes: yea, done with all the changes

src/plugins/tts/plugin.tts.js
cdrini, line 177: We will want the `if` to now be:

```js
if (bookLanguages.length > 1) { ... }
if (otherLanguages.length > 1) { ... }
```

to avoid rendering empty groups

cdrini (changes requested):
Nice! Everything looks good :) 

I'm noticing some issues with colours on Firefox:

image

We have css on `select.playback-voices option` that sets the background to `#333`; can you also add `optgroup` to that css?

And chrome is showing these centered, which looks odd :P Can you remove the `text-align: center`?

image

Then should be good to merge!

sancodes:
@cdrini I did work on that CSS file as you mentioned, but not so sure, if I did it correctly. 
Also, made changes to other files, as you mentioned. 

Help me double check, I will fix mistakes if any.

src/plugins/tts/plugin.tts.js
```diff
@@ -158,18 +158,35 @@ BookReader.prototype.initNavbar = (function (super_) {
...
+
         if (this.ttsEngine.getVoices().length > 1) {
-          voicesMenu.append(this.ttsEngine.getVoices().map(
-            voice =>
-              $(`<option value="${voice.voiceURI}">${voice.lang} - ${voice.name}</option>`)));
+          if (bookLanguages.length > 1) {
+            voicesMenu.append($(`<optgroup label="Book Language (${bookLanguage})"> ${renderVoiceOption(bookLanguages)} </optgroup>`));
+          }
+
+          if (otherLanguages.length > 1) {
```
cdrini, line 179: Oh whoops, my bad this just be `> 0`!
```suggestion
          if (bookLanguages.length) {
            voicesMenu.append($(`<optgroup label="Book Language (${bookLanguage})"> ${renderVoiceOption(bookLanguages)} </optgroup>`));
          }

          if (otherLanguages.length) {
```

cdrini (commented):
Hmm, I'm not seeing the css changes. Make sure to edit the `scss` file somewhere in the `src` directory!

cdrini:
Ah here's the section: https://github.com/internetarchive/bookreader/blob/956489cc92b6ea37908dc36107bf837773097c0a/src/css/_controls.scss#L42-L44

sancodes:
@cdrini I thought the css changes that needed to made was in a different file as it had the line
`select.playback-voices option` that sets the background to `#333`

https://github.com/internetarchive/bookreader/blob/5c14e7de8567f739dcf0cc1e04c6c23a0c79272b/BookReader/BookReader.css#L2709

so I was trying to do something like this: 
Screen Shot 2022-05-20 at 6 27 00 PM

cdrini:
The .css files are generated from the .scss files. Sorry, it is a little confusing since they show up when searching the code!

sancodes:
> The .css files are generated from the .scss files. Sorry, it is a little confusing since they show up when searching the code!

Oh, I wasn't aware of it, now I know. 
So tackling the scss file is the way to go right?

cdrini:
Yes exactly 👍 I'll be offline over the weekend, but will check in on my Monday. Have a good weekend!

sancodes:
@cdrini modified SCSS file. Can you help me verify my work, when you get time. Thank you.

src/plugins/tts/plugin.tts.js
```diff
@@ -158,18 +158,35 @@ BookReader.prototype.initNavbar = (function (super_) {
...
-          voicesMenu.append(this.ttsEngine.getVoices().map(
-            voice =>
-              $(`<option value="${voice.voiceURI}">${voice.lang} - ${voice.name}</option>`)));
+          if (bookLanguages.length) {
+            voicesMenu.append($(`<optgroup label="Book Language (${bookLanguage})"> ${renderVoiceOption(bookLanguages)} </optgroup>`));
+          }
+
+          if (otherLanguages.length) {
+            voicesMenu.append($(`<optgroup label="Other Languages"> ${renderVoiceOption(otherLanguages)} </optgroup>`));
+          }
```
cdrini, line 181: Actually, this provides useful context when no valid languages are available! Leaving in.

```suggestion
          voicesMenu.append($(`<optgroup label="Book Language (${bookLanguage})"> ${renderVoiceOption(bookLanguages)} </optgroup>`));
          voicesMenu.append($(`<optgroup label="Other Languages"> ${renderVoiceOption(otherLanguages)} </optgroup>`));
```

cdrini:
Merged! Great work, @sancodes !

sancodes:
@cdrini if there's another task, please let me know, I am willing to tackle it.

PR #1061, opened 2022-06-01
Allow for LCP download links in side panel

iisa:
- add link option for LCP in downloads section
- add helpful text section to download SimplyE, Aldiko Next, Thorium
- add in IA demo a toggle to show LCP download links

To test: https://deploy-preview-1061--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala
- once review app is loaded, open side menu, then click "LCP Download Option" -> options should update with the helper text showing

image


src/BookNavigator/downloads/downloads.js
```diff
@@ -40,6 +40,16 @@ export class IABookDownloads extends LitElement {
     ));
   }
 
+  /**
+   * checks if downloads list contains an LCP option
+   * @return {boolean}
+   */
+  get hasLCPOption() {
+    const regex = /^(LCP)/g;
+    const lcpAvailable = this.downloads.some(option => option.type?.match(regex));
```
nsharma123, line 49: So you are assuming that LCP will be available in both version always. Any change LCP available for PDF version only?

src/BookNavigator/downloads/downloads.js
```diff
@@ -59,12 +69,24 @@ export class IABookDownloads extends LitElement {
     `;
   }
 
+  get installSimplyEAldikoThoriumMsg() {
+    return html`
+    <p>For LCP downloads, make sure you have SimplyE or Aldiko Next installed on mobile or Thorium on desktop.</p>
+    <ul>
+      <li><a href="https://librarysimplified.org/simplye/" rel="noopener noreferrer" target="_blank">Install SimplyE</a></li>
+      <li><a href="https://www.demarque.com/en-aldiko" rel="noopener noreferrer" target="_blank">Install Aldiko</a></li>
+      <li><a href="https://www.edrlab.org/software/thorium-reader/" rel="noopener noreferrer" target="_blank">Install Thorium</a></li>
```
nsharma123, line 78: Did you thought about `nofollow` as well inside rel attribute?

src/BookNavigator/downloads/downloads.js
```diff
@@ -59,12 +69,24 @@ export class IABookDownloads extends LitElement {
...
+  }
+
   render() {
     return html`
       ${this.header}
       ${this.loanExpiryMessage}
       <ul>${this.renderDownloadOptions()}</ul>
-      ${this.isBookProtected ? this.accessProtectedBook : nothing}
+      ${this.hasLCPOption ? this.installSimplyEAldikoThoriumMsg : nothing}
+      ${this.isBookProtected && !this.hasLCPOption ? this.accessProtectedBook : nothing}
```
bfalling, line 89: TINY SUGGESTION: I think you could slightly simplify the logic by nesting the conditions, e.g.
```
${this.hasLCPOption
  ? this.installSimplyEAldikoThoriumMsg
  : (this.isBookProtected ? this.accessProtectedBook : nothing)
}
```
Not sure it's an improvement, but see what you think! ;)

bfalling (approved):
One tiny suggestion. Otherwise, LGTM!

src/BookNavigator/downloads/downloads.js
iisa: thanks for the catch. updated.

src/BookNavigator/downloads/downloads.js
iisa: > Any change LCP available for PDF or EPUB version only? 

unsure, but they can use any of the new option types to add a link so we should be covered. `option.type: 'LCP ePub' | 'LCP PDF'`

src/BookNavigator/downloads/downloads.js
iisa: hi, very good suggestion! updated.

PR #1073, opened 2022-08-02
keep  url param dynamic with patron's search request

iisa:
This PR removes `q=<term>` query param when search is cancelled or when search menu input is cleared

### DIFF
--- a/src/BookNavigator/search/search-provider.js
+++ b/src/BookNavigator/search/search-provider.js
@@ -75,6 +75,10 @@ export default class SearchProvider {
       searchCanceled: true
     };
     this.updateMenu(updateMenuFor);
+
+    if (this.bookreader.urlPlugin) {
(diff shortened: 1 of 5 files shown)


src/BookNavigator/search/search-provider.js
```diff
@@ -146,6 +150,21 @@ export default class SearchProvider {
     };
     this.updateMenu({ openMenu: false });
     this.bookreader?.searchView?.clearSearchFieldAndResults(false);
+    if (this.bookreader.urlPlugin) {
+      this.updateSearchInUrl();
+    }
+  }
+
+  /** update URL `q=<term>` query param in URL */
+  updateSearchInUrl() {
```
dualcnhq, line 159: happy to see URLPlugin is being used here 👍 😊

dualcnhq (approved):

Issue #173, opened 2019-10-29
Page flip direction off after toggling different view modes


shaneriley:
**Description**

When switching to 1up, toggling the menu, switching to thumbnails, clicking a page, switching back to thumbnails, toggling menu, then switching to 2up, the next forward page flip is not animated correctly.

**Evidence / Screenshot (if possible)**

https://drive.google.com/open?id=10GkJEL04jM1szcl_1lfGoneOC_K0GR7l

**Expectation**

Page should flip as in normal 2up mode

**Context**

Chrome 77, Mac OS

Sohan022:
I have not accessed to open google drive link which is mention in the evidence.

shaneriley:
Sorry. Here's a new link with changed permissions. https://drive.google.com/file/d/10GkJEL04jM1szcl_1lfGoneOC_K0GR7l/view?usp=sharing

PR #1077, opened 2022-08-10
Fix leaf positioning during rtl flip anim

latonv:
This issue affects right-to-left page flipping via the "next page" button in 2up mode. It mainly appears to affect texts with an even number of pages where the first and last pages present on their own.

If the new left page remains left-positioned upon flipping to the next page, its animation unfolds in the wrong direction. Ensuring before the animation that it is right-positioned at the center prevents this issue.

This PR should also address https://github.com/internetarchive/bookreader/issues/173 which appears to have the same root cause.

### Problem gif (left pages flip the wrong way following a wraparound from the end to the start of the book):
bookreader-page-flip-anim-problem

### Corrected gif (pages always flip the right way):
bookreader-page-flip-anim-corrected

### Testing
Preview app: https://deploy-preview-1077--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=CIA-RDP05S00620R000300830010-7

- Flip through all the pages of the given text until the end (e.g., with the right arrow key, clicking 'Next Page' button, or tapping/swiping right).
- From there, moving to the next page will wraparound to the beginning of the text.
- Continue moving to the next page repeatedly and verify that the page flip animations look correct -- after the right leaf has animated into the center, the left leaf should grow outward from the center (not inward from the left edge, as in the problem gif).
- Moving back a page (flipping left-to-right) at any point should also continue to look as expected, and not interfere with the forward movement animations after it.

### DIFF
--- a/tests/jest/BookReader/Mode2Up.test.js
+++ b/tests/jest/BookReader/Mode2Up.test.js
@@ -54,7 +54,7 @@ describe('page flip directions', () => {
 
     const fake = sinon.fake();
     const fakeAnimWithCB = sinon.fake.yields();
-    const fakeAnim = sinon.fake((...args) => 
+    const fakeAnim = sinon.fake((...args) =>
       typeof args[args.length - 1] === 'function' ? fakeAnimWithCB(...args) : fake
     );
     sinon.replace(jQuery.prototype, 'animate', fakeAnim);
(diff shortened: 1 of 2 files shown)


tests/jest/BookReader/Mode2Up.test.js
```diff
@@ -47,6 +47,29 @@ describe('zoom', () => {
...
+    sinon.replace(jQuery.prototype, 'animate', fakeAnim);
+
+    const fakeCSS = sinon.spy(jQuery.prototype, 'css');
+
+    br.next();
+
+    expect(fakeAnimWithCB.callCount).toBe(2);
+    // Find the call to .css() immediately preceding the second animation with a callback (i.e., the left page animation)
+    const preSecondAnimCssCallIndex = fakeCSS.getCalls().findIndex(call => call.calledAfter(fakeAnimWithCB.getCall(1))) - 1;
+    expect(fakeCSS.getCall(preSecondAnimCssCallIndex).args[0].left).toBe('');
```
cdrini, line 69: Oh that's a cool way to test this; nice!

cdrini (approved):
Lgtm! Tested on netlify that the bug used to occur and that it no longer occurs.

PR #1088, opened 2022-09-30
WEBDEV-5491 Ensure search results are sanitized

latonv:
**Problem:** Search results are rendered using the `unsafeHTML` directive, which allows HTML tags present in the underlying document to be injected into the DOM when they appear in search results. This creates an XSS vulnerability when, e.g., the underlying document contains `<script>` tags that will be executed upon rendering the search results.

This PR ensures that search results are inserted into the DOM as raw text, with the only active markup present being the `<mark>` tags added to highlight the search keywords.

### DIFF
--- a/src/BookNavigator/search/a-search-result.js
+++ b/src/BookNavigator/search/a-search-result.js
@@ -1,4 +1,4 @@
-import { escapeHTML } from '../../BookReader/utils';
+import { escapeHTML } from '../../BookReader/utils.js';
 import { html, LitElement, nothing } from 'lit';
 import { unsafeHTML } from 'lit/directives/unsafe-html.js';
 /** @typedef {import('@/src/plugins/search/plugin.search.js').SearchInsideMatch} SearchInsideMatch */
--- a/src/BookNavigator/search/a-search-result.js
+++ b/src/BookNavigator/search/a-search-result.js
@@ -1,4 +1,6 @@
+import { escapeHTML } from '../../BookReader/utils';
 import { html, LitElement, nothing } from 'lit';
+import { unsafeHTML } from 'lit/directives/unsafe-html.js';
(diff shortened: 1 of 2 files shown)


src/BookNavigator/search/a-search-result.js
```diff
@@ -12,17 +11,44 @@ export class BookSearchResult extends LitElement {
...
+        `);
+        index = match.index + match[0].length;
+      }
+    }
+
+    // Include any text from the last match to the end
+    templates.push(html`${hit.slice(index)}`);
+
+    // Squash everything into a single p template
+    return html`<p>${templates}</p>`;
```
cdrini, line 51: A simpler approach might be something like:

```js
return html`<p>${unsafeHTML(escapeHTML(hit).replace(/{{{/g, '<mark>').replace(/}}}/g, '</mark>'))}</p>`
```

escapeHTML exists in our utils :)

tests/jest/BookNavigator/search/search-results.test.js
```diff
@@ -46,6 +46,25 @@ const results = [{
   }],
 }];
 
+const resultWithScript = [{
```
cdrini, line 49: Woohoo! Thanks for adding a test to prevent this coming back in the future!

src/BookNavigator/search/a-search-result.js
latonv: Ah, figures we have an escaper for this already! Thanks for the suggestion!

cdrini (approved):
One small change; I can't recall if importing from pb might be an issue without the .js extension... CC @iisa Otherwise lgtm! Tested locally and fixed the bug with the dev environment failing for some URLs (pushing up in a separate PR)

PR #1090, opened 2022-09-30
Fix dev environment failing to load some books

cdrini:
Remove BRJSLocate call and hit BookReaderJSIA.php directly. It seems like JSLocate was creating the wrong manifest URL!


iisa (approved):

cdrini:
Lightning fast @iisa !!! 🚀:P

PR #1091, opened 2022-09-30
Escape search results being shown in the footer bar

latonv:
Follow-up to #1088 -- unescaped search results appear in the footer bar as well. This PR ensures they are escaped before being inserted into the DOM.

**Testing:** Visit https://deploy-preview-1091--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=book-reader-test-1234#mode/1up/search/searching and verify that no browser alert occurs (initially or upon performing different searches).


cdrini (approved):
Lgtm! Sigh we should DRY this at some point :P How did I miss the alert firing on the other PR...

Tested against master and confirmed no longer firing!

PR #1098, opened 2022-10-19
get identifier for subfiles as well

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-4812

Description: earlier we were getting identifier using `this.br.bookId` but it fails when item has sub-files. so we need to consider `identifier+sub-filename` as complete identifier and store bookmarks for identifier.


iisa (approved):
👍

PR #1099, opened 2022-10-25
Remove most class shims + v5 BR deprecations

cdrini:
These were added as part of the v5 transition to avoid breaking changes. Let's get rid of as many as possible.

A few are left because they are used in petabox, or because they were overriden in plugins.

### DIFF
--- a/src/plugins/plugin.autoplay.js
+++ b/src/plugins/plugin.autoplay.js
@@ -81,7 +81,7 @@ BookReader.prototype.autoToggle = function(overrides) {
   }
 
   // Change to autofit if book is too large
-  if (this.reduce < this.twoPageGetAutofitReduce()) {
+  if (this.reduce < this._modes.mode2Up.getAutofitReduce()) {
     this.zoom('auto');
   }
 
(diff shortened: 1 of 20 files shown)


src/plugins/plugin.autoplay.js
```diff
@@ -103,9 +103,9 @@ BookReader.prototype.autoToggle = function(overrides) {
       if (this.animating) return;
 
       if (Math.max(this.twoPage.currentIndexL, this.twoPage.currentIndexR) >= this.lastDisplayableIndex()) {
-        this.flipBackToIndex(1); // $$$ really what we want?
+        this.prev({ triggerStop: false }); // $$$ really what we want?
```
iisa, line 106: what's `triggerStop`?

iisa (approved):
code looks good!
I see that we are removing top level mode helpers and directly calling `this._modes[modeDuJour]` 👍

src/plugins/plugin.autoplay.js
cdrini: Darn I was trying to sneak this in :P Probably something we should clean up later, cause it's a little weird. I added a new param whether it should fire the "Stop" event, because previous, flipBackToIndex wouldn't fire, but I deleted that public method. So I added a flag here.

cdrini:
I'm not noticing any delay on my iPad!

PR #1104, opened 2022-11-11
Default BookReader to 1up instead of 2up on mobile

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-5253


src/BookReader.js
```diff
@@ -498,6 +498,12 @@ BookReader.prototype.getInitialMode = function(params) {
   if (!this.canSwitchToMode(nextMode)) {
     nextMode = this.constMode1up;
   }
+
+  // override defaults mode via `bookreader-defaults` metadata
+  if (null != this.options.defaults) {
+    nextMode = this.options.defaults === 'mode/1up' ? this.constMode1up : this.constMode2up;
```
iisa, line 504: :nice:

iisa (approved):
@nsharma123 code looks good but we cannot merge the build files into main as they get generated by release flow. can u remove those before merge? thanks!

iisa:
@nsharma123 please remove build files before merge. thanks!

iisa:
tests also need to be checked, they are failing

nsharma123:
yes, I will remove those and fix the tests as well. Thanks!

src/BookReader.js
```diff
@@ -473,6 +475,12 @@ BookReader.prototype.getInitialMode = function(params) {
   if (!this.canSwitchToMode(nextMode)) {
     nextMode = this.constMode1up;
   }
+
+  // override defaults mode via `bookreader-defaults` metadata
+  if (null != this.options.defaults) {
```
iisa, line 480: this conditional is "[yoda style](https://eslint.org/docs/latest/rules/yoda)". we don't usually use this style in BR.

can a falsey work? `if (!this.options.defaults)` ?

src/BookReader.js
```diff
@@ -473,6 +475,12 @@ BookReader.prototype.getInitialMode = function(params) {
...
+    nextMode = this.options.defaults === 'mode/1up' ? this.constMode1up : this.constMode2up;
```
iisa, line 481: Suggestion, potential blocker: we have possible 3rd option for thumbnail view - can you add support here for it? @nsharma123

iisa (changes requested):
@nsharma123 thanks for the fix! got a couple of suggestions, I just want us to ensure we cover thumbnail mode when it is set to default. thank you!

iisa (approved):
thanks for making sure all modes are covered

iisa:
looks good once tests pass

PR #1105, opened 2022-11-14
ia analytics, move to 1% sampled bucket

iisa:
Calls transferred to 1% sample bucket:
- on load, signal when book is restricted
- volume sorting when viewing a multi-volume book


iisa:
https://deploy-preview-1105--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala#page/24/mode/2up looks good and snappy on Chrome, but safari feels a bit slow toggling full screen. can u repro on your iPad @cdrini ?

dualcnhq (approved):

PR #1109, opened 2022-12-06
WEBDEV-5660 Replace dotAll (`s`) flag in search result regexes with better-supported alternative

latonv:
BookReader’s search result components use regexes to match `{{{}}}`-delimited keywords for conversion into highlighted text. Since these keyword matches can (rarely) include line breaks, the regexes have the `s` (“dotAll”) flag set that allows `.` to match line breaks. However, the `s` flag’s browser support only goes back a few years, and older browsers panic out entirely when they encounter it.

A viable alternative (that even _much_ older browsers support) is simply to use `[^]` (an empty negated character class) in place of `.` to match any character, including line breaks. This removes any need for the `s` flag.

### DIFF
--- a/src/BookNavigator/search/a-search-result.js
+++ b/src/BookNavigator/search/a-search-result.js
@@ -13,7 +13,7 @@ export class BookSearchResult extends LitElement {
   constructor() {
     super();
 
-    this.matchRegex = new RegExp('{{{(.+?)}}}', 'gs');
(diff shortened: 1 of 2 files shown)


iisa (approved):

iisa:
thanks for the fix @latonv. I see in this PR that this code has unit test coverage https://github.com/internetarchive/bookreader/pull/1091/files

PR #1112, opened 2022-12-16
Fix default iOS voice missing from menu

cdrini:
iOS 16 seems to somehow not show the default OS voice in the menu (Usually Samantha or Samantha Enhanced). So include an options specifically on these devices for the system default. This sets the voice to `null` under the hood when the speech utterance is made.

### DIFF
--- a/tests/jest/plugins/tts/WebTTSEngine.test.js
+++ b/tests/jest/plugins/tts/WebTTSEngine.test.js
@@ -1,5 +1,5 @@
 import sinon from 'sinon';
-import { WebTTSSound } from '@/src/plugins/tts/WebTTSEngine.js';
+import WebTTSEngine, { WebTTSSound } from '@/src/plugins/tts/WebTTSEngine.js';
 import { afterEventLoop, eventTargetMixin } from '../../utils.js';
 
 beforeEach(() => {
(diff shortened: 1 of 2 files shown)


latonv (approved):
LGTM! So weird that iOS neglects to surface its own default voice.

tests/jest/plugins/tts/WebTTSEngine.test.js
```diff
@@ -21,6 +22,51 @@ afterEach(() => {
...
+      },
+      {
+        default: true,
+        lang: "bg-BG",
+        localService: true,
+        name: "Daria",
+        voiceURI: "com.apple.voice.compact.bg-BG.Daria",
+      }
+    ];
+    const voices = WebTTSEngine.prototype.getVoices(null);
```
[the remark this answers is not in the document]
cdrini: Oh whoops! You're right, no need!

cdrini, line 44:
```suggestion
    const voices = WebTTSEngine.prototype.getVoices();
```

tests/jest/plugins/tts/WebTTSEngine.test.js
```diff
@@ -21,6 +22,51 @@ afterEach(() => {
...
+      },
+      {
+        default: false,
+        lang: "bg-BG",
+        localService: true,
+        name: "Daria",
+        voiceURI: "com.apple.voice.compact.bg-BG.Daria",
+      }
+    ];
+    const voices = WebTTSEngine.prototype.getVoices(null);
```
cdrini, line 66:
```suggestion
    const voices = WebTTSEngine.prototype.getVoices();
```

Issue #1129, opened 2023-01-16
Text selection broken on Safari due to browser bug with svg


cdrini:
See https://bugs.webkit.org/show_bug.cgi?id=242861

PR #1142, opened 2023-01-31
Add protected option to text selection plugin

### LINKED ISSUE #1140: Implement tweaks to text selection prototype

cdrini:
Closes #1140

Issues:

- Not allowing selection at all on FF Android
- Difficult to test on safari because of #1129


cdrini:
Auto-merging; more experimental feature, but will be useful to have while testing.

PR #1144, opened 2023-02-01
Dispatch custom event from bookreader for ia-book-actions

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-5711

Description: ia-book-actions uses new feature call `loan-auto-renew` where we need to send a specific pageChanged event when user clicked on new page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@internetarchive/bookreader",
-  "version": "5.0.0-53-alpha2",
+  "version": "5.0.0-53",
   "description": "The Internet Archive BookReader.",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@internetarchive/bookreader",
-  "version": "5.0.0-53",
+  "version": "5.0.0-53-alpha2",
   "description": "The Internet Archive BookReader.",
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


src/BookReader.js
```diff
@@ -1279,7 +1279,8 @@ BookReader.prototype.updateFirstIndex = function(
   if (this.options.initialSearchTerm && !suppressFragmentChange) {
     this.suppressFragmentChange = false;
   }
-  this.trigger('pageChanged');
+
+  this.trigger(BookReader.eventNames.pageChanged, { hasPageChanged: true });
```
iisa, line 1283: Q: when page changes - `hasPagedChanged` is always true. feels like we do not need that custom value.

nsharma123: `BookReader.eventNames.pageChanged` actually being called from multiple places and even execute when you reload the page as well. so I need to catch specific event when user actually Click on any book page or change the page using br controls. so I found this function is appropriate from the actual fragementChange event does when user interact.

iisa: resolving to re-think abstraction

iisa (approved):
looks good. linting needs update - missing `;`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@internetarchive/bookreader",
-  "version": "5.0.0-53",
+  "version": "5.0.0-53-alpha2",
```
iisa, line 3: blocker: please remove before merge

PR #1146, opened 2023-02-21
1143/bugfix/2up in fullscreen

### LINKED ISSUE #1143: BookReader always 1up when opening into fullscreen mode
**Description**

When BookReader opens in fullscreen mode (for instance when following a link from OpenLibrary) it seems to always open in 1up, even if the screen width is greater than the 800px cutoff.  Opening books from IA doesn't exhibit this behavior, because they don't start in fullscreen.

Open any book through the OpenLibrary borrow path (e.g. https://openlibrary.org/borrow/ia/dragonravenorday00hent?ref=ol) in a window wider than 800px.

**Expectation**

Books opening in larger browser windows should default to 2up.

**Context**

Verified in newest Firefox and Safari on MacOS and native browsers on iOS.

**Stakeholders**

@cdrini @mheiman

mheiman:
Closes #1143

Commit 19931be introduced a logic error in getInitialMode() that caused the mode to always be set to 1up if fullscreen mode was active, ignoring the mobile breakpoint.

Based on the comments in the new code and PR, the intention appears to have been to default to 1up whenever the screen size was within the mobile breakpoint (regardless of screen mode), but that's not what was happening. 

This PR removes the screen mode check, allowing the behavior intended by the previous PR. It also restores a comment that had gotten shifted from its intended location by a later edit.

### DIFF
--- a/src/BookReader.js
+++ b/src/BookReader.js
@@ -465,8 +465,8 @@ BookReader.prototype.getInitialMode = function(params) {
   };
   if ('undefined' != typeof(params.mode)) {
     nextMode = params.mode;
-  } else if ((this.ui == 'full' && this.isFullscreenActive) || ifMobileBreakpoint()) {
-    // In full mode OR device width, we set the default based on width
+  } else if (this.ui == 'full' && this.isFullscreenActive && ifMobileBreakpoint()) {
+    // In full mode, we set the default based on width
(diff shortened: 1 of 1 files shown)


cdrini (approved):
Lgtm! Wasn't able to fully test, but the logic looks solid, and low risk. Thanks @mheiman !

PR #1158, opened 2023-03-13
Update some ia packages

cdrini:
Update to latest to avoid duplicate inserts on prod.


cdrini:
Auto-merging since it's a small, dep update. I tested that the loading indicator works in the search panel, and that the delete modal fires when deleting a bookmark.

PR #1162, opened 2023-03-15
Fix incorrect filename in _icons.scss

### LINKED ISSUE #1160: _icons.scss has a reference with a misspelled filename which breaks webpack builds
The `bookreader/BookReader/icons/fullscreen_exit.svg` file is [misspelled here](https://github.com/internetarchive/bookreader/blob/30bcb573f0c85082997729b71eff6acf2ed001a1/src/css/_icons.scss#L40).

When importing the IA bookreader as an npm package, local webpack builds break with errors about not being able to resolve the file.

cdrini:
Closes #1160

### DIFF
--- a/src/css/_icons.scss
+++ b/src/css/_icons.scss
@@ -37,7 +37,7 @@
 }
 
 .icon-fullscreen-exit {
-  background-image: url("icons/fullscreen-exit.svg");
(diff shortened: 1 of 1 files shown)


cdrini:
Auto-merging since it is a very small change in a likely unused icon file.

PR #1171, opened 2023-03-27
Fix DOMException in sandboxed iframe in chrome

### LINKED ISSUE #1170: BookReader should work from a sandboxed iframe
- [x] Wrap cookie get's, since document.cookie.get throws errors in a sandboxed iframe
    - [x] #1161
- [ ] Fix bug in Webkit/chrome, causing history.replaceState to error in sandboxed mode

cdrini:
Closes #1170 

Happens when calling history.replaceState


iisa (approved):
👍 ah, we do it 2x bc there are 2 functions. ok. 🙂

PR #1173, opened 2023-03-28
Search inside plugin listen to config

iisa:
Now, search plugin instantiates by looking at `enableSearch` boolean flag inside Options config
+ demo
+ unit test for plugin
+ unit test for side panel

## Testing https://deploy-preview-1173--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala#mode/1up
- go to review app
- scroll down to "Remove Search Panel" -> click once
- ✅ book reloads with NO search shortcut & NO search side panel

### DIFF
--- a/src/BookNavigator/book-navigator.js
+++ b/src/BookNavigator/book-navigator.js
@@ -237,7 +237,11 @@ export class BookNavigator extends LitElement {
     }
 
     this.menuProviders = providers;
-    this.addMenuShortcut('search');
+    if (this.bookreader.options.enableSearch) {
+      this.addMenuShortcut('search');
+    } else {
+      this.removeMenuShortcut('search');
+    }
     this.addMenuShortcut('volumes');
     this.updateMenuContents();
   }
--- a/src/plugins/search/plugin.search.js
+++ b/src/plugins/search/plugin.search.js
@@ -45,6 +45,13 @@ jQuery.extend(BookReader.defaultOptions, {
 /** @override */
 BookReader.prototype.setup = (function (super_) {
   return function (options) {
+    // override with config
+    if (options.enableSearch === false) {
+      jQuery.extend(BookReader.defaultOptions, {
+        enableSearch: false
+      });
+    }
+
     super_.call(this, options);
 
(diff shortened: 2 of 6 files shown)


src/BookNavigator/book-navigator.js
```diff
@@ -237,7 +237,11 @@ export class BookNavigator extends LitElement {
     }
 
     this.menuProviders = providers;
-    this.addMenuShortcut('search');
+    if (this.bookreader.options.enableSearch) {
```
cdrini, line 240: Isn't this done by line 178?

src/plugins/search/plugin.search.js
```diff
@@ -45,6 +45,13 @@ jQuery.extend(BookReader.defaultOptions, {
 /** @override */
 BookReader.prototype.setup = (function (super_) {
   return function (options) {
+    // override with config
+    if (options.enableSearch === false) {
+      jQuery.extend(BookReader.defaultOptions, {
+        enableSearch: false
+      });
+    }
```
cdrini, line 53: This is changing the default bookreader options if the user option is set to false; are we sure this is needed?

iisa: hi drini, thanks for asking.. the problem i am trying to solve is - that `enableSearch` flag that gets set [above on line 40] (https://github.com/internetarchive/bookreader/blob/master/src/plugins/search/plugin.search.js#L35-L43) - when the search plugin script is loaded.  This overrides any config sets when BR gets loaded.

the check in the constructor uses the received value to see if it needs to be turned off.

the only other option that i cannot avail is to _not_ load the plugin script.

cdrini: The plugins are loaded and modify the default options, `BookReader.defaultOptions`. These are then overriden inside the bookreader contructor:

https://github.com/internetarchive/bookreader/blob/3755718097e0023609c0615553099f8572c4c568/src/BookReader.js#L57

So they _should_ always be happening in the right order?

iisa:
closing. not needed

PR #1176, opened 2023-03-30
Update test deps + improve jest errors

cdrini:
If an error is thrown in a promise, jest by default throws a very generic error and doesn't show the call stack. Change it to display more. Also do less transpiling when running jest to avoid confusing call stack entries from eg regenerator runtime (used to poly async/await).

Also never try to update eslint past v7. v8. Doesn't support decorators at all.


cdrini:
Auto-merging. Small tweaks + only affects CI

PR #1178, opened 2023-04-04
Call updateVisibleRegion after the UI has rendered

cdrini:
Addendum to #1177


cdrini:
Auto-merging; small addendum to #1177 .

PR #1179, opened 2023-04-04
Update all build deps

cdrini:
Getting annoying deprecation errors from babel that make test output and build output very difficult to read.


cdrini:
Closed by #1198

PR #1186, opened 2023-05-04
Fix mode2up preview pages hanging

cdrini:
Addendum to #1185


cdrini:
Auto-merging ; small follow up fix

PR #1187, opened 2023-05-04
Add GA event for text selection

cdrini:


src/plugins/plugin.text_selection.js
```diff
@@ -273,7 +274,20 @@ export class BookreaderWithTextSelection extends BookReader {
       // contains a reference to it.
       this.options.plugins.textSelection = options;
       this.textSelectionPlugin.init();
+
+      // Track how often selection is used
+      const sso = new SelectionStartedObserver('.textSelectionSVG', () => {
+        // Don't assume the order of the plugins ; the analytics plugin could
+        // have been added later. But at this point we should no for certain.
```
cdrini, line 281:
```suggestion
        // have been added later. But at this point we should know for certain.
```

src/plugins/plugin.text_selection.js
```diff
@@ -273,7 +274,20 @@ export class BookreaderWithTextSelection extends BookReader {
...
+        if (!this.archiveAnalyticsSendEvent) {
+          sso.detach();
+        } else {
+          this.archiveAnalyticsSendEvent?.('BookReader', 'SelectStart');
```
cdrini, line 285:
```suggestion
          this.archiveAnalyticsSendEvent('BookReader', 'SelectStart');
```

src/plugins/plugin.text_selection.js
```diff
@@ -273,7 +274,20 @@ export class BookreaderWithTextSelection extends BookReader {
...
+        }
+      });
+      sso.attach();
```
iisa, line 288: do you need to pin this somewhere to detach later?

iisa (approved):
looks good just a question

src/plugins/plugin.text_selection.js
cdrini: I have! I detach it weirdly just above :P

PR #1203, opened 2023-07-05
Search inside option for {{{/}}} + http support

### LINKED ISSUE #1188: Allow search API url to include port and to use http vs https
Would it be possible to allow the search url to include a port, and to use http instead of forcing https? Perhaps as properties in the `overrides` param that is passed in. 

Specifically here-
https://github.com/internetarchive/bookreader/blob/f3357ecdd72b7cf7209ae39fe3e21980837c3d1c/src/plugins/search/plugin.search.js#L172

cdrini:
Also closes #1188 

We will be switching from {{{/}}} to a less common string, since issues arise when the original OCR contains eg `{{{`.

Changes:
- New options for `searchInsidePreTag: {{{`, `searchInsidePostTag: }}}`
- DRY where a search is rendered
- Small style tweaks to search hover cards
- Also new option `searchInsideProtocol` for #1188 .

### DIFF
--- a/src/plugins/search/utils.js
+++ b/src/plugins/search/utils.js
@@ -10,8 +10,8 @@ export function renderMatch(match, preTag, postTag) {
   // Search results are returned as a text blob with the hits wrapped in
   // triple mustaches. Hits occasionally include text beyond the search
   // term, so everything within the staches is captured and wrapped.
-  const preTagRe = escapeRegExp(preTag);
-  const postTagRe = escapeRegExp(postTag);
+  const preTagRe = escapeRegExp(escapeHTML(preTag));
+  const postTagRe = escapeRegExp(escapeHTML(postTag));
   // [^] matches any character, including line breaks
   const regex = new RegExp(`${preTagRe}([^]+?)${postTagRe}`, 'g');
   return escapeHTML(match)
--- a/src/plugins/search/utils.js
+++ b/src/plugins/search/utils.js
@@ -1,0 +1,43 @@
+import { escapeHTML, escapeRegExp } from '../../BookReader/utils.js';
+
+/**
+ * @param {string} match
+ * @param {string} preTag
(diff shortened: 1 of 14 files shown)


src/plugins/search/utils.js
```diff
@@ -0,0 +1,43 @@
...
+ * @returns {string}
+ */
+export function renderMatch(match, preTag, postTag) {
+  // Search results are returned as a text blob with the hits wrapped in
+  // triple mustaches. Hits occasionally include text beyond the search
+  // term, so everything within the staches is captured and wrapped.
+  const preTagRe = escapeRegExp(preTag);
+  const postTagRe = escapeRegExp(postTag);
+  // [^] matches any character, including line breaks
+  const regex = new RegExp(`${preTagRe}([^]+?)${postTagRe}`, 'g');
```
iisa, line 16: 👍

tests/jest/plugins/search/utils.test.js
```diff
@@ -0,0 +1,18 @@
+import { marshallSearchResults } from '@/src/plugins/search/utils.js';
+import { deepCopy } from '@/tests/jest/utils.js';
+import { DUMMY_RESULTS } from './utils.js';
+
+describe('marshallSearchResults', () => {
+  test('Adds match index', () => {
+    const results = deepCopy(DUMMY_RESULTS);
+    marshallSearchResults(results, x => x.toString(), '{{{', '}}}');
```
iisa, line 8: q: do we need to show new delimiter characters as example?

iisa (approved):
looks good 👍 just a question about new character example..

tests/jest/plugins/search/utils.test.js
cdrini: Great idea! I added a test case to this file with the "official" delimiters.

PR #1207, opened 2023-08-01
Downgrade core-js to fix esm.sh build issue

cdrini:
This was a strange issue.

So core-js in commit https://github.com/zloirock/core-js/commit/25e93739aa8623a47526cc65f2d8e5d68bd3b107 changed some code from `typeof global.process` to `typeof process`. Unfortunately this resulted in some part of esm.sh when transpiling for deno to think that the node library `process` needed to be imported. You can see it on the first line here: https://esm.sh/v130/@internetarchive/bookreader@5.0.0-64/deno/BookReader/BookReader.js . 

So for now rollback core-js to the previous version before this change was introduced. Not sure where the bug lies here ; whether this is an issue with core-js or esm.sh . But this is the quickest way for us to unblock our deploy flow to archive.org

Created an issue on core-js at least: https://github.com/zloirock/core-js/issues/1277

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
         "babel-loader": "9.1.2",
         "codecov": "^3.8.3",
         "concurrently": "7.4.0",
-        "core-js": "3.30.2",
+        "core-js": "3.27.1",
         "cpx2": "4.2.3",
         "eslint": "^7.32.0",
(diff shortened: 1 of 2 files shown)


cdrini:
Auto-merging ; tiny change

PR #1209, opened 2023-08-10
Fix iOS pinch zoom + replace Hammer w interactjs

### LINKED ISSUE #1128: Pinch Zoom unusable on iOS

cdrini:
Closes #1128

* Switches from Hammer.js, which has been discontinued, to interact.js, which I found out has also been discontinued -_-
* Fixes broken pinch zoom on iOS devices. Only able to test on my physical iOS 15 iPad
* Fixes broken pinch zoom on Samsung Internet browser. Interact.js let as fix this one, which is why I didn't switch back to Hammer. Also interact is a little more recently updated.
* The main fix was forcing thing to happen a bit more sequentially ; forcing only one scrollTop change to be allowed at a time effectively.
* Increased the bundle size by 100k, which sucks, but that's about the size of one image, so feeling ok for now? Will investigate if I can partially load interactjs or even potentially switch back to hammerjs in the future.

Known issues:
* Going to fullscreen loses page position on iOS after a zoom has happened.


src/BookReader/Mode1Up.js
```diff
@@ -51,7 +51,7 @@ export class Mode1Up {
         new DragScrollable(this.mode1UpLit, {
           preventDefault: true,
           dragSelector: '.br-mode-1up__visible-world',
-          // Only handle mouse events; let browser/HammerJS handle touch
+          // Only handle mouse events; let browser/interact.js handle touch
```
dualcnhq, line 54: 👍

src/util/browserSniffing.js
dualcnhq, line 1: this util file is really 😎 🆒 ...

dualcnhq (approved):
:lgtm:

cdrini:
Awesome, thanks @dualcnhq !

PR #1210, opened 2023-08-11
add custom TTS playback speed

ivan20203:
--> what:
the TTS (text to speech) speed has fixed settings of 0.25-0.5-0.75-1.00-1.25- and so on. 

I implemented a custom speed option. 
So now you will be able to type in numbers between 0 to 5 and with 2 decimal places. 
So the value 1.1 will work. 

--> how:
After setting up my the dev environment, this is where I had like 74 updates in git for making the files multi line. 
I added it in BookReader/plugins/plugin.tts.js

After having a multiline implementation i asked ChatGPT to convert the new code i wrote into single line format that the current plugin.tts.js requires. 

So thats where we are. After testing it I am now making a pull request. 

Apologies for this if it was done the wrong way. 

Pictures of working version below: 

Screenshot 2023-08-10 at 17 23 28

Screenshot 2023-08-10 at 17 23 44

Screenshot 2023-08-10 at 17 23 56


ivan20203:
code changed: 

From: BookReader.prototype.initNavbar=function(t){return function() 
To: n.find(".BRcontrols .BRicon.zoom_out").closest("li"))}return n}}

changed in between them to add custom playback speed.

ivan20203:
Edited wrong files. Edited plugin.tts.js in BookReader instead of src

PR #1238, opened 2023-09-11
Encode page number in url

cdrini:
New page number detection logic now supports more complicated page numbers, which might require encoding! So add support for that.


cdrini:
@iisa CR'd over slack. Thanks @iisa !

Issue #1103, opened 2022-11-01
Some demos not working


navigatrum:
**Description**
I'm experiencing the same bug described here: https://stackoverflow.com/questions/71876803/how-to-get-the-demo-of-the-internet-archive-bookreader-working-in-2022

The problem seems related to `css`  `var`: `--br-height` not being defined, so that:
```css
.BookReader {
    height: var(--br-height);
  ...
}
```
in [BookReaderDemo.css#L32-L36](https://github.com/internetarchive/bookreader/blob/master/BookReaderDemo/BookReaderDemo.css#L32-L36) sets the `height` to `0` for the elements of class `BookReader` in the following, not-working, demos:

- [demo-advanced.html](https://github.com/internetarchive/bookreader/blob/master/BookReaderDemo/demo-advanced.html)
- [demo-iif](https://github.com/internetarchive/bookreader/blob/master/BookReaderDemo/demo-iiif.html)
- [demo-multiple.html](https://github.com/internetarchive/bookreader/blob/master/BookReaderDemo/demo-multiple.html)
- [demo-preview-pages.html](https://github.com/internetarchive/bookreader/blob/master/BookReaderDemo/demo-preview-pages.html)
- [demo-simple.html](https://github.com/internetarchive/bookreader/blob/master/BookReaderDemo/demo-simple.html)
- [immersion-1up.html](https://github.com/internetarchive/bookreader/blob/master/BookReaderDemo/immersion-1up.html)
- [viewmode-cycle.html](http://127.0.0.1:8000/BookReaderDemo/viewmode-cycle.html)

The problem doens't affect full-screen-like demos such as [demo-fullscreen-mobile.html](https://github.com/internetarchive/bookreader/blob/master/BookReaderDemo/demo-fullscreen-mobile.html) where the `height` for `.BookReader`  is defined inline in the `head` (https://github.com/internetarchive/bookreader/blob/master/BookReaderDemo/demo-fullscreen-mobile.html#L24-L27).

[...]

PR #1239, opened 2023-09-14
fix:Some demos not working.

xonx4l:
issue -:#1103


cdrini (approved):
Thank you @xonx4l ! Confirmed the simple demo now displays correctly.

PR #1240, opened 2023-09-14
Avoid ppi<10 causing UI to break

cdrini:
Testing: https://www-drini.archive.org/details/micro_IAMFTEST00_0022/page/60/mode/1up now works

### DIFF
--- a/src/BookReader/BookModel.js
+++ b/src/BookReader/BookModel.js
@@ -352,8 +352,8 @@ export class PageModel {
    * @param {PageIndex} index
    */
   constructor(book, index) {
-    // TODO: Get default from config
-    this.ppi = book._getDataProp(index, 'ppi', book.ppi);
+    // Values less than 10 cause the UI to not work correctly
+    this.ppi = Math.max(book._getDataProp(index, 'ppi', book.ppi), 10);
     this.book = book;
     this.index = index;
     this.width = book.getPageWidth(index);
--- a/src/BookReader/BookModel.js
+++ b/src/BookReader/BookModel.js
@@ -353,7 +353,8 @@ export class PageModel {
    */
   constructor(book, index) {
     // Values less than 10 cause the UI to not work correctly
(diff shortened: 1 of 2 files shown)


src/BookReader/BookModel.js
```diff
@@ -352,8 +352,8 @@ export class PageModel {
    * @param {PageIndex} index
    */
   constructor(book, index) {
-    // TODO: Get default from config
-    this.ppi = book._getDataProp(index, 'ppi', book.ppi);
+    // Values less than 10 cause the UI to not work correctly
+    this.ppi = Math.max(book._getDataProp(index, 'ppi', book.ppi), 10);
```
cdrini, line 356:
```suggestion
    // Values less than 10 cause the UI to not work correctly
    const pagePPI = book._getDataProp(index, 'ppi', book.ppi);
    this.ppi = Math.max(pagePPI < 10 ? book.ppi : pagePPI, 10);
```

src/BookReader/BookModel.js
```diff
@@ -352,8 +352,9 @@ export class PageModel {
    * @param {PageIndex} index
    */
   constructor(book, index) {
-    // TODO: Get default from config
-    this.ppi = book._getDataProp(index, 'ppi', book.ppi);
+    // Values less than 10 cause the UI to not work correctly
+    const pagePPI = book._getDataProp(index, 'ppi', book.ppi);
+    this.ppi = Math.max(pagePPI < 10 ? book.ppi : pagePPI, 10);
```
iisa, line 357: 👍 q: what does pagePPI default to?

iisa (approved):
🚢

src/BookReader/BookModel.js
cdrini: Oh the default is the last argument to getDataProp, book.ppi . This is the value in the metadata for that entire book

PR #1242, opened 2023-09-19
Import Share & Volumes panels from <iaux-item-navigator>

iisa:
## Testing:
 - ✅ multiple-files PR demo works like master branch demo app
   - non-default sort is updated in url
   - has multiple-files shortcut that opens to multiple-files-panel
   - has share panel
   - [optimized] PR branch now opens multi-files volume on load (regression that is fixed in this PR)

- DEMO URLS:
  - master branch: https://lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive
  - this pull request: https://deploy-preview-1242--lucid-poitras-9a1249.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala
  - go to demo -> scroll down & click "Multiple files" button to load proper demo


src/BookNavigator/volumes/volumes-provider.js
```diff
@@ -17,84 +13,71 @@ export default class VolumesProvider {
...
+    this.onProviderChange(this.bookreader);
   }
 
   /**
    * @param {'default' | 'title_asc' | 'title_desc'} sortByType
    */
-  sortVolumes(sortByType) {
-    let sortedFiles = [];
+  async handleFileListSorted(event) {
+    const { sortType, sortedFiles } = event.detail;
```
dualcnhq, line 60: can we make the naming consistent? the renaming of the variable `sortOptions` is better, so... in this case, can we make it `sortOption` instead?

dualcnhq (approved):
LGTM, just a little comment about the variable naming

src/BookNavigator/sharing.js
```diff
@@ -1,6 +1,6 @@
 import { html } from 'lit';
-import '@internetarchive/icon-share/icon-share';
-import '@internetarchive/ia-sharing-options';
+import { iauxShareIcon } from '@internetarchive/ia-item-navigator';
```
cdrini, line 2: This is a bit strange for a top-level export ; can we do something like this?

```suggestion
import { iauxShareIcon } from '@internetarchive/ia-item-navigator/sharePanel';
```

src/BookNavigator/sharing.js
```diff
@@ -12,16 +12,16 @@ export default class SharingProvider {
     const creatorToUse = Array.isArray(creator) ? creator[0] : creator;
     const subPrefix = bookreader.options.subPrefix || '';
     const label = `Share this book`;
-    this.icon = html`<ia-icon-share style="width: var(--iconWidth); height: var(--iconHeight);"></ia-icon-share>`;
+    this.icon = html`${iauxShareIcon}`;
     this.label = label;
     this.id = 'share';
-    this.component = html`<ia-sharing-options
+    this.component = html`<iaux-sharing-options
```
cdrini, line 18: I think we might want to standardize on a naming scheme for these ; I think `<iaux-sharing-panel>`, `<iaux-files-panel>`, etc might be the way to go! I generally change the prefix to be who 'owns' the components, so BookReader sub-components start with `br` (eg `<br-mode-1up>`). So maybe we could `<in-sharing-panel>` for `itemnav`?

src/BookNavigator/volumes/volumes-provider.js
cdrini, line 1: We can delete the `volumes` dir now! And maybe rename to something like `files-panel.js`

cdrini (commented):
Nice! Love how much is getting nixxed here 😊

I think there's some important data that might need to also be moved over, though, specifically for the volumes provider.

Otherwise some naming things that I think would make things a little clearer!

src/BookNavigator/volumes/volumes-provider.js
```diff
@@ -17,88 +16,73 @@ export default class VolumesProvider {
...
+    this.sortOrderBy = sortTypes.default;
 
-    this.sortOrderBy = sortType.default;
+    this.component = document.createElement("iaux-viewable-files");
+    this.component.addSortToUrl = true;
+    this.component.subPrefix = bookreader.options.subPrefix || "";
+    this.component.baseHost = baseHost;
+    this.component.fileList = [...this.viewableFiles];
+
+    this.sortFilesComponent = document.createElement("iaux-sort-viewable-files");
```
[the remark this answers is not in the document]
iisa: thought about it. right now,  viewable-files panel does not have access to the header of item nav, this piece must be passed through via config.  refactoring for otherwise is not in scope at the moment.

src/BookNavigator/volumes/volumes-provider.js
iisa: thanks, ive updated the file to use `sortTypes` and then added a new type to help make sense of usage around file

src/BookNavigator/volumes/volumes-provider.js
```diff
@@ -17,88 +16,73 @@ export default class VolumesProvider {
...
+  async handleFileListSorted(event) {
+    const { sortType, sortedFiles } = event.detail;
 
-    const files = this.viewableFiles;
-    sortedFiles = files.sort((a, b) => {
-      if (sortByType === sortType.title_asc) return a.title.localeCompare(b.title);
-      else if (sortByType === sortType.title_desc) return b.title.localeCompare(a.title);
-      else return a.orig_sort - b.orig_sort;
-    });
+    this.viewableFiles = sortedFiles;
```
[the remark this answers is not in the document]
iisa: yah. it isn't right now, but i think we should collapse even further. this will take a bit of time to rethink bc it involves item-nav menu config as button ingress. will give rethink without holding pdf theater feature longer than needed. 👍

iisa (commented):
addressed cr - new component names <iaux-in-share-panel>, <iaux-in-viewable-files-panel>

src/BookNavigator/sharing.js
iisa: yes I like that. updated to `<iaux-in-share-panel>` - keeping `iaux-` as latest preferred web component prefix

PR #1248, opened 2023-09-27
Fix TTS play/pause button remaining in paused state after re-opening

### LINKED ISSUE #1247: TTS play/pause button remains in paused state after re-opening TTS menu
**Description**

> What problem are we solving? What does the experience look like today? What are the symptoms? How can one reproduce the problem?

After pausing TTS and closing it by re-selecting 'Read this book aloud', if TTS is opened again, playback will start with the play/pause button still in the paused state. This forces the user to click the button to reset it to the play state before being able to press it again to pause playback. 

> Include evidence of the problem or provide a screenshot.

**Expectation**

> What should be happening? What will it look like / how will it behave when fixed?

When TTS is started by selecting the read aloud button the play/pause button should default to the play state since TTS playback has started.

**Context**

> What operating system? What browser type and version?

Issue on Chrome and Firefox, Windows

**Proposal & Constraints**

> What is the proposed solution / implementation? Is there a precedent of this approach succeeding elsewhere?

**Stakeholders**

> @-tag any stakeholders of this bug

@cdrini

sbwhitt:
Closes #1247 

Previously the play/pause button for TTS would remain in the paused state after pausing TTS and closing/re-opening. This forced the user to click twice to pause playback.

Testing:

- Open TTS on a book
- Pause TTS playback
- Close TTS by selecting the 'Read aloud' button
- Re-open by selecting the 'Read aloud' button
- Verify that the button is in the correct state (playing)
- Verify that clicking the pause button actually pauses playback

### DIFF
--- a/src/plugins/tts/plugin.tts.js
+++ b/src/plugins/tts/plugin.tts.js
@@ -251,7 +251,7 @@ BookReader.prototype.ttsPlayPause = function() {
     this.ttsToggle();
   } else {
     this.ttsEngine.togglePlayPause();
-    this.ttsUpdateState(this.ttsEngine.paused);
+    this.ttsUpdateState();
   }
 };
 
--- a/src/plugins/tts/AbstractTTSEngine.js
+++ b/src/plugins/tts/AbstractTTSEngine.js
@@ -80,6 +80,7 @@ export default class AbstractTTSEngine {
    */
   start(leafIndex, numLeafs) {
     this.playing = true;
+    this.paused = false;
     this.opts.onLoadingStart();
 
     this._chunkIterator = new PageChunkIterator(numLeafs, leafIndex, {
@@ -95,6 +96,7 @@ export default class AbstractTTSEngine {
   stop() {
     if (this.activeSound) this.activeSound.stop();
     this.playing = false;
+    this.paused = true;
     this._chunkIterator = null;
     this.activeSound = null;
     this.events.trigger('stop');


src/plugins/tts/plugin.tts.js
```diff
@@ -251,7 +251,9 @@ BookReader.prototype.ttsPlayPause = function() {
     this.ttsToggle();
   } else {
     this.ttsEngine.togglePlayPause();
-    this.ttsUpdateState(this.ttsEngine.paused);
+    // extra ttsUpdateState call?
+    // already bound to play/pause button in initNavbar
+    //this.ttsUpdateState(this.ttsEngine.paused);
```
sbwhitt, line 256: I noticed ttsUpdateState being called twice when clicking play/pause button. If this is deemed unnecessary I will remove the call and comments entirely.

cdrini: We might be able to remove this ; but it probably make more sense to remove from initNavBar. 
Although it being called twice is ok it just syncs the state, so it might need to happen twice at various points in the pipeline to ensure the state is in sync.

sbwhitt: I'm ok with keeping it then. I can add it back and take out the param to prevent confusion if that works

cdrini (approved):
Lgtm! Tested on Win/FF and worked like a charm!

PR #1267, opened 2023-11-02
Upgrade to Node 20

cdrini:


cdrini:
Auto-merging ; small change, rest of IA already uses node 20

PR #1268, opened 2023-11-03
Fix error with localStorage in sandboxed iframe

cdrini:
Tested:

- works in embedded iframe w sandboxing
- ReadAloud voice remembering works


iisa (approved):
lgtm! remember only when not iframe handles most of the readership cases 👍

PR #1287, opened 2023-12-05
Gitignore .env file

cdrini:
Use this to store helpful env variables like browserstack keys. Gitiginore that.


cdrini:
Auto-merging, small config change

PR #1293, opened 2023-12-19
5.0.0-76

dualcnhq:


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 5.0.0-76
+Feature: Update bookNav shortcut menu ordering
```
cdrini, line 2: Let's please follow the pattern here. Bullet, include the person, and it looks like this is more of a fix than a feature.

cdrini (approved):

PR #1329, opened 2024-06-19
Update test dependencies

cdrini:


cdrini:
Auto-merging, small dependency PR, since CI passes and no change to build file sizes.

PR #1343, opened 2024-09-23
Bump cache action to v4

williamtheaker:
There don't appear to be any breaking [changes between v3 and v4 as of writing](https://github.com/actions/cache/releases).

Fixes this warning:
> The following actions use a deprecated Node.js version and will be forced to run on node20: actions/cache@v3. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

https://github.com/internetarchive/bookreader/actions/runs/10930025193


cdrini (approved):
Lgtm, thanks @williamtheaker !

PR #1351, opened 2024-10-15
Remove polyfill.io - service no longer secure

cdrini:
See https://fossa.com/blog/polyfill-supply-chain-attack-details-fixes/


cdrini:
Auto-merging; small change and has security implications

PR #1352, opened 2024-10-15
Update build dependencies

cdrini:


cdrini:
Auto-merging; dependencies chore, and tests pass. Note e2e offline while archive.org is down, so skipping those.

PR #1353, opened 2024-10-15
Update some dev deps

cdrini:
Tested npm cmmands build and serve-dev work


cdrini:
Auto-merging; small dep update

PR #1355, opened 2024-11-14
WEBDEV-7113 Switch to new Thorium URL

rebecca-shoptaw:
Currently, our "Install Thorium" link takes you to a more general [about page](https://www.edrlab.org/software/thorium-reader/). To avoid patron confusion, we should instead link to their new [download page](https://thorium.edrlab.org/).


cdrini (approved):
Lgtm ; not able to test since the bookreader review app needs a bit of fiddling, but seems low risk ! Confirmed the URL works.

PR #1359, opened 2024-11-27
WEBDEV-7126 Revise LCP download message

rebecca-shoptaw:
Removes outdated Aldiko + SimplyE links and clarifies text re: Thorium. Also makes variable name less brittle.


cdrini (approved):
Lgtm :+1: Wasn't able to find a way to test this exact code in our local env, but it looks low risk.

PR #1365, opened 2025-01-02
Switch to new page number display format

pezvi:
- Split desktop and mobile page number displays
- CSS: Move 'hide' to .controls
- Add/split tests for desktop and mobile page number displays

### DIFF
--- a/src/BookReader/Navbar/Navbar.js
+++ b/src/BookReader/Navbar/Navbar.js
@@ -142,10 +142,34 @@ export class Navbar {
     if (this.br.options.bookType !== 'linerNotes') {
       if (this.br.refs.$brContainer.prop('clientWidth') < 640) {
         this.showMinimumNavbarControls();
+        this.showMinimumNavPageNum();
       } else {
         this.showMaximumNavbarControls();
+        this.showMaximumNavPageNum();
       }
     }
+  }
+
+  /**
+   * Switch Book Nav page number display to minimum/mobile
+   */
+  showMinimumNavPageNum() {
+    const minelement = document.querySelector(`.BRcurrentpage.BRmin`);
+    const maxelement = document.querySelector(`.BRcurrentpage.BRmax`);
+
+    if (minelement) minelement.classList.remove('hide');
+    if (maxelement) maxelement.classList.add('hide');
+  }
+
+  /**
+   * Switch Book Nav page number display to maximum/desktop
+   */
+  showMaximumNavPageNum() {
+    const minelement = document.querySelector(`.BRcurrentpage.BRmin`);
+    const maxelement = document.querySelector(`.BRcurrentpage.BRmax`);
+
+    if (minelement) minelement.classList.add('hide');
+    if (maxelement) maxelement.classList.remove('hide');
   }
 
   /**
@@ -203,7 +227,10 @@ export class Navbar {
                   <div class="BRpager"></div>
                   <div class="BRnavline"></div>
                 </div>
-                <p><span class='BRcurrentpage'></span></p>
+                <p>
+                  <span class="BRcurrentpage BRmax"></span>
+                  <span class="BRcurrentpage BRmin"></span>
+                </p>
               </li>
               ${this._renderControls()}
             </ul>
@@ -246,9 +273,10 @@ export class Navbar {
   /**
   * Returns the textual representation of the current page for the navbar
   * @param {number} index
+  * @param {"max"|"min"} size
   * @return {string}
   */
-  getNavPageNumString(index) {
+  getNavPageNumString(index, size) {
     const { br } = this;
     // Accessible index starts at 0 (alas) so we add 1 to make human
     const pageNum = br.book.getPageNum(index);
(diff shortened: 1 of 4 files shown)


src/BookReader/Navbar/Navbar.js
```diff
@@ -298,16 +328,28 @@ export class Navbar {
...
   const pageIsAsserted = pageNum[0] != 'n';
+  const pageIndex = index + 1;
 
   if (!pageIsAsserted) {
-    const pageIndex = index + 1;
-    return `(${pageIndex} of ${numLeafs})`; // Page (8 of 10)
+    pageNum = `—`;
+  }
+
+  if (size=="max") {
```
jbuckner, line 342: There's a spacing issue here and you probably want to use triple-equal so it does an exact match (double-equals does a loose comparison: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness)

src/BookReader/Navbar/Navbar.js
```diff
@@ -298,16 +328,28 @@ export class Navbar {
...
  */
-export function getNavPageNumHtml(index, numLeafs, pageNum, pageType, maxPageNum) {
+export function getNavPageNumHtml(index, numLeafs, pageNum, pageType, maxPageNum, size) {
   const pageIsAsserted = pageNum[0] != 'n';
+  const pageIndex = index + 1;
 
   if (!pageIsAsserted) {
-    const pageIndex = index + 1;
-    return `(${pageIndex} of ${numLeafs})`; // Page (8 of 10)
+    pageNum = `—`;
```
jbuckner, line 339: Minor nit, but since you're not doing string interpolation, you can just use normal single-quotes instead of backticks.

src/BookReader/Navbar/Navbar.js
pezvi: Thanks—yeah, I'm cleaning up the spacing issues and will switch to ===

src/BookReader/Navbar/Navbar.js
```diff
@@ -298,16 +328,28 @@ export class Navbar {
...
+  }
+
+  if (size === "max") {
+    return `Page ${pageNum} (${pageIndex}/${numLeafs})`;
   }
 
   const bookLengthLabel = (maxPageNum && parseFloat(pageNum)) ? ` of ${maxPageNum}` : '';
-  return `${pageNum}${bookLengthLabel}`;
+
+  if (size === "min" && !pageIsAsserted) {
```
rebecca-shoptaw, line 348: You don't need the `size === "min"` check here since we've already established it's not max (since anything with max would have already returned above)

src/BookReader/Navbar/Navbar.js
```diff
@@ -298,16 +328,28 @@ export class Navbar {
...
+    return `(${pageIndex} of ${numLeafs})`;
+  }
+
+  if (size === "min" && pageIsAsserted) {
```
rebecca-shoptaw, line 352: You actually don't need either `if` condition here, since you've already established the size is not `max` (see above), _and_ by the same logic you've already established that `pageIsAsserted` must be true.

I'd also recommend moving the `bookLengthLabel` calculation down here, since you won't need it for either of the other cases, i.e. this whole block could be:

```
if (size === "max") {
    return `Page ${pageNum} (${pageIndex}/${numLeafs})`;
  }

if (!pageIsAsserted) {
  return `(${pageIndex} of ${numLeafs})`;
}

const bookLengthLabel = (maxPageNum && parseFloat(pageNum)) ? ` of ${maxPageNum}` : '';
return `${pageNum}${bookLengthLabel}`;
```

tests/jest/BookReader/Navbar/Navbar.test.js
```diff
@@ -4,16 +4,29 @@ import BookReader from '@/src/BookReader.js';
...
+    expect(f(3, 40, '14', '', null, 'min')).toBe('14');
+  });
+
+  test('handle n-prefixed page numbers-max format', () => {
+    expect(f(3, 40, 'n3', '', 40, 'max')).toBe('Page — (4/40)');
   });
 
-  test('handle regular page numbers', () => {
-    expect(f(3, 40, '14', '', 40)).toBe('14 of 40');
+  test('handle regular page numbers-max format', () => {
```
rebecca-shoptaw, line 24: Nice!! Thank you for adding these 🙂

rebecca-shoptaw (commented):
Some philosophical/vibes/style thoughts, but it looks great to me! And thanks for adding the extra tests 🙂

src/BookReader/Navbar/Navbar.js
pezvi: Ok, yeah—I wanted to do this (skip the check for "min") but was concerned it wouldn't be quite as clear or even maybe some other value might sneak through... but I think you bringing it up as an alternative indicates it should be fine.

src/BookReader/Navbar/Navbar.js
```diff
@@ -142,12 +142,36 @@ export class Navbar {
...
+        this.showMaximumNavPageNum();
       }
     }
   }
 
+  /**
+   * Switch Book Nav page number display to minimum/mobile
+   */
+  showMinimumNavPageNum() {
+    const minelement = document.querySelector(`.BRcurrentpage.BRmin`);
```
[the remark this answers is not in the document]
pezvi: Thanks for catching that one

PR #1387, opened 2025-03-15
Add option to ChaptersPlugin for explicit openLibraryId

cdrini:
And rename some of the options -- with backwards compatibility.


cdrini:
Auto-merging; small addition, no breaking changes.

PR #1391, opened 2025-03-20
Fix TTS scrolling to top of page unexpectedly

### LINKED ISSUE #1316: Return OCR TTS to reading whole page, rather than section
> Is your feature request related to a problem? What is the motivation for implementing this feature?

Right now when you use TTS to read, it goes and highlights the coordinates of the words. 

It darts to it. So what ends up happening is at the bottom of the page, when the TTS hits the words it darts back up and then to the OCR coordinates. 

I remember using the archive reader before and it would not dart back and forth on new passages. It used to go all the way down until the passage is not seen, and then switch to a new page. 

> What is the proposed solution / implementation? Is there a precedent for this approach succeeding elsewhere?

The proposed solution is figure out what code was changed and revert to the old behavior of not darting back and forth on TTS.

cdrini:
Closes #1316

### DIFF
--- a/src/plugins/tts/plugin.tts.js
+++ b/src/plugins/tts/plugin.tts.js
@@ -278,10 +278,8 @@ export class TtsPlugin extends BookReaderPlugin {
    * @param {Number} leafIndex
    */
   async maybeFlipToIndex(leafIndex) {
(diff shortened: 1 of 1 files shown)


cdrini:
Auto-merging; very small fix.

PR #1394, opened 2025-03-26
Fix ctrl-c textsel when protected book

cdrini:


cdrini:
Auto-merging, small addendum to #1142

PR #1407, opened 2025-04-30
Remove BookReader/ build dir from repo

cdrini:
This is a breaking change on the off chance someone was using the static files directly from the repo, but I that is a very rare pattern now (even jquery no longer uses this pattern, as of [~9mo ago](https://github.com/jquery/jquery/commit/be048a027d0581746f71df7c8eb3ce1d9bd10a40)!). But if they are, they can now switch to unpkg (or similar) to get the same experience, eg https://unpkg.com/@internetarchive/bookreader@5.0.0-94/BookReader/BookReader.js , etc.


cdrini:
Auto-merging; this should have no difference on prod/etc, and team's been aligned on this for a while!

PR #1410, opened 2025-05-19
[GSoC] Auto-Translate: Add Properties, Show Translations as Lines

schu96:
1. Add type docs to properties of TranslatePlugin Class
2. Take translated paragraphs and divide into chunks using '.' as delimiter to show translations as lines to allow for graphics in digital works. 
    - Grabs CSS values from existing ```.BRlineElement``` and ```.BRparagraphElement``` to attempt to keep translated text within the bounds of the paragraph text being translated. 
    - Currently using overflow-y: auto to let overflow text be visible via scrolling
    - font-size property for spans are adjusted to be slightly smaller (5px smaller for now) since text from passages tends to go past the height of the container. 

image
image

### DIFF
--- a/BookReaderDemo/css/index.css
+++ b/BookReaderDemo/css/index.css
@@ -90,17 +90,3 @@
     box-shadow: 0 0 50px 50px #f8edc0ff;
     overflow-y: scroll;
   }
-
-  .translation .trParagraph {
-    pointer-events: all;
-    overflow-y: auto;
-    font-family: "Helvetica Neue", Arial, Verdana, sans-serif;
-    margin-bottom: 0;
-  }
-
-  .translation .trParagraph .trSentence {
-    white-space: preserve-spaces;
-    display: inline-block;
-    background: #f8edc0;
-    line-height: normal;
-  }
--- a/BookReaderDemo/css/index.css
+++ b/BookReaderDemo/css/index.css
@@ -85,11 +85,22 @@
     color: black;
     z-index: 9999;
     font-size: 2.85em;
-    height: 95%;
-    margin: 50px;
-    padding: 25px;
-    background-color: #f8edc0;
+    height: 100%;
     border-radius: 8px;
     box-shadow: 0 0 50px 50px #f8edc0ff;
     overflow-y: scroll;
+  }
+
+  .translation .trParagraph {
+    pointer-events: all;
+    overflow-y: auto;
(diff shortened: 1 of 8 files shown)


BookReaderDemo/css/index.css
```diff
@@ -85,11 +85,22 @@
...
-    margin: 50px;
-    padding: 25px;
-    background-color: #f8edc0;
+    height: 100%;
     border-radius: 8px;
     box-shadow: 0 0 50px 50px #f8edc0ff;
     overflow-y: scroll;
+  }
+
+  .translation .trParagraph {
```
cdrini, line 94: To keep the styling here consistent, giving the layer a `.BRtranslateLayer` class, and then the paragraphs/lines/etc can have the same classes (`.BRparagraphElement`, `.BRlineElement`, etc)

BookReaderDemo/css/index.css
```diff
@@ -85,11 +85,22 @@
...
+    pointer-events: all;
```
cdrini, line 95: Oh, we'll also want this CSS moved to one of the scss files; let's add it to `_TextSelection.scss` . Apologies, this was an issue with the original code!

cdrini (commented):
Nice, great start @schu96 !

src/plugins/translate/plugin.translate.js
```diff
@@ -151,13 +162,29 @@ export class TranslatePlugin extends BookReaderPlugin {
...
+        const fontSize = $(originalParagraphStyle).css("font-size");
+        const height = $(originalParagraphStyle).css("height");
+        const top = $(originalParagraphStyle).css("top");
+
+        translationPlaceholder.style.marginTop = marginTop;
+        translationPlaceholder.style.marginLeft = marginLeft;
+        // normal font size from transparent layer sometimes too big
+        translationPlaceholder.style.fontSize = `${parseInt(fontSize) - 5}px`;
+        translationPlaceholder.style.top = top;
+        translationPlaceholder.style.height = height;
```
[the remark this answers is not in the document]
cdrini: Ah right; good point! Let's skip for now :+1:

BookReaderDemo/css/index.css
cdrini: The former :+1: Note we shouldn't need a `tr` prefix on the paragraphs/lines/etc, but can re-use the class from the textLayer :+1:

PR #1415, opened 2025-06-24
Switch to using bergamot from npm instead of local copy

schu96:


src/plugins/translate/plugin.translate.js
```diff
@@ -99,20 +100,21 @@ export class TranslatePlugin extends BookReaderPlugin {
...
 
   /** @param {HTMLElement} page */
-  async translateRenderedLayer(page) {
+  async translateRenderedLayer(page, priority) {
     if (this.br.mode == this.br.constModeThumb) {
       return;
     }
 
     const pageIndex = page.dataset.index;
+    const pagePriority = parseInt(pageIndex) + priority;
```
cdrini: There's a bit of a gotcha with `parseInt` in js! It by default will not necessarily parse the string as base 10 😁 If the string begins with a "0x" (not just 0 as I said earlier!) it is parsed as base 16. So for best practice, we generally either always use the radix parameter, or we just use parseFloat (my preference) since it is quick and reliable/consistent.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#description
```suggestion
    const pagePriority = parseFloat(pageIndex) + priority;
```

src/plugins/translate/TranslationManager.js
```diff
@@ -67,54 +71,43 @@ export class TranslationManager {
...
-    }
-
     this.initPromise = new Promise((resolve, reject) => {
       this._initResolve = resolve;
       this._initReject = reject;
     });
+    const registryJson = await fetch("https://cors.archive.org/cors/mozilla-translate-models/registry.json").then(r => r.json());
+    for (const language of Object.values(registryJson)) {
+      for (const file of Object.values(language)) {
+        file.name = "https://cors.archive.org/cors/mozilla-translate-models/" + file.name;
```
cdrini: Let's update this to use the path of the registry url, eg
```suggestion
        file.name = registryUrl.replace('/[^\/]+$/', '/') + file.name;
```

That will make all the file names relative to the given url.

src/plugins/translate/TranslationManager.js
```diff
@@ -150,7 +143,7 @@ export class TranslationManager {
       reject: _reject,
     }
     this.currentModel = key;
-    this.worker.postMessage(["load_model", fromCode, toCode]);
+    // this.worker.postMessage(["load_model", fromCode, toCode]);
```
cdrini:
```suggestion
```

src/plugins/translate/TranslationManager.js
```diff
@@ -160,19 +153,22 @@ export class TranslationManager {
...
    */
 
-  getTranslation = async (fromLang, toLang, pageIndex, paragraphIndex, text) => {
+  getTranslation = async (fromLang, toLang, pageIndex, paragraphIndex, text, priority) => {
+    this.active = true;
     const key = `${fromLang}${toLang}-${pageIndex}:${paragraphIndex}`;
-
     const cachedEntry = this.alreadyTranslated.entries.find(x => x.index == key);
+    // Attempt to enforce ordered translation using priority pageIndex + paragraphIndex
+    const translationPriority = (priority ? priority : 0) + paragraphIndex;
```
cdrini: It seems like priority is always specified when we call this, so let's make it non-optional.

```suggestion
   * @param {number} priority
  * @return {Promise<string>} translated text
  */

  getTranslation = async (fromLang, toLang, pageIndex, paragraphIndex, text, priority) => {
    this.active = true;
const key = `${fromLang}${toLang}-${pageIndex}:${paragraphIndex}`;
const cachedEntry = this.alreadyTranslated.entries.find(x => x.index == key);
    // Attempt to enforce ordered translation using priority pageIndex + paragraphIndex
    const translationPriority = priority + paragraphIndex;
```

cdrini: Oh because we do the math for priority in the caller of this method, it would make sense to do this math in the parent as well. Here we can just forward along the set priority.

src/BookReader.js
```diff
@@ -330,6 +330,28 @@ BookReader.prototype.setup = function(options) {
...
+}
+
+BookReader.prototype.initializePlugin = function(pluginName) {
+  try {
+    this.plugins[pluginName].init();
+  }
+  catch (e) {
+    console.error(`Error initializing plugin outside of regular cycle ${pluginName}`, e);
+  }
+}
```
cdrini: Niiiice! These can be combined into one method; it's unlikely that we'll ever want to call these phases independently.

cdrini (changes requested):
Niice! A few small cleanups.

PR #1416, opened 2025-07-12
Translation Plugin QoL and Display improvements 

schu96:
Various QoL changes for display/fit
- Add lang to BRtranslateLayer
- Long hover over translated paragraph elements show scanned book image
- Enabled text selection for paragraph only
- Adjust font-size for visible page translated text paragraphs to prevent text overflow
- Minor styling changes throughout

### DIFF
--- a/src/BookReader/Mode2UpLit.js
+++ b/src/BookReader/Mode2UpLit.js
@@ -336,10 +336,12 @@ export class Mode2UpLit extends LitElement {
         },
         reduce: page.width / wToV(this.computePageWidth(page)),
       }).$container[0];
-
+    
+    // Should keep for initial render of the page
+    const wasVisible = pageContainerEl.classList.contains('BRpage-visible');
     const visibleStatus = pageContainerEl.classList.toggle('BRpage-visible', isVisible);
-    if (visibleStatus) {
-      this.br.trigger('pagevisible', {
+    if (visibleStatus && !wasVisible) {
+      this.br.trigger('pageVisible', {
         pageContainerEl,
       });
     }
@@ -604,7 +606,10 @@ export class Mode2UpLit extends LitElement {
       exitingPageAnimation.play();
       enteringPageAnimation.play();
 
-      nextPageContainers.forEach(c => c.$container.addClass('BRpage-visible'));
+      nextPageContainers.forEach(c => {
+        c.$container.addClass('BRpage-visible')
+
+      });
 
       await Promise.race([
         promisifyEvent(bookCenteringAnimation, 'finish'),
@@ -627,6 +632,11 @@ export class Mode2UpLit extends LitElement {
     this.visiblePages = (
       progression == 'lr' ? [nextSpread.left, nextSpread.right] : [nextSpread.right, nextSpread.left]
     ).filter(x => x);
+    nextPageContainers.forEach(c => {
+      this.br.trigger('pageVisible', {
+        pageContainerEl: c.$container[0]
+       });
+    });
   }
 
   /**
--- a/src/BookReader/Mode2UpLit.js
+++ b/src/BookReader/Mode2UpLit.js
@@ -337,7 +337,12 @@ export class Mode2UpLit extends LitElement {
         reduce: page.width / wToV(this.computePageWidth(page)),
       }).$container[0];
 
-    pageContainerEl.classList.toggle('BRpage-visible', isVisible);
+    const visibleStatus = pageContainerEl.classList.toggle('BRpage-visible', isVisible);
+    if (visibleStatus) {
+      this.br.trigger('pagevisible', {
+        pageContainerEl,
+      });
+    }
     return pageContainerEl;
   }
 
--- a/src/css/_TextSelection.scss
+++ b/src/css/_TextSelection.scss
@@ -135,26 +135,12 @@
     z-index: 3;
     --webkit-user-drag: none;
     user-select: text;
-    
-}
-
-.BRtranslateLayer .BRparagraphElementHover {
-    opacity: 0;
-    transition-delay: 1s;
-}
-
-.BRparagraphElement::selection {
-    background: hsla(210, 74%, 62%, 0.4);
-}
-.BRparagraphElement::-moz-selection {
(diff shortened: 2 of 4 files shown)


src/BookReader/Mode2UpLit.js
```diff
@@ -337,7 +337,12 @@ export class Mode2UpLit extends LitElement {
         reduce: page.width / wToV(this.computePageWidth(page)),
       }).$container[0];
 
-    pageContainerEl.classList.toggle('BRpage-visible', isVisible);
+    const visibleStatus = pageContainerEl.classList.toggle('BRpage-visible', isVisible);
+    if (visibleStatus) {
+      this.br.trigger('pagevisible', {
+        pageContainerEl,
+      });
+    }
```
cdrini: This is causing pagevisible to fire any time a scroll happens ; maybe only trigger it if it previously didn't have page-visible set? Ditto in 1up.

src/css/_TextSelection.scss
```diff
@@ -109,19 +116,45 @@
...
+    background: hsla(210, 74%, 62%, 0.4);
+    color:transparent;
 }
 
 .BRtranslateLayer .BRparagraphElement .BRlineElement {
     white-space: preserve-spaces;
     display: inline-block;
-    background: #f8edc0;
-    line-height: normal;
+    line-height: 1em;
```
cdrini: This will need to go on the BRparagraphElement, it's currently overriding the line height we set there.

src/css/_TextSelection.scss
```diff
@@ -109,19 +116,45 @@
...
     pointer-events: all;
+    backdrop-filter: none;
```
cdrini: Since this no longer has a background colour, this shouldn't be necessary here.
```suggestion
```

src/plugins/translate/plugin.translate.js
```diff
@@ -60,8 +60,25 @@ export class TranslatePlugin extends BookReaderPlugin {
...
+    this.br.on('pagevisible', (_, {pageContainerEl}) => {
+      const translationLayerEl = Array.from(pageContainerEl.childNodes).filter((ele) => {
+        return ele.classList.contains('BRtranslateLayer');
+      });
+      if (translationLayerEl) {
+        const translatedParagraphs = this.getTranslatedParagraphsOnPage(pageContainerEl);
+        for (const ele of translatedParagraphs) {
+          this.fitVisiblePage(ele);
+        }
+      }
```
cdrini: This can be done more simply with querySelectorAll, and removes the necessity for the helper function!

```js
for (const paragEl of pageContainerEl.querySelectorAll('.BRtranslateLayer > .BRparagraphElement')) {
    this.fitVisiblePage(paragEl);
}
```

src/plugins/translate/plugin.translate.js
```diff
@@ -156,6 +227,21 @@ export class TranslatePlugin extends BookReaderPlugin {
...
+  fitVisiblePage(ele) {
+    const originalFontSize = parseInt($(ele).css("font-size"));
+    let adjustedFontSize = `${originalFontSize - 1}px`;
+      while (ele.clientHeight < ele.scrollHeight) {
+        const currentFontSize = parseInt($(ele).css("font-size"));
+        adjustedFontSize = `${currentFontSize - 1}px`;
+        $(ele).css({
+          "font-size": adjustedFontSize,
+        });
+    }
```
cdrini, line 221: After the while loop would be a good place to adjust the line-height to fill up the space. This should work better than taking it from the original. Do note we'll have to remove the line-height we set up above, so that the default css specified for line height, 1em is set. That is necessary for being able to compute this. Something like:

```js
const textHeight = ele.firstElementChild.clientHeight;
const lines = textHeight / adjustedFontSize;
const newLineHeight = scrollHeight / lines;
$(ele).css({ "line-height": `${newLineHeight}px`  });
```

Then at this point, we should also set `text-align: justify`. Note in my testing I found this sometimes resulting in make the paragraph horizontally scrollable for some reason? So maybe before setting it, check if it needs to scroll horizontally, and if it doesn't also set `overflow: visible` on the paragraph.

Note we can only run this once when the line-height is 1em.

src/plugins/translate/plugin.translate.js
```diff
@@ -127,27 +154,71 @@ export class TranslatePlugin extends BookReaderPlugin {
...
           "margin-top": $(originalParagraphStyle).css("margin-top"),
           "top": $(originalParagraphStyle).css("top"),
           "height": $(originalParagraphStyle).css("height"),
-          "font-size": `${parseInt($(originalParagraphStyle).css("font-size")) - 3}px`,
           "width": $(originalParagraphStyle).css("width"),
+          "font-size": fontSize,
+          "line-height": `${lineHeightCalc}px`,
+        });
+
+        translatedParagraph.addEventListener('mousedown', (e) => {
```
cdrini, line 158: Let's drop the hover effect for now in light of our discussion with Jim; we will switch to the floating button later.

cdrini (changes requested):
Niiiice great work @schu96 ! This starting to look/feel like magic :blush:

A few tweaks to the code, and two extensions for `text-align: justify` and `line-height`.

src/plugins/translate/plugin.translate.js
```diff
@@ -156,6 +227,21 @@ export class TranslatePlugin extends BookReaderPlugin {
...
   };
 
+  /**
+   * @param {Element} ele 
+   */
+  fitVisiblePage(ele) {
+    const originalFontSize = parseInt($(ele).css("font-size"));
+    let adjustedFontSize = `${originalFontSize - 1}px`;
+      while (ele.clientHeight < ele.scrollHeight) {
+        const currentFontSize = parseInt($(ele).css("font-size"));
```
cdrini: Oh one last comment, can we avoid reading this from the HTML/parsing it, and instead use something like:

```js
let adjustFontSize = originalFontSize;
while (...) {
    adjustFontSize--;
    $(ele).css({ "font-size": `${adjustFontSize}px` });
}
```

PR #1419, opened 2025-07-28
Small translation tweaks

cdrini:
- Creates a new option to control what experiments are enabled
- Uses @pezvi 's proper icon for experiments


schu96:
Looks good to me!

PR #1422, opened 2025-08-13
Switch to internetarchive fork of bergamot package to set workerUrl

cdrini:


schu96:
@cdrini Looks good to me!

PR #1423, opened 2025-08-13
Remove some unused translations code

cdrini:


schu96:
Thanks for catching this, forgot to trim this part out after we pivoted to the Bergamot module

PR #1430, opened 2025-08-20
Translation Plugin integration with TTS plugin

schu96:
General improvements for Translate Plugin:
- Checks original work's translation language and sets that as the `from` language if metadata includes language
- Add button in translation sidebar that allows users to trigger the start of translation

Implemented functionality between Translation and TTS plugin 
- If Translation plugin is active, will automatically select the `translate to` language for TTS
- Highlight currently read paragraph element (currently does not have the ability to read line by line similarly to the original functionality)
- Re-render voices menu if translation is toggled on or off.


src/plugins/translate/plugin.translate.js
```diff
@@ -43,7 +44,16 @@ export class TranslatePlugin extends BookReaderPlugin {
...
 
+  /**
+   * @type {boolean} userToggleTranslate - Checks if user has initiated translation
+   * Should synchronize with the state of TranslationManager's active state
+   */
+  userToggleTranslate;
+
   async init() {
+    const currentLanguage = toISO6391(this.br.options.bookLanguage.replace(/[.,/#!$%^&*;:{}=\-_`~()]/g, ""));
+    this.langFromCode = currentLanguage ?? "en";
```
cdrini, line 55: This doesn't seem to be working for me here: https://deploy-preview-1430--ia-bookreader.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=romoetjuliettet00shakgoog#page/n27/mode/2up . But I tested the `toISO6391` on "French" and it worked... 🤔

cdrini: Oh it's because french is not one of the supported languages with these models!! Of course :D Lgtm!

PR #1431, opened 2025-08-20
TTS PageChunk draft

schu96:
Pushing up work so far on PageChunks for review

### DIFF
--- a/src/plugins/translate/plugin.translate.js
+++ b/src/plugins/translate/plugin.translate.js
@@ -3,6 +3,7 @@ import { html, LitElement } from 'lit';
 import { BookReaderPlugin } from '../../BookReaderPlugin.js';
 import { customElement, property } from 'lit/decorators.js';
 import { langs, TranslationManager } from "./TranslationManager.js";
+import { toISO6391 } from '../tts/utils.js';
 
 // @ts-ignore
 const BookReader = /** @type {typeof import('@/src/BookReader.js').default} */(window.BookReader);
@@ -27,10 +28,10 @@ export class TranslatePlugin extends BookReaderPlugin {
   toLanguages = [];
 
   /**
-   * Current language code that is being translated From
+   * Current language code that is being translated From. Defaults to EN currently
    * @type {!string}
    */
-  langFromCode = "en";
+  langFromCode
 
   /**
    * Current language code that is being translated To
@@ -43,7 +44,15 @@ export class TranslatePlugin extends BookReaderPlugin {
    */
   _panel;
 
+  /**
+   * @type {boolean} userToggleTranslate - Checks if user has initiated translation
+   * Should synchronize with the state of TranslationManager's active state
+   */
+  userToggleTranslate;
+
   async init() {
+    const currentLanguage = toISO6391(this.br.options.bookLanguage.replace(/[.,\/#!$%\^&\*;:{}=\-_`~()]/g, ""));
+    this.langFromCode = currentLanguage ?? "en";
 
     if (!this.options.enabled) {
       return;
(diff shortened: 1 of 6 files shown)


src/plugins/tts/AbstractTTSEngine.js
```diff
@@ -86,9 +86,8 @@ export default class AbstractTTSEngine {
 
     this._chunkIterator = new PageChunkIterator(numLeafs, leafIndex, {
       pageChunkUrl: this.opts.pageChunkUrl,
-      pageBufferSize: 5,
+      pageBufferSize: 1,
```
cdrini: Can we try bumping this up to maybe 3? Does that still work smoothly?

src/plugins/translate/plugin.translate.js
```diff
@@ -128,11 +128,8 @@ export class TranslatePlugin extends BookReaderPlugin {
     if (this.br.mode == this.br.constModeThumb || !this.userToggleTranslate) {
       return;
     }
-    if (!page.querySelector('.BRtextLayer')) {
-      console.log("there is no selection layer here", page.dataset.index);
-      return;
-    }
 
+    // this.br.trigger('translationEnabled', { }); // fires too often?
```
```suggestion
```

src/plugins/translate/plugin.translate.js
```diff
@@ -128,11 +128,8 @@ export class TranslatePlugin extends BookReaderPlugin {
     if (this.br.mode == this.br.constModeThumb || !this.userToggleTranslate) {
       return;
     }
-    if (!page.querySelector('.BRtextLayer')) {
```
cdrini, line 131: Hmm, we might have a bit of a race condition here, if the text layer isn't loaded yet. We can fix in a future PR by:

1. Change the spot below where we do `page.querySelector('.BRtextLayer')` to instead do `this.br.plugins.textSelection.getTextLayer(page)`
2. Create a new method on `TextSelectionPlugin`, `getTextLayer` that basically just calls `createTextLayer` ; that method already avoids creating if necessary
3. Move the `createDIVPageLayer` method up to ~the top of the `createTextLayer` method ; that'll ensure that it won't call multiple times if we're mid-fetching.

But ok as-is until then.

src/plugins/tts/plugin.tts.js
```diff
@@ -165,13 +165,28 @@ export class TtsPlugin extends BookReaderPlugin {
 
     const renderVoicesMenu = (voicesMenu) => {
       voicesMenu.empty();
-      const bookLanguage = this.ttsEngine.opts.bookLanguage;
+      let bookLanguage = this.ttsEngine.opts.bookLanguage;
+      let translatedLanguages;
+      let toLang;
```
cdrini: Let's move these variables into the block ~line 181; they're only used in there. Then we can also switch from `let` to `const`!

src/plugins/tts/plugin.tts.js
```diff
@@ -208,6 +225,18 @@ export class TtsPlugin extends BookReaderPlugin {
   }
 
   start(startTTSEngine = true) {
+    if (this.br.plugins.translate?.translationManager?.active) {
+      const toLangISO = toISO6391(this.br.plugins.translate.langToCode);
+      const toLanguageVoice = this.ttsEngine.getVoices().filter(v => v.lang.startsWith(toLangISO));
+      this.ttsEngine.setVoice(toLanguageVoice[0].voiceURI); // arbitrarily pick 1st choice if possible
+      const voicesMenu = this.br.refs.$BRReadAloudToolbar.find('[name=playback-voice]');
```
cdrini: This is duplicating some of the logic in the TTS `getBestVoice` method. Let's add a new parameter to `getBestVoice`. That method is pretty simple:

https://github.com/internetarchive/bookreader/blob/d99fda860c35c048b0e55b044339c5a26bd4e0b3/src/plugins/tts/AbstractTTSEngine.js#L207-L210

So if we add a new parameter to `languageOverride`, we can do something like:

```js
  /** Convenience wrapper for {@see AbstractTTSEngine.getBestVoice} */
  getBestVoice(languageOverride) {
    return AbstractTTSEngine.getBestBookVoice(this.getVoices(), this.opts.bookLanguage || languageOverride);
  }
```

cdrini (approved):
Greeeat! Lgtm!

PR #1432, opened 2025-08-30
Redesign Translation Side Menu

schu96:
Translation Plugin sidebar menu updated to using feedback from UI/UX meeting

- "From" language -> "Source" language. 
- Source language field/dropdown is not immediately available and is only visible after a user clicks on "Change"
- Add status message in sidebar if "Translate To" and "Source" languages are the same, which hints to why a translation may not work for the page.
- Misc styling

Before
----
image

After
----
image image

### DIFF
--- a/src/plugins/translate/TranslationManager.js
+++ b/src/plugins/translate/TranslationManager.js
@@ -1,6 +1,7 @@
 // @ts-check
 import { Cache } from '../../util/cache.js';
 import { BatchTranslator } from '@internetarchive/bergamot-translator/translator.js';
+import { toISO6391 } from '../tts/utils.js';
 
 export const langs = /** @type {{[lang: string]: string}} */ {
   "bg": "Bulgarian",
@@ -68,7 +69,7 @@ export class TranslationManager {
       this._initResolve = resolve;
       this._initReject = reject;
     });
-    const registryUrl = "https://cors.archive.org/cors/mozilla-translate-models/";
+    const registryUrl = "https://cors.archive.org/cors/mozilla-translate-models/firefox_models/";
     const registryJson = await fetch(registryUrl + "registry.json").then(r => r.json());
     for (const language of Object.values(registryJson)) {
       for (const file of Object.values(language)) {
@@ -97,10 +98,10 @@ export class TranslationManager {
       // List of dev models found here https://github.com/mozilla/firefox-translations-models/tree/main/models/base
       // There are also differences between the model types in the repo above here: https://github.com/mozilla/firefox-translations-models?tab=readme-ov-file#firefox-translations-models
       if (firstLang !== "en") {
-        this.fromLanguages.push({code: firstLang, name:langs[firstLang], type: "prod"});
+        this.fromLanguages.push({code: firstLang, name: toISO6391(firstLang, true), type: "prod"});
       }
       if (secondLang !== "en") {
-        this.toLanguages.push({code: secondLang, name:langs[secondLang], type: "prod"});
+        this.toLanguages.push({code: secondLang, name: toISO6391(secondLang, true), type: "prod"});
       }
     }
     this._initResolve([this.modelRegistry]);
--- a/src/plugins/translate/plugin.translate.js
+++ b/src/plugins/translate/plugin.translate.js
@@ -386,7 +386,7 @@ export class BrTranslatePanel extends LitElement {
         <label>
           <span style="font-size: 12px;color: #ccc;">Translate To</span>
           <select id="lang-to" name="to" class="lang-select" style="display:block; width:100%;" @change="${this._onLangToChange}">
-          ${sortBy(this.toLanguages, ((lang) => [...lang.name].reduce((a, char) => a + char.charCodeAt() - 96 , 0)
+          ${sortBy(this.toLanguages, ((lang) => lang.name.toLowerCase()
     )).map((lang) => {
       return html`<option value="${lang.code}" 
                         ?selected=${lang.code == this.detectedToLang}
@@ -411,7 +411,7 @@ export class BrTranslatePanel extends LitElement {
             </i> Change 
           </summary>
           <select id="lang-from" name="from" class="lang-select" value=${this.detectedFromLang} @change="${this._onLangFromChange}" style="width:65%; margin-top: 3%; margin-bottom: 3%">
-          ${sortBy(this.fromLanguages, ((lang) => [...lang.name].reduce((a, char) => a + char.charCodeAt() - 96, 0)
+          ${sortBy(this.fromLanguages, ((lang) => lang.name.toLowerCase()
           )).map((lang) => {
             return html`<option value="${lang.code}"
                       ?selected=${lang.code == this.detectedFromLang}
--- a/src/plugins/translate/plugin.translate.js
+++ b/src/plugins/translate/plugin.translate.js
@@ -53,8 +53,6 @@ export class TranslatePlugin extends BookReaderPlugin {
    */
   userToggleTranslate;
 
-  modelStatus = false;
-
   async init() {
     const currentLanguage = toISO6391(this.br.options.bookLanguage.replace(/[.,/#!$%^&*;:{}=\-_`~()]/g, ""));
     this.langFromCode = currentLanguage ?? "en";
@@ -361,7 +359,6 @@ export class BrTranslatePanel extends LitElement {
   @property({ type: Boolean }) userTranslationActive = false;
   @property({ type: String }) detectedFromLang = '';
   @property({ type: String }) detectedToLang = '';
-  @property({ type: Boolean }) modelStatus;
 
   /** @override */
   createRenderRoot() {
@@ -415,7 +412,7 @@ export class BrTranslatePanel extends LitElement {
           )).map((lang) => {
             return html`<option value="${lang.code}"
                       ?selected=${lang.code == this.detectedFromLang}
-                      >${lang.name ? lang.name : lang.code} </option`;
+                      >${lang.name ? lang.name : lang.code} </option>`;
           })
           }
           </select>
@@ -456,7 +453,7 @@ export class BrTranslatePanel extends LitElement {
   }
 
   _getSelectedLang(type) {
-    /** @type HTMLSelectElement */
+    /** @type {HTMLSelectElement} */
     const dropdown = this.querySelector(`#lang-${type}`);
     return dropdown ? dropdown.value : '';
   }
@@ -474,18 +471,6 @@ export class BrTranslatePanel extends LitElement {
     });
     this.userTranslationActive = !this.userTranslationActive;
     this.dispatchEvent(toggleTranslateEvent);
-  }
-
-  _visibilityFromLang() {
-    /** @type HTMLElement */
-    const dropdown = this.querySelector("#lang-from");
-    const currentVisibility = dropdown.style.visibility;
-    if (!currentVisibility) {
-      dropdown.style.visibility = "hidden";
-    } else {
-      dropdown.style.visibility = "";
-    }
-    return;
   }
 
   // TODO: Hardcoded warning message for now but should add more statuses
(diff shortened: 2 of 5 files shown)


src/plugins/translate/plugin.translate.js
```diff
@@ -348,31 +348,18 @@ export class BrTranslatePanel extends LitElement {
...
-        </label>
-      </div>
+    return html`<div class="app" style="margin-top: 5%">
       <div class="panel panel--to">
         <label>
-          To
-          <select id="lang-to" name="to" class="lang-select" @change="${this._onLangToChange}">
+          Translate To
+          <select id="lang-to" name="to" class="lang-select" style="display:block; width:65%;" @change="${this._onLangToChange}">
             ${this.toLanguages.map(
```
cdrini: Can we make these languages appear sorted alphabetically in both dropdowns?

src/plugins/translate/plugin.translate.js
```diff
@@ -381,9 +368,32 @@ export class BrTranslatePanel extends LitElement {
...
+          </div>
+          <select id="lang-from" name="from" class="lang-select" value=${this.detectedFromLang} @change="${this._onLangFromChange}" style="visibility:hidden; width:65%">
+            ${this.fromLanguages.map((lang) => {
+      return html`<option
+                value="${lang.code}"
+                ?selected=${lang.code == this.detectedFromLang}
+                >${lang.name}</option>`;
+    },
+    )}
+          </select>
```
cdrini: This would be a great place for a `<details>` element! We can put the whole thing in the `<summary>` tag; then clicking on "Change" or anywhere will show the option. And then we don't need any extra JS to manage the visibility!

cdrini (commented):
Niiiice looking great! A few tweaks to the style now that we're seeing it live:

<img width="366" height="248" alt="image" src="https://github.com/user-attachments/assets/e8b7f995-3cbc-489d-a28f-419e6988daff" />

Note:
- the "Translate" button switches to "Stop Translating" while active
- I don't think we need to show the warning if the languages are the same ; showing a warning in the default state can be a bit daunting! Although let's disable the "Translate" button if the languages are the same.

Here's the html from my devtools that might be a useful reference. I changed some of the HTML position and styling.

```html
<div class="app" style="margin-top: 5%;padding-right: 5px;">
      <div class="disclaimer" id="disclaimerMessage" style="background-color: rgba(255,255,255,0.1);padding: 10px;border-radius: 8px;font-size: 12px;margin-bottom: 10px;color: rgba(255,255,255, 0.9);"> <!--?lit$436990525$-->Translations are in alpha </div><div class="panel panel--to" style="padding: 0 10px;">
        <label style="font-size: 12px;color: #ccc;">
  <span style="">Translate To</span>
  
          <select id="lang-to" name="to" class="lang-select" style="display:block; width:100%;">
            <!--?lit$436990525$--><!----><option value="en"><!--?lit$436990525$-->English</option><!----><!----><option value="bg"><!--?lit$436990525$-->Bulgarian</option><!----><!----><option value="cs"><!--?lit$436990525$-->Czech</option><!----><!----><option value="de"><!--?lit$436990525$-->German</option><!----><!----><option value="es"><!--?lit$436990525$-->Spanish</option><!----><!----><option value="et"><!--?lit$436990525$-->Estonian</option><!----><!----><option value="pt"><!--?lit$436990525$-->Portuguese</option><!----><!----><option value="fa"><!--?lit$436990525$-->Persian</option><!----><!----><option value="it"><!--?lit$436990525$-->Italian</option><!----><!----><option value="ru"><!--?lit$436990525$-->Russian</option><!---->
          </select>
        </label>
      </div>
      
      <div class="panel panel--start" style="text-align: right;padding: 0 10px;/*! font-size: 18px; */margin-top: 10px;">
        <label><button class="start-translation-brn" style="/*! font-size: 14px; */">
              <!--?lit$436990525$-->Translate
            </button>
        </label>
      </div>
      
      
      
      
      
      
    <div class="panel panel--from" style="font-size: 12px;color: #ccc;text-align: center;padding: 8px 10px;">
        <label>
          <i>
            Source: <!--?lit$436990525$-->English <!--?lit$436990525$-->(detected)
          </i>
        </label>
          <div class="change-from-lang" style="text-decoration: underline white; cursor:pointer; width:fit-content; display:inline-block;">
            Change
          </div>
          <select id="lang-from" name="from" class="lang-select" style="width: 65%; visibility: hidden;" value="en">
            <!--?lit$436990525$--><!----><option value="en" selected=""><!--?lit$436990525$-->English</option><!----><!----><option value="bg"><!--?lit$436990525$-->Bulgarian</option><!----><!----><option value="cs"><!--?lit$436990525$-->Czech</option><!----><!----><option value="de"><!--?lit$436990525$-->German</option><!----><!----><option value="es"><!--?lit$436990525$-->Spanish</option><!----><!----><option value="et"><!--?lit$436990525$-->Estonian</option><!----><!----><option value="it"><!--?lit$436990525$-->Italian</option><!----><!----><option value="nb"><!--?lit$436990525$-->Norwegian Bokmål</option><!----><!----><option value="pt"><!--?lit$436990525$-->Portuguese</option><!----><!----><option value="fa"><!--?lit$436990525$-->Persian</option><!----><!----><option value="is"><!--?lit$436990525$-->Icelandic</option><!----><!----><option value="nn"><!--?lit$436990525$-->Norwegian Nynorsk</option><!----><!----><option value="ru"><!--?lit$436990525$-->Russian</option><!---->
          </select>
      </div></div>
```

src/plugins/translate/plugin.translate.js
schu96: The behavior using `<details>` and `<summary>` now means that you have to press "Change" again to hide the dropdown list. Is that alright or do I need to do something else to make the `<select>` automatically hide? 

```
        <details style="display: contents">
          <summary style="text-decoration: underline white; cursor:pointer; display:inline-block"> Change </summary>
          <select id="lang-from" name="from" class="lang-select" value=${this.detectedFromLang} @change="${this._onLangFromChange}" style="width:65%">
            ${this.fromLanguages.sort((a, b) => {
              if (!a.name || !b.name) {
                return -1
              }
              return a.name.localeCompare(b.name);
            }).map((lang) => {
      return html`<option
                value="${lang.code}"
                ?selected=${lang.code == this.detectedFromLang}
                >${lang.name ? lang.name : lang.code }</option>`;
    },
    )}
          </select>
      </details>

Comment (cdrini) [reply src/plugins/translate/plugin.translate.js]: Yep, I think that's fine :+1: Although I think putting the entire `Source: {lang} Change` in the summary might be nice ; a little easier to press.

Oh and for the sorting we can use the util.js `sortBy` helper ; that should make it a little cleaner.

File: src/plugins/translate/TranslationManager.js [hunk 3]
```diff
@@ -97,10 +98,10 @@ export class TranslationManager {
       // List of dev models found here https://github.com/mozilla/firefox-translations-models/tree/main/models/base
       // There are also differences between the model types in the repo above here: https://github.com/mozilla/firefox-translations-models?tab=readme-ov-file#firefox-translations-models
       if (firstLang !== "en") {
-        this.fromLanguages.push({code: firstLang, name:langs[firstLang], type: "prod"});
```
cdrini, line 100: I think we can delete the `langs` object now!

src/plugins/translate/plugin.translate.js
```diff
@@ -372,42 +375,54 @@ export class BrTranslatePanel extends LitElement {
...
+            return html`<option value="${lang.code}"
+                      ?selected=${lang.code == this.detectedFromLang}
+                      >${lang.name ? lang.name : lang.code} </option`;
+          })
+          }
+          </select>
+      </details>
+      <div class="footer" id="status" style="margin-top:5%">
+      ${this._statusWarning()}
+      </div>
```
cdrini, line 422: I don't think we need to show this warning. It's a bit daunting to be presented with a warning as soon as you open the panel! Since English will likely be the default. Instead, let's when "Translate" is clicked, show a `window.alert` with this error message.

src/plugins/translate/plugin.translate.js
```diff
@@ -372,42 +375,54 @@ export class BrTranslatePanel extends LitElement {
...
+            <i>
+              Source: ${this._getLangName(this.detectedFromLang)} ${this.prevSelectedLang ? "" : "(detected)"} 
+            </i> Change 
+          </summary>
+          <select id="lang-from" name="from" class="lang-select" value=${this.detectedFromLang} @change="${this._onLangFromChange}" style="width:65%; margin-top: 3%; margin-bottom: 3%">
+          ${sortBy(this.fromLanguages, ((lang) => lang.name.toLowerCase()
+          )).map((lang) => {
+            return html`<option value="${lang.code}"
+                      ?selected=${lang.code == this.detectedFromLang}
+                      >${lang.name ? lang.name : lang.code} </option`;
```
```suggestion
                      >${lang.name ? lang.name : lang.code} </option>`;
```

src/plugins/translate/plugin.translate.js
```diff
@@ -436,30 +452,11 @@ export class BrTranslatePanel extends LitElement {
...
-    // Update the "To" dropdown to the previous language
-    const toDropdown = this.querySelector('#lang-to');
-    if (toDropdown) {
-      toDropdown.value = prevLang;
-    }
+    this.detectedToLang = event.target.value;
   }
 
   _getSelectedLang(type) {
+    /** @type HTMLSelectElement */
```
```suggestion
    /** @type {HTMLSelectElement} */
```

src/plugins/tts/utils.js
```diff
@@ -19,38 +19,48 @@ export function isAndroid(userAgent = navigator.userAgent) {
   return /android/i.test(userAgent);
 }
 
+/** @type {{[lang: string]: string}} */
+// Handle odd one-off language pairs that might show up over time
+const specialLangs =  {
```
cdrini, line 24: A better name for this might be `languageNameOverrides` 😁

src/plugins/tts/utils.js
```diff
@@ -19,38 +19,48 @@ export function isAndroid(userAgent = navigator.userAgent) {
...
+ * Can also retrieve language native name + ISO 639-1 code
  **/
 
 /**
  * @param {string} language in some format
+ * @param {boolean?} getName gets Native Name + language code if true
  * @return {ISO6391?}
  */
-export function toISO6391(language) {
+export function toISO6391(language, getName = false) {
```
cdrini, line 38: This API design is a bit confusing ; it's a bit unexpected a method called `toISO63961` would have an option called `getName`; that seems like a separate method/intent which contradicts the old name.

A cleaner design would be to:

1. Change `searchForISO6391` to `findLanguageByField`, and then have it return the full lang object, not just the iso field. This method is only used by `toISO6391`
2. Change `toISO6391` to `findLanguage`, and likewise have it return the full lang object.
3. Create a new method `toISO639` that basically just does `findLanguage(language)?.iso639_1`. Note it might make sense to dissolve this method entirely, since it's now so simple 😁 But to avoid having to update all the references, this is easier.
4. Then create a new method `findLanguageNativeName(language)` that does basically the same thing, but also takes into account the overrides.

src/plugins/translate/plugin.translate.js
```diff
@@ -358,7 +360,8 @@ export class BrTranslatePanel extends LitElement {
   @property({ type: String }) disclaimerMessage = '';
   @property({ type: Boolean }) userTranslationActive = false;
   @property({ type: String }) detectedFromLang = '';
-
+  @property({ type: String }) detectedToLang = '';
+  @property({ type: Boolean }) modelStatus;
```
cdrini: This appears unused?
```suggestion
```

src/plugins/translate/plugin.translate.js
```diff
@@ -52,6 +53,8 @@ export class TranslatePlugin extends BookReaderPlugin {
    */
   userToggleTranslate;
 
+  modelStatus = false;
+
```
```suggestion
```

src/plugins/translate/plugin.translate.js
```diff
@@ -478,5 +475,25 @@ export class BrTranslatePanel extends LitElement {
...
+    /** @type HTMLElement */
+    const dropdown = this.querySelector("#lang-from");
+    const currentVisibility = dropdown.style.visibility;
+    if (!currentVisibility) {
+      dropdown.style.visibility = "hidden";
+    } else {
+      dropdown.style.visibility = "";
+    }
+    return;
+  }
```
cdrini: No longer needed!

```suggestion
```

PR #1433, opened 2025-09-03
Fix blend mode in ms edge

cdrini:
Fix Edge tts/translations blending when they shouldn't.


cdrini:
Auto-merging; small mostly CSS-only change.

PR #1435, opened 2025-09-08
Fix header footer paragraphs from being read aloud

schu96:
Checks for `x-role="ocr-role-header-footer"` from text selection's djvuXML and assigns new class to translation paragraphs to prevent headers/footers from being assigned a PageChunk for ReadAloud functionality.


cdrini (approved):
Woohoo works like a charm!

PR #1437, opened 2025-09-24
Show loading indicator when downloading language models

schu96:
Added a visual indicator to let users know that the translation plugin is actively attempting to download the language models from IA servers

- Use `getTranslationModel` method within `TranslatorBacking` class to show a loading indicator when the translation plugin is enabled and a language model begins downloading. 

- Same loading indicator as seen in `search-results.js`

### DIFF
--- a/src/plugins/translate/TranslationManager.js
+++ b/src/plugins/translate/TranslationManager.js
@@ -156,8 +156,7 @@ export class TranslationManager {
   getTranslationModel = async(fromLanguage, toLanguage) => {
     return this.translator.backing.getTranslationModel({from: fromLanguage, to: toLanguage})
       .catch((err) => {
-      console.error(`An error occurred while trying to retreive the ${fromLanguage}-${toLanguage} model`);
-    })
-    ;
+        console.error(`An error occurred while trying to retreive the ${fromLanguage}-${toLanguage} model`);
+      });
   }
 }


src/plugins/translate/plugin.translate.js
```diff
@@ -485,5 +506,18 @@ export class BrTranslatePanel extends LitElement {
     }
     return "";
   }
+
+  _languageModelStatus() {
+    if (this.userTranslationActive) {
+      if (!this.langDownloadCompleted) {
+        return html`
+          <p>Downloading language model</p>
+          <ia-activity-indicator mode="processing" style="display:block; width: 40px; height: 40px; margin: 0 auto;"></ia-activity-indicator>
```
cdrini: For consistency with where we use this elsewhere
```suggestion
          <ia-activity-indicator mode="processing" style="display:block; width: 40px; height: 40px; margin: 0 auto;"></ia-activity-indicator>
          <p>Downloading language model</p>
```

src/plugins/translate/TranslationManager.js
```diff
@@ -167,4 +167,17 @@ export class TranslationManager {
...
     return promise;
   }
+
+  /**
+   * Checks and updates the status of a model that is currently being retreived from IA
+   * @param {string} fromLanguage
+   * @param {string} toLanguage
+   * @returns {Promise<boolean>}
+   */
+  checkModels = async(fromLanguage, toLanguage) => {
```
cdrini: The naming here could be a bit better; maybe `getTranslationModel` is the simplest, and have it return the model itself. Then this becomes a simple facade method.

src/plugins/translate/plugin.translate.js
```diff
@@ -198,6 +205,10 @@ export class TranslatePlugin extends BookReaderPlugin {
 
       if (paragraph.textContent.length !== 0) {
         const pagePriority = parseFloat(pageIndex) + priority + pidx;
+        this.translationManager.checkModels(this.langFromCode, this.langToCode).then((resp) => {
+          this.langDownloadCompleted = resp;
+          this._render();
```
cdrini: Oh I just realized we call `_render()` in a few spots; we shouldn't re-render the entire panel; that'll cause all the panel's internal state to reset. We have access to the panel via `this._panel`, and since it's a reactive lit component, all we have to do is eg `this._panel.langDownloadCompleted = resp`, and it'll be updated accordingly.

src/plugins/translate/plugin.translate.js
```diff
@@ -53,6 +54,12 @@ export class TranslatePlugin extends BookReaderPlugin {
    */
   userToggleTranslate;
 
+  /**
+   * @type {boolean} langDownloadCompleted - Updates to true when language
+   * model pairs are successfully downloaded from IA
+   */
+  langDownloadCompleted = false;
```
cdrini: Then actually I don't think we need this on the plugin at all, it can exist only on the panel.

src/plugins/translate/plugin.translate.js
```diff
@@ -364,6 +380,7 @@ export class BrTranslatePanel extends LitElement {
   @property({ type: Boolean }) userTranslationActive = false;
   @property({ type: String }) detectedFromLang = '';
   @property({ type: String }) detectedToLang = '';
+  @property({ type: Boolean }) langDownloadCompleted;
```
cdrini: We generally track whether it's loading instead of whether it's completed as a pattern, eg

```suggestion
  @property({ type: Boolean }) loadingModel = false;
```

cdrini (changes requested):
Niiice! The new loading indicator looks great! One small UX tweak, and a few small code improvements.

src/plugins/translate/TranslationManager.js
```diff
@@ -167,4 +146,17 @@ export class TranslationManager {
...
+   * Checks and updates the status of a model that is currently being retreived from IA
+   * @param {string} fromLanguage
+   * @param {string} toLanguage
+   * @returns {Promise<boolean>}
+   */
+  getTranslationModel = async(fromLanguage, toLanguage) => {
+    return this.translator.backing.getTranslationModel({from: fromLanguage, to: toLanguage}).then(() => {
+      // TODO check what happens if the download fails somehow (i.e. invalid SHA hash, download connection failed, etc)
+      return true;
+    });
```
cdrini: This can return the model directly; eg `return this.translator.backing.getTranslationModel({from: fromLanguage, to: toLanguage})`;

src/plugins/translate/plugin.translate.js
```diff
@@ -293,13 +298,17 @@ export class TranslatePlugin extends BookReaderPlugin {
...
+        code: selectedLangFrom,
+        name: this.translationManager.fromLanguages.find((entry) => entry.code == selectedLangFrom).name,
+      });
     }
 
     // Update the 'To' languages list and set the correct 'To' language
     this._panel.toLanguages = this.translationManager.toLanguages;
 
     console.log(this.langFromCode, this.langToCode);
+    this._render();
```
cdrini, line 318: We should avoid calling `_render`, since that will cause the panel to be rebuilt from scratch. This will cause bugs where its state isn't handled correctly. You can see eg after you hit "translate", and then close/open the panel, the "Stop translate" button is reset and has the wrong state! I think this might be due to the spots where we're calling `_render()`.

It seems like we're already setting the `toLanguages` property on the panel ; can this line be dropped entirely?

```suggestion
```

src/plugins/translate/plugin.translate.js
```diff
@@ -308,6 +317,7 @@ export class TranslatePlugin extends BookReaderPlugin {
   handleToLangChange = async (e) => {
     this.clearAllTranslations();
     this.langToCode = e.detail.value;
+    this._render();
```
cdrini, line 327: Ditto here. Is this necessary? If you're having issues where it's not responding, you might be able to call `.requestUpdate()` on the `_panel`. See https://lit.dev/docs/components/lifecycle/#reactive-update-cycle

src/plugins/translate/plugin.translate.js
```diff
@@ -485,5 +502,19 @@ export class BrTranslatePanel extends LitElement {
...
+        <p>Downloading language model</p>
+        `;
+      }
+      return html`<p>Language model loaded</p>`;
+    }
+    return "";
+  }
 }
 
+// TODO - Shqip (sq) breaks?
```
cdrini: Let's create an issue for this; how is this one breaking?

src/plugins/translate/plugin.translate.js
```diff
@@ -198,6 +200,9 @@ export class TranslatePlugin extends BookReaderPlugin {
 
       if (paragraph.textContent.length !== 0) {
         const pagePriority = parseFloat(pageIndex) + priority + pidx;
+        this.translationManager.getTranslationModel(this.langFromCode, this.langToCode).then((resp) => {
+          this._panel.loadingModel = resp;
```
cdrini: This feels flipped; `loadingModel` should be true _only while_ the model is loading. So when the promise resolves, the model should be finished. eg 

```js
this._panel.loadingModel = true;
this.translationManager.getTranslationModel(this.langFromCode, this.langToCode).then((resp) => {
    this._panel.loadingModel = false;
});
```

cdrini (changes requested):
Nice! A few more small fixes, and a question around the `_render()` calls. That'll cause a few headaches/subtle bugs, so would be nice to find a way to avoid it!

src/plugins/translate/TranslationManager.js
schu96: Is there a reason why we need the model back when we call this?

src/plugins/translate/plugin.translate.js
schu96: Oh I fixed this issue but forgot to remove this note for myself :sweat_smile: 

We have some older code that tries to add a language to the From / To array if it doesn't have a counterpart in the To / From array but it was referring to our old dictionary of supported languages

src/plugins/translate/TranslationManager.js
cdrini: It makes the method name more in sync with its result, and also make it a bit simpler!

src/plugins/translate/plugin.translate.js
schu96: Not sure what the issue is, I noticed the exact issue with the side menu values not persisting which is why I was using `this._render()` for the to/from languages. 

I have tried changing to using `this._panel.requestUpdate()` but it doesn't seem to work as intended. Without calling `this._render()` it doesn't seem to keep the changes even though they're changing within the Translate plugin itself.

cdrini: Ahhh, I think I see the issue! The issue is the panel isn't keeping track of what's selected. It's kind of combining two different concepts, `selectedFromLang` and `detectedFromLang`. These should likely be separate properties; having them as one is what's causing this to reset every time the panel is re-opened.

cdrini: We debugged on a call, and it seems like the menu slider is reconstructing each component anyways -_- and because we have this inside the template literal, it is _copying_ any primitive values at the very beginning. So ok with calling `_render()` for now.

src/plugins/translate/TranslationManager.js
```diff
@@ -167,4 +146,19 @@ export class TranslationManager {
...
+  /**
+   * Checks and updates the status of a model that is currently being retreived from IA
+   * @param {string} fromLanguage
+   * @param {string} toLanguage
+   * @returns {Promise<boolean>}
+   */
+  getTranslationModel = async(fromLanguage, toLanguage) => {
+    return this.translator.backing.getTranslationModel({from: fromLanguage, to: toLanguage}).then((resp) => {
+      return resp;
+    }).catch((err) => {
```
cdrini:
```suggestion
    return this.translator.backing.getTranslationModel({from: fromLanguage, to: toLanguage})
      .catch((err) => {
```

src/plugins/translate/plugin.translate.js
```diff
@@ -290,16 +301,22 @@ export class TranslatePlugin extends BookReaderPlugin {
 
     // Update the from language
     this.langFromCode = selectedLangFrom;
+    console.log("this is langFromCode", this.langFromCode, selectedLangFrom);
```
```suggestion
```

src/plugins/translate/TranslationManager.js
```diff
@@ -167,4 +146,18 @@ export class TranslationManager {
...
+   * @param {string} fromLanguage
+   * @param {string} toLanguage
+   * @returns {Promise<boolean>}
+   */
+  getTranslationModel = async(fromLanguage, toLanguage) => {
+    return this.translator.backing.getTranslationModel({from: fromLanguage, to: toLanguage})
+      .catch((err) => {
+      console.error(`An error occurred while trying to retreive the ${fromLanguage}-${toLanguage} model`);
+    })
+    ;
```
cdrini:
```suggestion
      .catch((err) => {
        console.error(`An error occurred while trying to retreive the ${fromLanguage}-${toLanguage} model`);
      });
```

PR #1439, opened 2025-09-26
Move the silent mp3 out of the code into separate file

cdrini:
Loading media via data may be blocked by CSP, so put the mp3 into its own file.


bfalling (approved):
Seem fine!

PR #1441, opened 2025-10-09
Add dockable slider, left/right buttons

pezvi:


src/css/_BRnav.scss
```diff
@@ -325,3 +329,52 @@
...
+
+.BRnavMobile {
+  box-sizing: content-box;
+}
+
+.BRnavMobile.docked {
+  height: 0px;
+  border: none;
+  .BRcontrols .scrubber .BRpager.ui-slider {
+    top: 0px;
```
cdrini: It seems like it's getting partially obstructed by the rest of the toolbar ; setting this to `-1px` makes it actually have a visible height of `2px` as stated below.

```suggestion
    top: -1px;
```

cdrini (approved):
Niiiice this looks fantastic Jim! I tested on FF/Chrome+Windows, Chrome+Android, iOS (BrowserStack), across a few view modes/searches/TOCs, and everything worked smoothly! Although I think the 1px height is visually inaccessible to me ; can we make that 2px? I note in the CSS it has a height of 2px; I think it's the `top` that's causing it to be `1px`.

Otherwise lgtm!

PR #1454, opened 2025-11-26
Improve accessibility of search inside panel

cdrini:
Fixes:

* Search results now tab/keyboard-focussable buttons
* "Jump to last result" and "Jump to first result" skip links for easy non-screenreader keyboard navigation
* Attach the label help text correctly to the search input
* The loading indicator is now correctly invisible to screen readers
* Verbal announcement when search results start loading
* Verbal announcement when search results are done
* Label the search results with the book's language to allow the screenreader to use the correct voice
* Add hover state for the search result buttons

Other:

* Remove an unused option to display a cover in order to simplify the CSS (no longer need grid)

### DIFF
--- a/src/BookNavigator/search/search-results.js
+++ b/src/BookNavigator/search/search-results.js
@@ -342,14 +342,22 @@ export class IABookSearchResults extends LitElement {
 
       .result-item {
         cursor: pointer;
-        margin-top: 12px;
+        margin-top: 8px;
         font-size: 12px;
+        padding: 5px;
+        border-radius: 4px;
+        display: block;
+        width: 100%;
         /* Reset button styles */
         background: none;
         border: none;
-        padding: 0;
         text-align: left;
         font-family: inherit;
+        transition: background-color 0.2s;
+      }
+
+      .result-item:hover {
+        background-color: rgba(255, 255, 255, 0.1);
       }
 
       .loading {


src/BookNavigator/search/search-results.js
```diff
@@ -129,44 +127,45 @@ export class IABookSearchResults extends LitElement {
...
-      <ul class="results ${resultsClass}">
+    <nav aria-label="Search results">
+      <div>
+        ${this.resultsCount}
+        <a
+          href="#"
+          class="skip-link"
+          @click=${(e) => {
+      e.preventDefault();
+      this.shadowRoot.querySelector('.results li:last-child .result-item').focus();
```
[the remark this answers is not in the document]
cdrini: It's only rendered if there are more than 0 results

src/BookNavigator/search/search-results.js
```diff
@@ -176,21 +176,29 @@ export class IABookSearchResults extends LitElement {
...
-  }
-
   render() {
     const showSearchCTA = (!this.queryInProgress && !this.errorMessage)
     && (!this.queryInProgress && !this.results.length);
     return html`
       ${this.headerSection}
-      ${this.searchForm}
-      <div class="results-container">
+      <form action="" method="get" @submit=${this.performSearch}>
```
latonv, line 184: Q: Does `action=""` do something that just omitting the attribute doesn't?

src/BookNavigator/search/search-results.js
```diff
@@ -176,21 +176,29 @@ export class IABookSearchResults extends LitElement {
...
+        ${this.searchMultipleControls}
+        <input
+          type="search"
+          name="query"
+          id="br-search-input"
+          @keyup=${this.setQuery}
+          @search=${this.setQuery}
```
latonv, line 191: It seems like the `search` event is non-standard and not supported in FF/Safari. Should this also be listening for `change` events?

src/BookNavigator/search/search-results.js
cdrini: Interesting good question... I'm not too sure! This was there before, I just copied it over ; so opting to let sleeping dogs lie for now!

src/BookNavigator/search/search-results.js
cdrini: I thought that event looked weird :P This was also there before, so going to opt to leave as-is as well, in case this is a Chesterton's Fence!

PR #1463, opened 2025-12-07
Refactor ISO lang code utils functions

schu96:
Refactored some util functionality using the feedback from @cdrini 
https://github.com/internetarchive/bookreader/pull/1432#discussion_r2352863213

`findLanguage` is a function that returns a `Code` object when an input language is matched by its name, native name, or from different iso639-1 abbreviations. This `Code` object can then be used to find the specific language name field via `toISO6391` or `toNativeName`

### DIFF
--- a/src/plugins/tts/utils.js
+++ b/src/plugins/tts/utils.js
@@ -1,3 +1,4 @@
+// @ts-check
 import langs from 'iso-language-codes';
 
 /**
@@ -40,7 +41,11 @@ export function toISO6391(language) {
   if (specialLangs[language]) {
     return language;
   }
-  const codeObj = findLanguage(language, 'iso639_1') || findLanguage(language, 'iso639_2T') || findLanguage(language, 'iso639_2B');
+  const codeObj = (
+    findLanguage(language, ['iso639_1']) ||
+    findLanguage(language, ['iso639_2T']) ||
+    findLanguage(language, ['iso639_2B', 'name', 'nativeName'])
+  );
   if (codeObj) return codeObj.iso639_1;
   return null;
 }
@@ -56,7 +61,11 @@ export function toNativeName(language) {
   if (specialLangs[language]) {
     return specialLangs[language];
   }
-  const codeObj = findLanguage(language, 'iso639_1') || findLanguage(language, 'iso639_2T') || findLanguage(language, 'iso639_2B');
+  const codeObj = (
+    findLanguage(language, ['iso639_1']) ||
+    findLanguage(language, ['iso639_2T']) ||
+    findLanguage(language, ['iso639_2B', 'name', 'nativeName'])
+  );
   if (codeObj?.nativeName) return codeObj.nativeName.split(", ")[0];
   return null;
 }
@@ -65,18 +74,17 @@ export function toNativeName(language) {
 
 /**
  * @param {string} language
+ * @param {Array<'iso639_1' | 'iso639_2T' | 'iso639_2B' | 'nativeName' | 'name'>} fields
  * @returns {Code | null}
  */
-function findLanguage(language, codeType) {
+function findLanguage(language, fields) {
   if (!language) return null;
   language = language.toLowerCase();
   for (const lang of langs) {
-    if (lang[codeType].toLowerCase().split(", ").includes(language)) {
-      return lang;
-    } else if (lang.name.toLowerCase().split(", ").includes(language)) {
-      return lang;
-    } else if (lang.nativeName.toLowerCase().split(", ").includes(language)) {
-      return lang;
+    for (const codeType of fields) {
+      if (lang[codeType]?.toLowerCase().split(", ").includes(language)) {
+        return lang;
+      }
     }
   }
   return null;


src/plugins/tts/utils.js
```diff
@@ -24,43 +24,62 @@ export function isAndroid(userAgent = navigator.userAgent) {
...
+    return language;
+  }
+  const langCode = findLanguage(language);
+  if (langCode) return langCode.iso639_1;
+  return null;
+}
 
-  return searchForISO6391(language, ['iso639_1'], getName) ||
-    searchForISO6391(language, ['iso639_2T'], getName) ||
-    searchForISO6391(language, ['iso639_2B', 'nativeName', 'name'], getName);
```
cdrini, line 44: I think we want to keep this logic here ; this sets a priority for which matches first in the case where something can match multiple things. I can't recall for certain, but I think there are a few cases where the same language code appears multiple times for different languages.

src/plugins/tts/utils.js
```diff
@@ -24,43 +24,62 @@ export function isAndroid(userAgent = navigator.userAgent) {
...
+function findLanguage(language) {
+  if (!language) return null;
+  language = language.toLocaleLowerCase();
   for (const lang of langs) {
-    for (const colName of columnsToSearch) {
-      if (lang[colName].split(', ').map(x => x.toLowerCase()).indexOf(language) != -1) {
-        if (getName) return `${lang.nativeName.split(", ")[0]} (${lang.iso639_1})`;
-        return lang.iso639_1;
-      }
+    if (lang.iso639_1 == language || lang.iso639_2B == language || lang.iso639_2T == language) {
```
cdrini: Are we certain these fields will already be lowercased?

src/plugins/tts/utils.js
```diff
@@ -24,43 +24,62 @@ export function isAndroid(userAgent = navigator.userAgent) {
...
  * @param {string} language
- * @param {Array<keyof import('iso-language-codes').Code>} columnsToSearch
- * @param {boolean?} getName
- * @return {ISO6391?}
+ * @returns {Code | null}
  */
-function searchForISO6391(language, columnsToSearch, getName) {
+function findLanguage(language) {
+  if (!language) return null;
+  language = language.toLocaleLowerCase();
```
cdrini: We likely want `.toLowerCase` ; to keep it consistent with the method we call below. Otherwise they might not match.

src/plugins/tts/utils.js
```diff
@@ -24,43 +24,62 @@ export function isAndroid(userAgent = navigator.userAgent) {
...
-      }
+    if (lang.iso639_1 == language || lang.iso639_2B == language || lang.iso639_2T == language) {
+      return lang;
+    } else if (lang.name.toLowerCase().split(", ").includes(language)) {
+      return lang;
+    } else if (lang.nativeName.toLowerCase().split(", ").includes(language)) {
+      return lang;
     }
   }
-  if (specialLangs[language]) return `${specialLangs[language]} (${language})`;
```
cdrini, line 63: Do we no longer need this?

src/plugins/tts/utils.js
schu96: Yes, I checked the iso-language-codes file that we reference but if you'd prefer that we are 100% sure of this I can add toLowerCase() for some more peace of mind

src/plugins/tts/utils.js
schu96: We shouldn't need it anymore, there was a small change in the TranslationManager file that makes sure that it is formatted similarly to the rest of the languages 

`Español (es)`
`中文 (Zhōngwén) (zh-hans)`

src/plugins/tts/utils.js
```diff
@@ -24,43 +24,61 @@ export function isAndroid(userAgent = navigator.userAgent) {
...
-      if (lang[colName].split(', ').map(x => x.toLowerCase()).indexOf(language) != -1) {
-        if (getName) return `${lang.nativeName.split(", ")[0]} (${lang.iso639_1})`;
-        return lang.iso639_1;
-      }
+    if (lang[codeType].toLowerCase().split(", ").includes(language)) {
+      return lang;
+    } else if (lang.name.toLowerCase().split(", ").includes(language)) {
+      return lang;
+    } else if (lang.nativeName.toLowerCase().split(", ").includes(language)) {
+      return lang;
```
cdrini: This makes the signature of this method a little confusing ; it's exposing an interface like `findLanguage(query, field)`, but then secretly also checking the `name` and `nativeName` field. I think something similar to what was there before, where we iterate overall the requested field names, might be a clearer interface.

src/plugins/tts/utils.js
```diff
@@ -24,43 +24,61 @@ export function isAndroid(userAgent = navigator.userAgent) {
...
 
+/** @typedef {import('iso-language-codes').Code} Code */
+
 /**
- * Searches for the given long in the given columns.
  * @param {string} language
- * @param {Array<keyof import('iso-language-codes').Code>} columnsToSearch
- * @param {boolean?} getName
- * @return {ISO6391?}
+ * @returns {Code | null}
```
cdrini, line 78:
```suggestion
 * @param {'iso639_1' | 'iso639_2T' | 'iso639_2B' | 'nativeName' | 'name'} codeType
 * @returns {Code | null}
```

cdrini (approved):
Sweeet nice! The interface look much cleaner now! And the unit test are great!

PR #1472, opened 2025-12-22
Switch to npm trusted publishing + Update to Node 24

cdrini:
Needed a new version of node to support npm's [trusted publishers](https://docs.npmjs.com/trusted-publishers).


cdrini:
Auto-merging to test the new flow. Discussed with rest of team, and everyone's on board to update to node 24.

PR #1477, opened 2026-01-03
Fix CSS scaling issue in Chrome for Translation plugin

schu96:
Makes CSS property assignments for translation text layers before class name `showingTranslation` is added

Includes a small CSS change that hides the `BRtranslateLayer` elements during the page flip animation

### DIFF
--- a/src/css/_BRpages.scss
+++ b/src/css/_BRpages.scss
@@ -101,7 +101,7 @@ svg.BRPageLayer {
 }
 
 // Hides page layers during page flip animation
-.BRpageFlipping .BRtextLayer, .BRpageFlipping .BRtranslateLayer {
+.BRpageFlipping .BRtextLayer {
   display: none;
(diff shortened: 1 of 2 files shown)


src/css/_BRpages.scss
```diff
@@ -101,7 +101,7 @@ svg.BRPageLayer {
 }
 
 // Hides page layers during page flip animation
-.BRpageFlipping .BRtextLayer {
+.BRpageFlipping .BRtextLayer, .BRpageFlipping .BRtranslateLayer {
```
cdrini: This is done for the textlayer because it contains so many DOM elements (1 for every word :P) that it makes the animation slow/janky. We might be ok with the translate layer ; and it creates a more immersive experience/illusion of the translated book!

```suggestion
.BRpageFlipping .BRtextLayer {
```

PR #1479, opened 2026-01-07
Fix gitignore to only ignore root-level BookReader dir

cdrini:


cdrini:
Auto-merging ; small dev fix

PR #1480, opened 2026-01-08
More accessibility improvements

cdrini:
QA: https://deploy-preview-1480--ia-bookreader.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala#page/198/mode/2up

Fixes 3 points from the audit:

* Inappropriate label "Toggle fullscreen" is specified for the "Enter fullscreen / Exit fullscreen" button
* On activating "Flip left" and "Flip right" buttons present in the "Bookreader Item Preview", the current page and the total number of pages gets updated visually but not aurally
* The slider is not labelled with `role=slider`
* Sidebar buttons missing `aria-expanded` ; will include once it's merged: https://github.com/internetarchive/iaux-item-navigator/pull/24

And a few other small points:
* Screenreaders were having difficulty reading the text layer, because they were ignoring the elements that contained the spaces! Implemented a bit of hack there, but it worked well in NVDA; would be nice to test JAWS.

A few considerations:

* I had to rejigger the slider quite a bit to make it report updates at the correct time and avoid some odd behaviour. This:
    * Fixed an issue where if you focussed the slider and pressed/held right, it would jitter and behave oddly, but
    * ~Introduce an odd experience where the keyboard shortcuts when the slider has focus _are different_ than the shortcuts when it doesn't. Without focus, left/right goes back/forward, and up/down goes back/forward, and pgup/pgdown move by a single page. But since the slider is now handling its keyboard shortcuts, up/down _move in the opposite direction_, and pgup/pgdown move in jumps. Note sure which is better here, but I think it could be confusing if a user click the slider, is then given focus, and then the keyboard shortcuts change.~
         * Discussed with @pezvi , and we decided removing up/down to flip was the best option. Folks can still use left/right of course! And this also frees up/down for scrolling across all modes.
* In order to get the page change announcement working, I did a bigger rejiggering of the `jumpToIndex` method to take an optional `ariaLive` parameter so that we can decide in what contexts we do indeed want the new page to be announced. Otherwise the page change was announced/interrupted the read aloud experience, and also was duplicated when using the slider. This resulted in a bigger refactor with a BREAKING CHANGE to the `jumpToIndex` method's interface, from:

`BookReader.jumpToIndex(index: PageIndex, pageX = 0, pageY = 0, noAnimate = false)` to `BookReader.jumpToIndex(indexOrDirection: PageIndex | 'left' | 'right' | 'next' | 'prev', {pageX = 0, pageY = 0, noAnimate = false, flipSpeed = null, ariaLive = false} = {}`

These other parameters weren't really functional / used, so I don't think it'll break any usecases.

### DIFF
--- a/src/BookReader.js
+++ b/src/BookReader.js
@@ -876,18 +876,14 @@ BookReader.prototype.setupKeyListeners = function () {
       e.preventDefault();
       this.last();
       break;
-    case "ArrowDown":
     case "PageDown":
-    case "Down": // hack for IE and old Gecko
       // In 1up and thumb mode page scrolling handled by browser
       if (this.constMode2up === this.mode) {
         e.preventDefault();
         this.next();
       }
       break;
-    case "ArrowUp":
     case "PageUp":
-    case "Up": // hack for IE and old Gecko
       // In 1up and thumb mode page scrolling handled by browser
       if (this.constMode2up === this.mode) {
         e.preventDefault();
--- a/src/BookReader/Navbar/Navbar.js
+++ b/src/BookReader/Navbar/Navbar.js
@@ -328,6 +328,23 @@ export class Navbar {
       'aria-valuemax': br.book.getNumLeafs(),
     });
 
+    // Ignore up/down arrow keys and page up/down keys, since they're confusingly different
+    // between slider movement and page movement. Down decreases slider value, which would move
+    // the scroll _up_ in 1up.
+    $sliders.find('.ui-slider-handle').off('keydown').on('keydown', function (event) {
+      switch (event.keyCode || event.which) {
+      case $.ui.keyCode.UP:
+      case $.ui.keyCode.DOWN:
+      case $.ui.keyCode.PAGE_UP:
+      case $.ui.keyCode.PAGE_DOWN:
+        return;
+      default:
+        // Forward along to the default handler for other keys
+        $.ui.slider.prototype._handleEvents.keydown.call($(this).parent().data('ui-slider'), event);
+        return;
+      }
+    });
+
     $sliders.on('slide', (event, ui) => this.updateNavPageNum(ui.value));
 
     $sliders.on('slidechange', (event, ui) => {
--- a/tests/jest/BookReader.keyboard.test.js
+++ b/tests/jest/BookReader.keyboard.test.js
@@ -73,16 +73,6 @@ describe('Keyboard shortcuts', () => {
     expect(keyEvent.preventDefault).toHaveBeenCalledTimes(1);
   });
 
-  test('ArrowDown key', () => {
-    br.mode = br.constMode2up;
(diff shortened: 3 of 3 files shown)


src/BookReader.js
```diff
@@ -1145,27 +1143,35 @@ BookReader.prototype._isIndexDisplayed = function(index) {
...
+ * @param {boolean} [options.ariaLive]
+ * @param {boolean} [options.triggerStop] - whether to trigger the stop event; default true; maybe deprecated?
  */
-BookReader.prototype.jumpToIndex = function(index, pageX, pageY, noAnimate) {
-  // Don't jump into specific unviewable page
-  const page = this.book.getPage(index);
+BookReader.prototype.jumpToIndex = function(indexOrDirection, {pageX = 0, pageY = 0, noAnimate = false, flipSpeed = null, ariaLive = false, triggerStop = true} = {}) {
+  const page = this.activeMode.parsePageSpecifier(indexOrDirection);
+  flipSpeed = utils.parseAnimationSpeed(flipSpeed) || this.flipSpeed;
+  if (!page || page.index == this.currentIndex()) {
```
bfalling, line 1158: SUGGESTION: Prefer `===`?

src/BookReader/Mode2Up.js
```diff
@@ -70,14 +72,27 @@ export class Mode2Up {
...
     this.mode2UpLit.jumpToIndex(index);
   }
 
+  /**
+   * @override
+   * @param {PageIndex | 'left' | 'right' | 'next' | 'prev'} indexOrDirection
+   * @return {import('./BookModel.js').PageModel}
+   */
+  parsePageSpecifier(indexOrDirection) {
+    const isRTL = this.book.pageProgression == 'rl';
```
bfalling, line 90: SUGGESTION: Prefer `===`?

src/BookReader/utils.js
```diff
@@ -55,7 +55,7 @@ export function getActiveElement(doc = document, recurseShadowDom = true) {
 /** Check if an input field/textarea is active. Also checks shadow DOMs. */
 export function isInputActive(doc = document) {
   const activeEl = getActiveElement(doc);
-  return activeEl?.tagName == "INPUT" || activeEl?.tagName == "TEXTAREA";
+  return activeEl?.tagName == "INPUT" || activeEl?.tagName == "TEXTAREA" || activeEl?.getAttribute('role') == 'slider';
```
bfalling, line 58: SUGGESTION: Prefer `===`?

src/BookReader.js
cdrini: Thanks for the CR @bfalling ! Opting out of this one; since the fields are backed by types, the `==`/`===` distinction is less critical. I'm also a fan of having things like that be enforced by the linter! We can add a rule there to force it if we wish.

PR #1481, opened 2026-01-09
Auto-hiding BookReader Chrome

cdrini:
In order to allow for better usage of screen real-estate, especially on mobile, make the footer toolbar and the minimized side menu disappear if not used in a few seconds. They will re-appear if:

1. The user taps/clicks somewhere -- 2up has been adjusted so that only the outer 2/3s of the page will trigger a flip, allowing the centre/spine region to show the controls
3. They scroll up
5. They exit full screen -- at which point they persist as they usually do

Note it will likely not work with the lending bar yet ; once that is wired up, we can take it out of experiment.

To QA:

1. Go to https://deploy-preview-1481--ia-bookreader.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala#page/viii/mode/2up
2. Chrome should *not* hide in the default view
3. Go full screen
4. Chrome should now auto-hide!

Depends on https://github.com/internetarchive/iaux-item-navigator/pull/27


src/css/_BRnav.scss
```diff
@@ -97,106 +98,27 @@
   }
 }
 
-.BRReadAloudToolbar {
```
cdrini, line 100: This was duplicated from another file, so could be deleted here.

PR #1492, opened 2026-01-21
Combine book-navigator and ia-bookreader web components

cdrini:
While working on hideable chrome, I will want to later also make the book actions bar be hideable. The positioning of the toolbars is a bit difficult (especially when going fullscreen), and the two layers of web components here made this more difficult. Since we know we long-term want one main wrapping component, `ia-bookreader` with `ia-item-navigator` as its shell, combining `book-navigator` into `ia-bookreader` removes a decent amount of layers of DOM and helps us get closer to that vision!

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/src/css/icon_checkmark.js
+++ b/src/css/icon_checkmark.js
@@ -1,5 +1,9 @@
+import { html } from 'lit';
+
+export default html`
 <svg height="10" viewBox="0 0 13 10" width="13" xmlns="http://www.w3.org/2000/svg">
     <path
         d="m4.33333333 10-4.33333333-4.16666667 1.73333333-1.66666666 2.6 2.5 6.93333337-6.66666667 1.7333333 1.66666667z"
         fill="#fff" fill-rule="evenodd" />
 </svg>
+`;
--- a/src/plugins/search/search-results.js
+++ b/src/plugins/search/search-results.js
@@ -2,16 +2,16 @@
 import { unsafeHTML } from 'lit/directives/unsafe-html.js';
 import { css, html, LitElement, nothing } from 'lit';
 import '@internetarchive/ia-activity-indicator/ia-activity-indicator.js';
-import checkmarkIconSvg from '../../assets/icons/checkmark.svg?raw';
-import closeIconSvg from '@internetarchive/icon-close/close.svg?raw';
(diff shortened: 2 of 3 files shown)


src/ia-bookreader/ia-bookreader.js
```diff
@@ -1,15 +1,26 @@
...
 import { LitElement, html, css } from 'lit';
 
 import '@internetarchive/ia-item-navigator';
-import '../BookNavigator/book-navigator.js';
-// eslint-disable-next-line no-unused-vars
-import { ModalManager } from '@internetarchive/modal-manager';
 import '@internetarchive/modal-manager';
 import { SharedResizeObserver } from '@internetarchive/shared-resize-observer';
+import iaLogo from '../BookNavigator/assets/ia-logo.js';
+import SearchProvider from '../BookNavigator/search/search-provider.js';
```
jbuckner: Should we consolidate these files into the `ia-bookreader` directory?

src/ia-bookreader/ia-bookreader.js
```diff
@@ -38,63 +65,305 @@ export class IaBookReader extends LitElement {
...
+    ];
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this.unloadSharedObserver();
   }
 
-  updated() {
+  firstUpdated() {
```
jbuckner, line 123: Oh nice catch!

src/ia-bookreader/ia-bookreader.js
cdrini: Great idea! Will push that up.

src/ia-bookreader/ia-bookreader.js
```diff
@@ -38,63 +65,305 @@ export class IaBookReader extends LitElement {
     /** @type {SharedResizeObserver} */
     this.sharedObserver = undefined;
     this.loaded = false;
+    /** @type {Array<{ id: string, icon: string | import('lit').TemplateResult }>} */
     this.menuShortcuts = [];
     this.menuContents = [];
     this.openMenuName = '';
```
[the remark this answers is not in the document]
cdrini: Seems to be working, and this was, I believe, the way it was before.

PR #1501, opened 2026-01-28
fix: use referrer for toolbar return link when available

### LINKED ISSUE #1494: Add ability to return to previous BookServer Explorer App page from BookReader
Forwarded from https://github.com/ArchiveLabs/pyopds2_openlibrary/issues/26 via @MarcCoquand:

> I think if referrer is reader or openlibrary, there should be some way to return a user to a specific page when returning a book. Right now when browsing, the flow breaks because the user is sent back to the publication, rather than the book they were reading.

ParthAggarwal16:
fixes the toolbar return link to use the referrer when available instead of a static URL

This change updates the toolbar to use document.referrer when it’s available, ensuring that the return link directs the user back to the actual previous page

closes #1494


MarcCoquand:
Hi! Is this still being worked on @ParthAggarwal16 ?

MarcCoquand:
It looks good to me

PR #1512, opened 2026-02-18
"Copy link to highlight" experiment

schu96:
Experiment adding a selection menu with a "Copy link to highlight" experiment. Uses the browser-native [TextFragment](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Text_fragments)s to do the highlighting. In a future PR we will switch away from this to native 

Known issues:
* The browser-native text fragment auto-scrolls, which creates a confusing UI.
* Some quotes/selections don't work correctly in some browsers. There are some very subtle discrepancies between how the browsers handle TextFragment support.

https://github.com/user-attachments/assets/c68b5eba-0611-48e7-a287-3902959f014b

### DIFF
--- a/src/plugins/plugin.experiments.js
+++ b/src/plugins/plugin.experiments.js
@@ -51,20 +51,20 @@ export class ExperimentsPlugin extends BookReaderPlugin {
     localStorageKey: 'BrExperiments',
 
     /** The experiments that should be shown in the experiments panel */
-    enabledExperiments: ['translate', 'copyToSelectionUrl'],
+    enabledExperiments: ['translate', 'copyLinkToHighlight'],
   }
 
   /** @type {ExperimentModel[]} */
   allExperiments = [
     new class extends ExperimentModel {
-      name = 'copyToSelectionUrl';
+      name = 'copyLinkToHighlight';
       title = 'Copy to Selection URL';
       description = 'Share text selection via URL';
       learnMore = 'none';
       icon = null;
       enabled = false;
       async enable ({ manual = false }) {
-        this.br.plugins.textSelection.enableExperiment();
+        this.br.plugins.textSelection.enableSelectionMenu();
       }
       async disable() {
         sleep(0).then(() => {
--- a/src/plugins/plugin.experiments.js
+++ b/src/plugins/plugin.experiments.js
@@ -51,7 +51,7 @@ export class ExperimentsPlugin extends BookReaderPlugin {
     localStorageKey: 'BrExperiments',
 
     /** The experiments that should be shown in the experiments panel */
-    enabledExperiments: ['translate'],
+    enabledExperiments: ['translate', 'hypothesis'],
   }
 
   /** @type {ExperimentModel[]} */
--- a/src/plugins/plugin.experiments.js
+++ b/src/plugins/plugin.experiments.js
@@ -51,17 +51,17 @@ export class ExperimentsPlugin extends BookReaderPlugin {
     localStorageKey: 'BrExperiments',
 
     /** The experiments that should be shown in the experiments panel */
-    enabledExperiments: ['translate', 'highlightUrl'],
+    enabledExperiments: ['translate', 'copyToSelectionUrl'],
   }
 
   /** @type {ExperimentModel[]} */
   allExperiments = [
     new class extends ExperimentModel {
-      name = 'highlightUrl';
-      title = 'URL Link to Highlight';
-      description = 'Share highlighted content via URL';
+      name = 'copyToSelectionUrl';
+      title = 'Copy to Selection URL';
+      description = 'Share text selection via URL';
       learnMore = 'none';
-      icon = 'images/translate.svg';
+      icon = null;
       enabled = false;
       async enable ({ manual = false }) {
         this.br.plugins.textSelection.enableExperiment();
--- a/src/plugins/plugin.experiments.js
+++ b/src/plugins/plugin.experiments.js
@@ -51,11 +51,27 @@ export class ExperimentsPlugin extends BookReaderPlugin {
     localStorageKey: 'BrExperiments',
 
     /** The experiments that should be shown in the experiments panel */
-    enabledExperiments: ['translate'],
+    enabledExperiments: ['translate', 'highlightUrl'],
   }
 
   /** @type {ExperimentModel[]} */
   allExperiments = [
+    new class extends ExperimentModel {
+      name = 'highlightUrl';
+      title = 'URL Link to Highlight';
+      description = 'Share highlighted content via URL';
+      learnMore = 'none';
+      icon = 'images/translate.svg';
+      enabled = false;
+      async enable ({ manual = false }) {
+        this.br.plugins.textSelection.enableExperiment();
+      }
+      async disable() {
+        sleep(0).then(() => {
+          window.location.reload();
+        });
+      }
+    }(),
     new class extends ExperimentModel {
       name = 'translate';
       title = 'Translate Plugin';
--- a/src/plugins/plugin.experiments.js
+++ b/src/plugins/plugin.experiments.js
@@ -51,7 +51,7 @@ export class ExperimentsPlugin extends BookReaderPlugin {
     localStorageKey: 'BrExperiments',
 
     /** The experiments that should be shown in the experiments panel */
-    enabledExperiments: ['translate', 'hypothesis'],
+    enabledExperiments: ['translate'],
   }
 
   /** @type {ExperimentModel[]} */
--- a/src/plugins/plugin.experiments.js
+++ b/src/plugins/plugin.experiments.js
@@ -139,7 +139,9 @@ export class ExperimentsPlugin extends BookReaderPlugin {
     for (const experiment of this.allExperiments) {
       // TODO: imagesBaseURL should be replaced with assetRoot everywhere
       experiment.assetRoot = this.br.options.imagesBaseURL.replace(/images\/$/, '');
-      experiment.icon = experiment.buildAssetPath(experiment.icon);
+      if (experiment.icon) {
+        experiment.icon = experiment.buildAssetPath(experiment.icon);
+      }
       experiment.br = this.br;
     }
 
@@ -285,7 +287,7 @@ export class BrExperimentToggle extends LitElement {
     return html`
       <div class="experiment-card" style="margin-bottom: 10px;">
         <div style="display: flex; align-items: center; gap: 10px;">
-          <img src="${this.icon}" style="width: 20px; height: 20px;" alt="" />
+          ${this.icon ? html`<img src="${this.icon}" style="width: 20px; height: 20px;" alt="" />` : ''}
           <div style="flex-grow: 1; font-weight: bold;">${this.title}</div>
         </div>
         <p style="opacity: 0.9">
--- a/src/plugins/plugin.text_selection.js
+++ b/src/plugins/plugin.text_selection.js
@@ -63,7 +63,7 @@ export class TextSelectionPlugin extends BookReaderPlugin {
     this.textSelectionManager.init();
   }
 
-  enableExperiment() {
+  enableSelectionMenu() {
     this.textSelectionManager.selectionMenuEnabled = true;
     this.textSelectionManager.renderSelectionMenu();
   }
--- a/src/plugins/plugin.text_selection.js
+++ b/src/plugins/plugin.text_selection.js
@@ -53,6 +53,12 @@ export class TextSelectionPlugin extends BookReaderPlugin {
   init() {
     if (!this.options.enabled) return;
 
+    this.br.on('pageVisible', (_, {pageContainerEl}) => {
+      if (pageContainerEl.querySelector('.BRtextLayer')) {
+        this.br.trigger('textLayerVisible', {pageContainerEl});
+      }
+    });
+
     this.loadData();
     this.textSelectionManager.init();
   }
@@ -202,6 +208,11 @@ export class TextSelectionPlugin extends BookReaderPlugin {
       pageIndex,
       pageContainer,
     });
+
+    // Check if page is visible
+    if ($container.hasClass('BRpage-visible')) {
+      this.br.trigger('textLayerVisible', {pageContainerEl: $container[0]});
+    }
   }
 
   /**
--- a/src/plugins/plugin.text_selection.js
+++ b/src/plugins/plugin.text_selection.js
@@ -64,9 +64,8 @@ export class TextSelectionPlugin extends BookReaderPlugin {
   }
 
   enableExperiment() {
-    if (!this.textSelectionManager) return;
-    this.textSelectionManager.highlightEnabled = true;
-    this.textSelectionManager.addHighlightBar();
+    this.textSelectionManager.selectionMenuEnabled = true;
+    this.textSelectionManager.renderSelectionMenu();
   }
 
   /**
--- a/src/plugins/plugin.text_selection.js
+++ b/src/plugins/plugin.text_selection.js
@@ -188,6 +188,7 @@ export class TextSelectionPlugin extends BookReaderPlugin {
       paragEl.style.marginTop = `${newTop}px`;
       yAdded += newTop;
       textLayer.appendChild(paragEl);
+      textLayer.appendChild(document.createTextNode('\n'));
     }
     $container.append(textLayer);
     this.textSelectionManager.stopPageFlip($container);
--- a/src/plugins/plugin.text_selection.js
+++ b/src/plugins/plugin.text_selection.js
@@ -57,6 +57,12 @@ export class TextSelectionPlugin extends BookReaderPlugin {
     this.textSelectionManager.init();
   }
 
+  enableExperiment() {
+    if (!this.textSelectionManager) return;
+    this.textSelectionManager.highlightEnabled = true;
+    this.textSelectionManager.addHighlightBar();
+  }
+
   /**
    * @override
    * @param {PageContainer} pageContainer
--- a/src/plugins/url/plugin.url.js
+++ b/src/plugins/url/plugin.url.js
@@ -155,7 +155,7 @@ BookReader.prototype.urlUpdateFragment = function() {
     } else {
       const baseWithoutSlash = this.options.urlHistoryBasePath.replace(/\/+$/, '');
       const newFragmentWithSlash = newFragment === '' ? '' : `/${newFragment}`;
-      let textFragment = ""
+      let textFragment = "";
       if (this.urlPlugin.retrieveTextFragment(newQueryString)) {
         textFragment = this.urlParamsFiltersOnlySearch(this.readQueryString());
         // newQueryString = newQueryString.replace(this.urlPlugin.retrieveTextFragment(newQueryString)[0], "").replace(/(\?text=)/, "")
@@ -165,11 +165,11 @@ BookReader.prototype.urlUpdateFragment = function() {
       console.log("this is newURLPath", newUrlPath);
       try {
         window.history.replaceState({}, null, newUrlPath);
-        this.oldLocationHash = newFragment + newQueryString;
-        if (textFragment) {
-          window.location.replace('#' + textFragment);
-          this.oldLocationHash = textFragment;
-        }
+        this.oldLocationHash = newFragment + newQueryString + textFragment;
+        // if (textFragment) {
+        //   window.location.replace('#' + textFragment);
+        //   this.oldLocationHash = textFragment;
+        // }
       } catch (e) {
         // DOMException on Chrome when in sandboxed iframe
         this.options.urlMode = 'hash';
@@ -210,7 +210,6 @@ BookReader.prototype.urlParamsFiltersOnlySearch = function(url) {
 BookReader.prototype.urlReadFragment = function() {
   const { urlMode, urlHistoryBasePath } = this.options;
   if (urlMode === 'history') {
-    console.log("within plugin.url", window.location);
     return window.location.pathname.substr(urlHistoryBasePath.length);
   } else {
     return this.urlPlugin.getHash();
@@ -231,7 +230,9 @@ export class BookreaderUrlPlugin extends BookReader {
       const location = this.getLocationSearch();
       if (location.includes("text=")) {
         this.on('textLayerRendered', (_, {pageIndex, container}) => {
-          console.log("line 235 plugin.url.js window.location.replace", `#${this.oldLocationHash}`);
+          window.location.replace(`#${this.oldLocationHash}`);
+        });
+        this.on('pageVisible', (_) => {
           window.location.replace(`#${this.oldLocationHash}`);
         });
       }
--- a/src/plugins/url/plugin.url.js
+++ b/src/plugins/url/plugin.url.js
@@ -255,7 +255,12 @@ export class BookreaderUrlPlugin extends BookReader {
       if (location.includes("text=")) {
         this.on('textLayerVisible', async (_, {pageContainerEl}) => {
           const visiblePageNum = pageContainerEl.getAttribute('data-page-num');
-          await new Promise(resolve => setTimeout(resolve, 100));
+          let renderPageDelay = 100;
+          if (this.mode === 1) {
+            // maybe add some more time to have the page "settle down" from user scrolling
+            renderPageDelay = 900;
+          }
+          await new Promise(resolve => setTimeout(resolve, renderPageDelay));
           // No textFragment found or the textFragment stored doesn't match current visible page loaded
           if (!this.textFragment || this.textFragmentPage !== visiblePageNum) return;
           if (this.options.urlMode === 'history') {
--- a/src/plugins/url/plugin.url.js
+++ b/src/plugins/url/plugin.url.js
@@ -144,7 +144,8 @@ BookReader.prototype.urlUpdateFragment = function() {
   const currFragment = this.urlReadFragment();
   const currQueryString = this.getLocationSearch();
   const newQueryString = this.queryStringFromParams(params, currQueryString, this.options.urlMode);
-  if (currFragment === newFragment && currQueryString === newQueryString) {
+  const hasTextParam = currQueryString.includes("text");
+  if (currFragment === newFragment && currQueryString === newQueryString && !hasTextParam) {
     return;
   }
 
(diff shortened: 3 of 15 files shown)


src/plugins/url/plugin.url.js
```diff
@@ -154,11 +169,22 @@ BookReader.prototype.urlUpdateFragment = function() {
     } else {
       const baseWithoutSlash = this.options.urlHistoryBasePath.replace(/\/+$/, '');
       const newFragmentWithSlash = newFragment === '' ? '' : `/${newFragment}`;
-
+      const textFragment = this.urlPlugin.retrieveTextFragment(newQueryString);
       const newUrlPath = `${baseWithoutSlash}${newFragmentWithSlash}${newQueryString}`;
+      let extractedPage = newFragmentWithSlash.match(/(?<=\/)n?\d+(?=\/)/);
+      if (extractedPage) {
+        extractedPage = extractedPage[0];
+      }
```
cdrini: This will be more reliable; page can also be something like "xvi".

```suggestion
      const extractedPage = this.urlPlugin.urlStringToUrlState(newFragmentWithSlash).page;
```

src/plugins/url/plugin.url.js
```diff
@@ -180,9 +222,13 @@ BookReader.prototype.urlUpdateFragment = function() {
...
  * @return {string}
  * */
+
+// testing with this URL http://127.0.0.1:8000/BookReaderDemo/demo-internetarchive.html?ocaid=adventureofsherl0000unse&text=Well%2C I found my plans very seriously menaced.&q=breaking the law#page/18/mode/2up
 BookReader.prototype.urlParamsFiltersOnlySearch = function(url) {
   const params = new URLSearchParams(url);
-  return params.has('q') ? `?${new URLSearchParams({ q: params.get('q') })}` : '';
+  let output = '';
+  output += params.has('q') ? `?${new URLSearchParams({ q: params.get('q') })}` : '';
+  return output;
```
cdrini: I think this can be rolled back

src/plugins/url/plugin.url.js
```diff
@@ -168,8 +194,24 @@ BookReader.prototype.urlUpdateFragment = function() {
 
   if (this.options.urlMode === 'hash')  {
     const newQueryStringSearch = this.urlParamsFiltersOnlySearch(this.readQueryString());
-    window.location.replace('#' + newFragment + newQueryStringSearch);
-    this.oldLocationHash = newFragment + newQueryStringSearch;
+    let textFragment = this.urlPlugin.retrieveTextFragment(this.readQueryString());
```
cdrini, line 195: Ditto here

src/plugins/url/plugin.url.js
```diff
@@ -210,6 +256,22 @@ export class BookreaderUrlPlugin extends BookReader {
   init() {
     if (this.options.enableUrlPlugin) {
       this.urlPlugin = new UrlPlugin(this.options);
+      const location = this.getLocationSearch();
+      if (location.includes("text=")) {
+        this.on('textLayerVisible', async (_, {pageContainerEl}) => {
+          const visiblePageNum = pageContainerEl.getAttribute('data-page-num');
+          await new Promise(resolve => setTimeout(resolve, 100));
+          if (this.options.urlMode === 'history') {
+            if (this.textFragment && this.textFragmentPage == visiblePageNum) {
```
cdrini: We can pull this out from both if/else clauses and have it bail early ; eg

```js
if (!this.textFragment || this.textFragmentPage !== visiblePageNum) return;

File: src/plugins/url/plugin.url.js [hunk 5]
```diff
@@ -210,6 +256,22 @@ export class BookreaderUrlPlugin extends BookReader {
...
+              window.location.replace(`#${this.textFragment}`);
+            }
+          } else {
+            if (this.textFragment && this.textFragmentPage == visiblePageNum) {
+              window.location.replace(`#${this.oldLocationHash}`);
```
cdrini: Some comments would be nice here to note why different values are used.

src/plugins/plugin.experiments.js
```diff
@@ -51,11 +51,27 @@ export class ExperimentsPlugin extends BookReaderPlugin {
...
   }
 
   /** @type {ExperimentModel[]} */
   allExperiments = [
+    new class extends ExperimentModel {
+      name = 'highlightUrl';
+      title = 'URL Link to Highlight';
+      description = 'Share highlighted content via URL';
+      learnMore = 'none';
+      icon = 'images/translate.svg';
```
cdrini: Let's make icon optional in the rendering, and set it to null.

src/plugins/plugin.experiments.js
```diff
@@ -51,11 +51,27 @@ export class ExperimentsPlugin extends BookReaderPlugin {
...
+      enabled = false;
+      async enable ({ manual = false }) {
+        this.br.plugins.textSelection.enableExperiment();
```
cdrini: This could use a more descriptive name, since it's not clear what the experiment is; eg `enableHighlightBar`. Actually, instead of highlightBar, selectionMenu is a bit clearer; since "highlight" is often associated with annotations. Selection is the more common term for this interaction.

src/plugins/plugin.text_selection.js
```diff
@@ -53,10 +53,22 @@ export class TextSelectionPlugin extends BookReaderPlugin {
...
+        this.br.trigger('textLayerVisible', {pageContainerEl});
+      }
+    });
+
     this.loadData();
     this.textSelectionManager.init();
   }
 
+  enableExperiment() {
+    if (!this.textSelectionManager) return;
```
cdrini: This can never be null
```suggestion
```

src/plugins/plugin.text_selection.js
```diff
@@ -53,10 +53,22 @@ export class TextSelectionPlugin extends BookReaderPlugin {
...
+    this.textSelectionManager.highlightEnabled = true;
```
cdrini: Like so,
```suggestion
    this.textSelectionManager.selectionMenuEnabled = true;
```

src/plugins/plugin.text_selection.js
```diff
@@ -53,10 +53,22 @@ export class TextSelectionPlugin extends BookReaderPlugin {
...
+    this.textSelectionManager.addHighlightBar();
```
cdrini:
```suggestion
    this.textSelectionManager.renderSelectionMenu();
```

src/util/TextSelectionManager.js
```diff
@@ -280,3 +437,93 @@ export function* walkBetweenNodes(start, end) {
 
   yield* walk(start);
 }
+
+@customElement('br-highlight-bar')
+class brHighlightBar extends LitElement {
```
cdrini: Our naming convention is usually upper camel case for classes. The custom element name `br-highlight-bar` is good, but for the class we'd want something like `BRSelectionMenu`.

src/util/TextSelectionManager.js
```diff
@@ -23,6 +28,9 @@ export class TextSelectionManager {
     this.selectionElement = selectionElement;
     this.selectionObserver = new SelectionObserver(this.layer, this._onSelectionChange);
     this.options.maxProtectedWords = maxWords ? maxWords : 200;
+    this.highlightBar = new brHighlightBar(br);
+
+    this.highlightBar.className = "textFragmentBar";
```
cdrini: We often use a pattern like this to keep the naming consistent.

```suggestion
    this.highlightBar.className = "br-selection-menu__root";
```

src/util/TextSelectionManager.js
```diff
@@ -35,13 +43,35 @@ export class TextSelectionManager {
         // Set a class on the page to avoid hiding it when zooming/etc
         this.br.refs.$br.find('.BRpagecontainer--hasSelection').removeClass('BRpagecontainer--hasSelection');
         $(window.getSelection().anchorNode).closest('.BRpagecontainer').addClass('BRpagecontainer--hasSelection');
+        this.highlightBar.addHighlightButtonStyling();
```
cdrini: The naming of this method is a bit confusing ; what is the function of the styles being added? Is this what causes it to show up? To toggle? Something like `this.selectionMenu.showMenu()` or `this.selectionMenu.showIfNeeded()` or `this.selectionMenu.toggleMenu()` are clearer, and hide the implementation details of the function's internals.

src/util/TextSelectionManager.js
```diff
@@ -35,13 +43,35 @@ export class TextSelectionManager {
...
+
+      }
+
+      if (selectEvent == 'focusChanged') {
+        // hide the button as user changes their selection
+        if (this.mouseIsDown) {
+          this.highlightBar.hideHighlightStyle();
```
cdrini: Ditto with this one.

src/util/TextSelectionManager.js
```diff
@@ -35,13 +43,35 @@ export class TextSelectionManager {
...
+
+        } else {
+          if (window.getSelection().toString().length != 0) {
+            this.highlightBar.addHighlightButtonStyling();
+
+          }
+        }
```
cdrini:
```suggestion
        } else if (window.getSelection().toString()) {
          this.highlightBar.addHighlightButtonStyling();
        }
```

src/util/TextSelectionManager.js
```diff
@@ -35,13 +43,35 @@ export class TextSelectionManager {
...
       }
     }).attach();
   }
 
   // Need attach + detach methods to toggle w/ Translation plugin
   attach() {
     this.selectionObserver.attach();
+    if (!this.br.plugins.translate && this.highlightEnabled) {
+      document.body.append(this.highlightBar);
+    }
```
cdrini:
```suggestion
    this.addHighlightBar();
```

src/util/TextSelectionManager.js
```diff
@@ -67,15 +97,23 @@ export class TextSelectionManager {
     }
   }
 
+  addHighlightBar() {
+    if (document.querySelector('.textFragmentBar')) return;
+    if (this.highlightEnabled && !this.br.plugins.translate) {
```
cdrini: This will prevent the menu from rendering and appearing if the translate plugin is simply enabled ; we don't want that! We want to prevent it from appearing if a translation is active. That'll likely be better put inside the `showMenu` method.

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+ * Can import something that handles this more gracefully? see - https://web.dev/articles/text-fragments#:~:text=In%20its%20simplest%20form%2C%20the%20syntax%20of,percent%2Dencoded%20text%20I%20want%20to%20link%20to.
+ */
+/**
+ * Builds a string in the format of a TextFragment.
+ * See https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Text_fragments
+ * @param {Selection} selection - document.getSelection()
+ * @param {HTMLElement} pageLayer - anchorNode.parentElement.closest('.BRtextLayer')
+ * @returns
+ */
+export function createParam(selection, pageLayer) {
```
cdrini: This needs a clearer name; it's unclear what param it's building from the name. And the docstring needs some docs about the return type. Especially helpful to note things like: does it include `text=` ? Does it include `:~:`? Etc.

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+  const highlightedText = selection.toString().replace(/[\s]+/g, " ").trim().split(" ");
+  const anchorWord = selection.anchorNode.textContent;
+  const focusWord = selection.focusNode.textContent;
+  let textStart, textEnd; // :~:text=[prefix-,]textStart[,textEnd][,-suffix]
+  const direction = selection.direction;
+
+  // Duplicated spaces in pageLayer.textContent for some reason
+  const wholePageText = pageLayer.textContent.replaceAll("  ", " ");
```
cdrini: `wholePageText` isn't used until quite a bit lower in this method ; we can move it down to keep the related code close together.

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+ * @param {HTMLElement} pageLayer - anchorNode.parentElement.closest('.BRtextLayer')
+ * @returns
+ */
+export function createParam(selection, pageLayer) {
+  const highlightedText = selection.toString().replace(/[\s]+/g, " ").trim().split(" ");
+  const anchorWord = selection.anchorNode.textContent;
+  const focusWord = selection.focusNode.textContent;
+  let textStart, textEnd; // :~:text=[prefix-,]textStart[,textEnd][,-suffix]
+  const direction = selection.direction;
+
```
cdrini, line 272: We can avoid some of the if/else statements below for direction if we do something like:


```suggestion
const startNode = direction == 'backward' ? focusNode : anchorNode;
const endNode = direction == 'backward' ? anchorNode : focusNode;
const startWord = startNode.textContent;
const endWord = endNode.textContent;
```

That should simplify the logic of the method quite a bit.

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+    textEnd = anchorWord.replace(/[\s]+/g, "") ? anchorWord : highlightedText[highlightedText.length - 1];
+  } else {
+    textStart = anchorWord.replace(/[\s]+/g, "") ? anchorWord : highlightedText[0];
+    textEnd = focusWord.replace(/[\s]+/g, "") ? focusWord : highlightedText[highlightedText.length - 1];
+  }
+
+  const escapedStart = RegExp.escape(textStart);
+  const escapedEnd = RegExp.escape(textEnd);
+
+  // added 's' regex modifier to ensure that the matchAll actually captures instances where the escaped start / end words have whitespace
```
cdrini:
```suggestion
  // 's' regex modifier ensures the `.` also captures newline characters
```

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+  const testRegExp = new RegExp(String.raw`(?=(${escapedStart}).*?(?:(${escapedEnd})))`, "gis");
```
cdrini: This variable could use a better name; what is it a regexp for?

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+  if (direction == 'backward') {
+    textStart = focusWord.replace(/[\s]+/g, "") ? focusWord : highlightedText[0];
+    textEnd = anchorWord.replace(/[\s]+/g, "") ? anchorWord : highlightedText[highlightedText.length - 1];
+  } else {
+    textStart = anchorWord.replace(/[\s]+/g, "") ? anchorWord : highlightedText[0];
+    textEnd = focusWord.replace(/[\s]+/g, "") ? focusWord : highlightedText[highlightedText.length - 1];
+  }
+
+  const escapedStart = RegExp.escape(textStart);
+  const escapedEnd = RegExp.escape(textEnd);
```
cdrini: "escaped" is a bit of a generic term in programming, it's unclear what they have been escaped for. Is escaped to include in html? In a URL? in Regex? A clearer name is `textStartRe`. It's not uncommon to see `Re` added as a suffix like this.

```suggestion
  const textStartRe = RegExp.escape(textStart);
  const textEndRe = RegExp.escape(textEnd);
```

src/util/TextSelectionManager.js
cdrini: Since we only use this to count the number of matches, we can use a simpler regex without the lookbefore group. Although double check the tests still pass in case this was doing something special! 😁 

```suggestion
  const testRegExp = new RegExp(String.raw`(${escapedStart}).*?(${escapedEnd})`, "gis");
```

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+
+  } else {
+    preStartRange.setStart(pageLayer.firstElementChild, 0);
+    preStartRange.setEnd(selection.anchorNode, 0);
+
+    postEndRange.setStart(selection.focusNode, selection.focusNode.textContent.length);
+    postEndRange.setEnd(pageLayer.lastElementChild, pageLayer.lastElementChild.childElementCount);
+  }
+
+  const startRegex = new RegExp(String.raw`(\s+\S+){1,3}\s*?$`);
```
cdrini: We've used the words start/end in most of this function as defined by the TextFragment spec to denote the text start/end of the quote. So these regexes would be better named `prefixRe` and `suffixRe` to be consistent.

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+
+  const escapedStart = RegExp.escape(textStart);
+  const escapedEnd = RegExp.escape(textEnd);
+
+  // added 's' regex modifier to ensure that the matchAll actually captures instances where the escaped start / end words have whitespace
+  const testRegExp = new RegExp(String.raw`(?=(${escapedStart}).*?(?:(${escapedEnd})))`, "gis");
+
+  const foundMatches = wholePageText.matchAll(testRegExp).toArray();
+  if (foundMatches.length == 1) {
+    return `text=${encodeURIComponent(textStart)},${encodeURIComponent(textEnd)}`;
```
cdrini: Some comments here would be useful:

```suggestion
    // If the `start,end` quote is unambiguous and occurs only once, we can return it without a prefix/suffix
    return `text=${encodeURIComponent(textStart)},${encodeURIComponent(textEnd)}`;
```

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+  }
+  const preStartRange = document.createRange();
```
cdrini, line 291:
```suggestion
  // Otherwise we need a few words before/after the quote to disambiguate it
  const preStartRange = document.createRange();
```

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+    postEndRange.setStart(selection.focusNode, selection.focusNode.textContent.length);
+    postEndRange.setEnd(pageLayer.lastElementChild, pageLayer.lastElementChild.childElementCount);
+  }
+
+  const startRegex = new RegExp(String.raw`(\s+\S+){1,3}\s*?$`);
+  const endRegex = new RegExp(String.raw`^\S*?(\s+\S+){1,3}`);
+  // prefixes/suffixes cannot contain paragraph breaks
+  const textFragmentArr = [];
+  let [prefixes, suffixes] = ["", ""];
+  if (preStartRange.toString().length !== 0) {
```
cdrini: We can drop this outer if statement ; if it's an empty string, the code will all do the right thing.

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+
+    postEndRange.setStart(selection.focusNode, selection.focusNode.textContent.length);
+    postEndRange.setEnd(pageLayer.lastElementChild, pageLayer.lastElementChild.childElementCount);
+  }
+
+  const startRegex = new RegExp(String.raw`(\s+\S+){1,3}\s*?$`);
+  const endRegex = new RegExp(String.raw`^\S*?(\s+\S+){1,3}`);
+  // prefixes/suffixes cannot contain paragraph breaks
+  const textFragmentArr = [];
+  let [prefixes, suffixes] = ["", ""];
```
cdrini: It's confusing why these are plural; it's only a single string, so `prefix` and `suffix` should do the trick.

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+      prefixes = `${preStartRange.toString()
+        .replace(/[ ]+/g, " ")
+        .trim()
+        .replace(/^[^\n]*\n/gm, "")}-`;
+    } else {
+      prefixes = `${preStartRange.toString().match(startRegex)[0]
+        .replace(/[ ]+/g, " ")
+        .trim()
+        .replace(/^[^\n]*\n/gm, "")}-`;
+    }
```
cdrini: It would clean this up if we created a helper method called `getFirstWords(3, preStartRange.toString())`, that returns an empty string if there are no words. Then this becomes:

```suggestion
    const prefix = `${getFirstWords(3, preStartRange.toString())
        .replace(/[ ]+/g, " ")
        .trim()
        .replace(/^[^\n]*\n/gm, "")}-`;
```

(similarly for post logic)

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+        .replace(/[ ]+/g, " ")
+        .trim()
+        .replace(/\n[^\n]*$/gm, "")}`;
+    }
+  }
+
+  if (textStart === textEnd) {
+    textEnd = "";
+  }
+
```
cdrini: Why is this? This could use a comment.

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+  const constructHighlight = selection.toString().replace(/[\s]+/g, " ").split(/[ ]+/g);
+  if (direction == 'backward') {
+    constructHighlight[0] = selection.focusNode.textContent;
+    constructHighlight[constructHighlight.length - 1] = selection.anchorNode.textContent;
+  } else {
+    constructHighlight[0] = selection.anchorNode.textContent;
+    constructHighlight[constructHighlight.length - 1] = selection.focusNode.textContent;
+  }
```
cdrini: This could use a comment ; we're doing this because text fragment requires the start/end words to be "full" words, right?

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+  const fullHighlight = constructHighlight.join(" ").trim().split(" ");
+  let quote = [fullHighlight.join(" ")];
+  if (fullHighlight.length > 6) {
+    if (direction == 'backward') {
+      quote = [fullHighlight.slice(0, 3).join(" "), fullHighlight.slice(-3).join(" ")];
+    } else {
+      quote = [fullHighlight.slice(0, 3).join(" "), fullHighlight.slice(-3).join(" ")];
+    }
```
cdrini: These are identical? 😁

src/util/TextSelectionManager.js
```diff
@@ -206,6 +249,120 @@ export class TextSelectionManager {
...
+  }
+  textFragmentArr.push(...quote);
+  if (suffixes.length != 0) {
+    textFragmentArr.push(suffixes);
+  }
```
cdrini: Since suffix has to be at the end, let's put all the `textFragmentArr` logic here in one spot.

src/util/TextSelectionManager.js
```diff
@@ -280,3 +437,93 @@ export function* walkBetweenNodes(start, end) {
...
+  /** @override */
+  createRenderRoot() {
+    // Disable shadow DOM; that would require a huge rejiggering of CSS
+    return this;
+  }
+
+  /** @override */
+  connectedCallback() {
+    super.connectedCallback();
+    this.dispatchEvent(new CustomEvent('showingHighlightBar'));
```
cdrini: These show/hide events aren't being used anywhere, so let's drop them for now.

src/util/TextSelectionManager.js
```diff
@@ -280,3 +437,93 @@ export function* walkBetweenNodes(start, end) {
...
+  }
+
+  render() {
+    return html`
+      <button @click=${this.handleClick} class="textFragmentButton">
+      </button>
+    `;
+  }
+
+  handleClick(e) {
```
cdrini:
```suggestion
  handleCopyLinkToHighlight(e) {
```

src/util/TextSelectionManager.js
```diff
@@ -280,3 +437,93 @@ export function* walkBetweenNodes(start, end) {
...
+    e.preventDefault();
+    this.dispatchEvent(new CustomEvent('textFragmentButtonClicked'));
```
cdrini: This event also appears unused; let's drop.

src/util/TextSelectionManager.js
```diff
@@ -280,3 +437,93 @@ export function* walkBetweenNodes(start, end) {
...
+    this.style.top = `${hlButtonTop}px`;
+    this.style.left = `${hlButtonLeft}px`;
+    this.style.zIndex = '1';
+    this.style.position = 'absolute';
+    this.style.display = 'inline';
+  }
+
+  attachSelectionListeners = () => {
+    return;
+  }
```
cdrini: Unused

src/util/TextSelectionManager.js
```diff
@@ -280,3 +437,93 @@ export function* walkBetweenNodes(start, end) {
...
+
+  handleClick(e) {
+    e.preventDefault();
+    this.dispatchEvent(new CustomEvent('textFragmentButtonClicked'));
+    const currentUrl = window.location;
+    let currentParams = this.br.readQueryString();
+    const currentSelection = window.getSelection();
+    /** @type {HTMLElement} */
+    const textLayer = currentSelection.anchorNode.parentElement.closest('.BRtextLayer');
+    if (currentParams.includes('text')) {
```
cdrini: This is too generic; it'll also match other parameters.

cdrini (commented):
Looking great! A number of code quality fixes / naming / etc. Note I got up to `src/util/TextSelectionManager.js`, but still have the rest of the diff to get through.

src/util/TextSelectionManager.js
schu96: ~~Checked and no changes in behavior after removing the lookbefore~~

the lookbefore is needed to capture other potential overlapping texts, reverting change back

src/util/TextSelectionManager.js
```diff
@@ -269,86 +269,55 @@ export function createTextFragmentUrlParam(selection, pageLayer) {
...
-
-  } else {
-    preStartRange.setStart(pageLayer.firstElementChild, 0);
-    preStartRange.setEnd(selection.anchorNode, 0);
-
-    postEndRange.setStart(selection.focusNode, selection.focusNode.textContent.length);
-    postEndRange.setEnd(pageLayer.lastElementChild, pageLayer.lastElementChild.childElementCount);
-  }
+  postEndRange.setStart(endNode, endNode.textContent.length);
+  postEndRange.setEnd(pageLayer.lastElementChild, pageLayer.lastElementChild.childElementCount);
```
cdrini: Actually this looks off, should it be the _length_ of the last element's text content?

src/util/TextSelectionManager.js
```diff
@@ -269,86 +269,55 @@ export function createTextFragmentUrlParam(selection, pageLayer) {
...
-    postEndRange.setEnd(pageLayer.lastElementChild, pageLayer.lastElementChild.childElementCount);
-
-  } else {
-    preStartRange.setStart(pageLayer.firstElementChild, 0);
-    preStartRange.setEnd(selection.anchorNode, 0);
-
-    postEndRange.setStart(selection.focusNode, selection.focusNode.textContent.length);
-    postEndRange.setEnd(pageLayer.lastElementChild, pageLayer.lastElementChild.childElementCount);
-  }
+  postEndRange.setStart(endNode, endNode.textContent.length);
```
cdrini, line 296: Got an error on this line ; sometimes it seems like the the endNode can have no selection in it!

To repro:
http://127.0.0.1:8000/details/adventureofsherl0000unse/page/62/mode/2up

1. Highlight "It missed him" on the bottom of the right page.
2. Double/triple click on the selection until the entire line is highlighted
3. Try to copy link

Errors:

Image

cdrini (approved):
Lgtm! Updated the description with our notes/next steps. This shouldn't otherwise cause any issues with text selection, although I will do a thorough re-test of text selection before this is deployed out. Otherwise tested it and it worked -- given the known issues in the description.

PR #1519, opened 2026-03-16
Fix text selection in "Search inside" panel results

anuragc0des:
Previously, search results in the "Search inside" panel were rendered as buttons, and the global button styles applied `user-select: none`. This prevented users from selecting and copying text from search snippets.

I Updated .result-item styles in search-results.js to allow text selection in the Search inside panel.

Changes

Added user-select: text (with vendor prefixes) to .result-item to allow selecting and copying text from search results.

Preserved the existing button behavior so clicking a result still triggers the selectResult event.

Added an inline comment documenting this intentional override.


Yoglina:
Thank you, but what's the hold up? It's been 4 months already, why isn't text selection **yet** been returned to the IA book reader?

PR #1520, opened 2026-03-19
Add resolveEl() for shadow DOM compatibility

jbuckner:
## Summary
- Adds `BookReader.prototype.resolveEl()` that handles `el` as either a DOM Element or CSS selector string
- Updates `mainBRContainer` getter in ia-bookreader to use `resolveEl()`
- Updates search plugin's `renderModalMessage` to use `resolveEl()`

## Problem
`document.querySelector(br.el)` fails when:
- `br.el` is a DOM Element (not a string) — throws `SyntaxError: '[object HTMLDivElement]' is not a valid selector`
- `br.el` is a selector like `#BookReader` but the target element is inside a shadow DOM tree — returns `null`

This is needed for integrating BookReader into apps that use shadow DOM (e.g. offshoot, which wraps the details page inside `app-root`'s shadow root).

## Test plan
- [ ] Verify BookReader initializes normally with `el: '#BookReader'` (string selector, light DOM)
- [ ] Verify BookReader initializes with `el: document.getElementById('BookReader')` (DOM element)
- [ ] Verify search modal renders correctly in both modes
- [ ] Verify `mainBRContainer` returns the correct element in both modes

### DIFF
--- a/src/BookReader.js
+++ b/src/BookReader.js
@@ -789,6 +789,24 @@ BookReader.prototype.unbind = function(name, callback) {
 };
 
 /**
+ * Resolves `this.el` to a DOM Element, whether it was provided as an
+ * Element reference or a CSS selector string.
+ *
+ * This is needed because some consumers (e.g. apps using shadow DOM)
+ * must pass a DOM element directly — `document.querySelector` cannot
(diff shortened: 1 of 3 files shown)


jbuckner:

PR #1521, opened 2026-03-19
Fix hash mode to preserve URL pathname

jbuckner:
## Summary
- Replace `window.location.replace('#fragment')` with `window.location.hash = fragment`
- Applied to both `plugin.url.js` and `UrlPlugin.js`

## Problem
When `urlMode` is `'hash'`, BookReader calls `window.location.replace('#page/n5/mode/2up')`. This is a relative URL navigation that replaces the **entire URL** — not just the hash. In SPA contexts, this wipes the route path:

```
Before: /labs/details/goody
After:  /#page/n5/mode/2up     ← path gone!
```

`window.location.hash = ...` only modifies the hash fragment and always preserves the existing pathname.

## Test plan
- [ ] Verify hash mode URL updates work on `archive.org/details/*` pages
- [ ] Verify page navigation updates the hash fragment without changing the pathname
- [ ] Verify the URL is correctly restored on page reload

### DIFF
--- a/src/plugins/url/UrlPlugin.js
+++ b/src/plugins/url/UrlPlugin.js
@@ -151,7 +151,9 @@ export class UrlPlugin {
     }
 
     if (this.urlMode == 'hash') {
-      window.location.replace('#' + concatenatedPath);
+      // Use location.hash instead of location.replace('#...') to preserve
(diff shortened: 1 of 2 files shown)


jbuckner:

PR #1538, opened 2026-04-20
Switch "Copy Link to Highlight" to use custom highlight rendering + beginnings of annotations support

schu96:
Pivoting from using native browser Text Fragments 

Highlights in the bookreader can now be stored in browser localStorage, with plans for annotations coming soon

### DIFF
--- a/src/plugins/plugin.experiments.js
+++ b/src/plugins/plugin.experiments.js
@@ -60,11 +60,25 @@ export class ExperimentsPlugin extends BookReaderPlugin {
       name = 'copyLinkToHighlight';
       title = 'Copy to Selection URL';
       description = 'Share text selection via URL';
-      learnMore = 'none';
       icon = null;
       enabled = false;
       async enable ({ manual = false }) {
         this.br.plugins.textSelection.enableSelectionMenu();
+      }
+      async disable() {
+        sleep(0).then(() => {
+          window.location.reload();
+        });
+      }
+    }(),
+    new class extends ExperimentModel {
+      name = 'annotateHighlight';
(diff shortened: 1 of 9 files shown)


src/plugins/url/UrlPlugin.js
```diff
@@ -208,4 +208,31 @@ export class UrlPlugin {
...
+  retrieveHighlightContext(urlString, type) {
+    const regexString = new RegExp(String.raw`(?<=[&?]?${type}=)[^&]*`, "gis");
+    return urlString.match(regexString);
+  }
+
+  /**
+   * @param {string} urlString
+   * @returns {object}
+   */
+  convertParamToHighlight(urlString) {
```
cdrini: This would be better as a single url parameter, similar to our TextFragment from before. Eg `?text=5:prefix,quote,suffix` . Note the tweak that includes the page index at the front to help with disambiguation.

src/plugins/plugin.experiments.js
```diff
@@ -51,7 +51,7 @@ export class ExperimentsPlugin extends BookReaderPlugin {
     localStorageKey: 'BrExperiments',
 
     /** The experiments that should be shown in the experiments panel */
-    enabledExperiments: ['translate', 'copyLinkToHighlight'],
+    enabledExperiments: ['translate', 'copyLinkToHighlight', 'annotateHighlight'],
```
cdrini, line 54:
```suggestion
    enabledExperiments: ['translate', 'copyLinkToHighlight'],
```

cdrini:
Closing as the changes here are now in #1548 and #1554

PR #1544, opened 2026-05-14
Switch page display to be 0-indexed

pezvi:
Per bk request, because it was causing confusion when the number in the UI was different from the n-number in the URL (eg `/page/n0`).

### DIFF
--- a/src/BookReader/Navbar/Navbar.js
+++ b/src/BookReader/Navbar/Navbar.js
@@ -445,7 +445,7 @@ export class Navbar {
 
 /**
  * Renders the html for the page string
- * @param {number} index 0-indexed leaf position (0..numLeafs-1)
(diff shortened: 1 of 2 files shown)


src/BookReader/Navbar/Navbar.js
```diff
@@ -446,19 +445,19 @@ export class Navbar {
 
 /**
  * Renders the html for the page string
- * @param {number} index
- * @param {number} numLeafs
+ * @param {number} index 0-indexed leaf position (0..numLeafs-1)
```
cdrini: We try to use "PageIndex" as the unambiguous concept name in the repo.

```suggestion
 * @param {PageIndex} index 0-indexed page index (0..numLeafs-1)
```

cdrini (approved):
Lgtm, thank you! Tested on netlify and is working as expected.

PR #1549, opened 2026-06-03
Do not fetch/render text layer for unviewable pages

cdrini:


cdrini:
Auto-merging, small one line fix. Confirmed text layer works on normal books.

PR #1553, opened 2026-06-11
Add support for quote start/end to Copy Link to Highlight

cdrini:
Add support for including only the quote start and end in the url when using Copy Link to Highlight, instead of having very long URLs.

Eg https://deploy-preview-1553--ia-bookreader.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala&text=221:one%20of%20those.-,Socr.%20Neither%2C%20my,speak%20to%20you#page/220/mode/2up

### DIFF
--- a/src/util/TextSelectionManager.js
+++ b/src/util/TextSelectionManager.js
@@ -790,80 +790,62 @@ function replaceWhitespace(string) {
 }
 
 /**
- * Finds ordered, non-overlapping quote matches within a range and returns a Range for each.
- * @param {String[]} quotes - Ordered quote parts to find
+ * Finds all regex matches within a range and returns their ranges.
+ * @param {RegExp} regex
  * @param {Range} range - the range to search in
  * @param {Node[]} textNodes - visible text nodes within the range
- * @returns {Range[] | undefined} Ranges for each quote part, or undefined if no full ordered match is found
- *
- * Adapted from
- * https://github.com/GoogleChromeLabs/text-fragments-polyfill/blob/abc6ed408b3f20e91d9cbda9977748459f5e3877/src/text-fragment-utils.js#L765
+ * @param {{ normalize?: function(string): string }} [options]
+ * @returns {Range[] | undefined} Range for each match, or undefined if no matches are found
  */
-export function findRangesForQuotes(quotes, range, textNodes) {
-  if (!quotes.length) return undefined;
+export function findRangeForRegExp(regex, range, textNodes, { normalize = (s) => s } = {}) {
+  if (!textNodes.length) return undefined;
 
   const startOffset = textNodes[0] === range.startContainer ?
     range.startOffset :
     0;
-  const normalizedWholePageString = replaceWhitespace(range.toString());
-  const normalizedQuotes = quotes.map(replaceWhitespace);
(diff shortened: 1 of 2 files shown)


src/util/TextSelectionManager.js
```diff
@@ -788,53 +790,62 @@ function replaceWhitespace(string) {
...
- * @param {String} quote - The text to find
+ * Finds all regex matches within a range and returns their ranges.
+ * @param {RegExp} regex
  * @param {Range} range - the range to search in
  * @param {Node[]} textNodes - visible text nodes within the range
- * @returns {Range | undefined} Range that encompasses the quote, or undefined if no match is found
- *
- * Lightly adapted from
- * https://github.com/GoogleChromeLabs/text-fragments-polyfill/blob/abc6ed408b3f20e91d9cbda9977748459f5e3877/src/text-fragment-utils.js#L765
+ * @param {{ normalize?: function(string): string }} [options]
```
schu96, line 797: I'm not a fan of this, it feels like it just adds complexity for relatively low payoff

src/util/TextSelectionManager.js
```diff
@@ -788,53 +790,62 @@ function replaceWhitespace(string) {
...
 
 /**
  * Uses the index that matches the quote string and normalizes the string contents to find the correct node
  * @param {Number} index
  * @param {Node[]} nodes
  * @param {boolean} isEnd
+ * @param {{ normalize?: function(string): string }} [options]
  */
-export function getBoundaryPointAtIndex(index, nodes, isEnd) {
+export function getBoundaryPointAtIndex(index, nodes, isEnd, { normalize = (s) => s } = {}) {
```
schu96, line 848: Similar feedback here, I can't really anticipate a case where the normalization method for the text would be different than replacing the whitespace

schu96 (approved):
Tested the branch, the truncated highlight selection functionality is working as expected and much cleaner compared to what it was before! 

Aside from my personal preference of excluding the `normalize` param in some of the functions this looks good to me. 

I did notice that the copy link to highlight target quotes doesn't let a user add the highlight on their end. The newly created `mark` HTML elements seem to be the cause of the issue, would probably be better to address this after the annotations feature is reviewed and merged in

src/util/TextSelectionManager.js
cdrini: Fair! I've been having some issues with the API design of this field, because of the non-standard URL encoding, and have been debating potentially making it normalize to not include any of the delimiting characters, eg ":" or "," or "-," or ",-" . That would allow the field to be url encoded all together potentially? So this change was made as a stepping stone to that potential future.

It was also made to try to communicate that it's really important that the same normalization is used throughout the entire pipeline, otherwise strings might never match! By making it configurable, the "parent" function that calls both these other functions is in control of that normalization, and making sure the same normalization is used. They can't fall out of sync.

cdrini:
> I did notice that the copy link to highlight target quotes doesn't let a user add the highlight on their end. The newly created `mark` HTML elements seem to be the cause of the issue, would probably be better to address this after the annotations feature is reviewed and merged in

I'm not sure I follow; can you expand on what's not working?

PR #1554, opened 2026-06-12
Migrate annotation changes

schu96:
Migrated annotation code to work with newly improved main branch

### DIFF
--- a/src/plugins/plugin.annotations.js
+++ b/src/plugins/plugin.annotations.js
@@ -11,9 +11,6 @@ export class BRAnnotationModal extends LitElement {
   /** @type {import('../BookReader.js').default} */
   br;
 
-  @property({type: Boolean})
-  allowAnnotationEditing = false;
-
   HIGHLIGHT_YELLOW = "#ffff00";
   HIGHLIGHT_PINK = "#ffc0cb";
   HIGHLIGHT_ORANGE = "#ffa500";
@@ -26,6 +23,10 @@ export class BRAnnotationModal extends LitElement {
   currentAnnotationNodes;
 
   storageService = new AnnotationStorageService({storageMethod: window.localStorage});
+
+  @property({type: Boolean})
+  showColorOptions = false;
+
 
   /**
    *
@@ -48,38 +49,85 @@ export class BRAnnotationModal extends LitElement {
     this.setAttribute('aria-label', 'Annotation actions');
   }
 
-  showExistingAnnotation() {
+  showTextEditArea() {
     return html`
-      <div class="br-annotate-menu__comment" aria-hidden="true">
-        <span class="br-annotate-menu__displayAnnotation">${this.getAnnotationText()}</span>
-      </div>
-      <button @click=${this.handleEditAnnotation}>Edit annotation</button>
-      <button @click=${this.hide}>Cancel</button>
+    <div class="br-annotate-menu__body">
+      <textarea 
+        class="br-annotate-menu__textArea" 
+        id="annotateTextArea" 
+        placeholder="Add note..."
+        >${this.getAnnotationText()}</textarea>
+    </div>
+    <div class="br-annotate-menu__footer">
+        ${this.renderColorOptions()}
+        <div class="br-annotate-menu__editOptions">
+      <button 
+        @click=${this.handleDeleteHighlight}
+        class="br-annotate-menu__option">Delete
+      </button>
+      <button
+      @click=${this.handleSaveAnnotation}
+      class="br-annotate-menu__option save"
+      >Save</button>
+    </div>
     `;
   }
 
-  showTextEditArea() {
+  renderColorOptions() {
+    const colors = {
+      '#00ff00': 'green',
+      '#ffc0cb': 'pink',
+      '#ffff00': 'yellow',
+      '#ffa500': 'orange',
+    };
+    const color = this.getHighlightColor();
     return html`
-      <textarea class="br-annotate-menu__textarea" id="annotateTextArea" @click=${this.handleInputClick}>${this.getAnnotationText()}</textarea>
-      <button @click=${this.handleSaveAnnotation}>Save annotation</button>
-      <button @click=${this.hide}>Cancel</button>
+    <div class="br-annotate-menu__colorOptions">
+      <button
+        @click=${this.handleColorChange}
+        class="br-annotate-menu__color ${colors[color]}"
+        value=${color}
+      >
+      </button>
+      ${this.showColorOptions ? this.renderColorDropdown() : ""}
+      <button class="br-annotate-menu__carot"
+      @click=${this.handleShowColor}>${this.showColorOptions ? '<' : '>'}</button>
+    </div>
+    `;
+  }
+
+  handleShowColor() {
+    this.showColorOptions = !this.showColorOptions;
+  }
+
+  renderColorDropdown() {
+    const colors = {
+      'green': this.HIGHLIGHT_GREEN,
+      'pink' : this.HIGHLIGHT_PINK,
+      'orange' : this.HIGHLIGHT_ORANGE,
+      'yellow': this.HIGHLIGHT_YELLOW,
+    };
+    const currentColor = this.getHighlightColor();
+    const allColorOptions = Array.from(Object.keys(colors)).map((color) => {
+      if (colors[color] === currentColor) return;
+      return html`
+      <button
+        @click=${this.handleColorChange}
+        class="br-annotate-menu__color ${color}"
+        value=${colors[color]}>
+        </button>`;
+    });
+    return html`
+      <div class="br-annotate-menu__pipe">|</div>
+      <div class="br-annotate-menu__colorDropdown">
+        ${allColorOptions}
+      </div>
     `;
   }
 
   render() {
     return html`
-      <input type="color"
-        id="annotateInputColor"
-        value=${this.getHighlightColor()}
-        class="br-annotate-menu__color"
-        @change=${this.handleColorChange}
-      />
-      <button @click=${this.handleDeleteHighlight}
-        class="br-annotate-menu__option"
-        <ia-icon-share class="br-annotate-menu__icon" aria-hidden="true"></ia-icon-share>
-        <span class="br-annotate-menu__label">Delete Highlight and Annotation</span>
-      </button>
-      ${this.allowAnnotationEditing ? this.showTextEditArea() : this.showExistingAnnotation()}
+      ${this.showTextEditArea()}
     `;
   }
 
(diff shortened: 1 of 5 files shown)


src/util/TextSelectionManager.js
```diff
@@ -723,22 +748,239 @@ class BRSelectMenu extends LitElement {
...
-  clearNodesForRemoval = () => {
-    this.nodesForRemoval = null;
+  clearActiveHighlightNodes = () => {
+    this.activeHighlightNodes = null;
+    this.requestUpdate();
+  }
+}
+
+@customElement('br-annotation-menu')
+class BRAnnotationMenu extends LitElement {
```
cdrini: This more of a modal then a menu ; so let's call call it BRAnnotationModal or *Popup or *Form or something.

cdrini: Also let's create a new file, plugin.annotations.js and move this there and then import it into this file. That'll let us start structuring these in the way we'll eventually want them to be structured, can keep this file a bit more focussed/targetted.

src/util/TextSelectionManager.js
```diff
@@ -723,22 +748,239 @@ class BRSelectMenu extends LitElement {
...
+  allowAnnotationEditing = false;
+
+  @property({type: String})
+  HIGHLIGHT_YELLOW = "#ffff00";
+  @property({type: String})
+  HIGHLIGHT_PINK = "#ffc0cb";
+  @property({type: String})
+  HIGHLIGHT_ORANGE = "#ffa500";
+  @property({type: String})
+  HIGHLIGHT_GREEN = "#00ff00"
```
cdrini: These are constant values, so don't need to be `@property`. `@property` is what makes a variable reactive in lit. If these are constants they won't much benefit from that!

src/util/TextSelectionManager.js
```diff
@@ -723,22 +748,239 @@ class BRSelectMenu extends LitElement {
...
+
+  @property({type: String})
+  lastHighlightColorUsed = this.HIGHLIGHT_YELLOW;
+
+  currentAnnotationNodes;
```
cdrini: jsdoc/types

src/util/TextSelectionManager.js
```diff
@@ -723,22 +748,239 @@ class BRSelectMenu extends LitElement {
...
+
+  /** @override */
+  createRenderRoot() {
+    return this;
+  }
+
+  /** @override */
+  connectedCallback() {
+    super.connectedCallback();
+    this.setAttribute('role', 'menu');
```
cdrini: I don't think this would be considered an aria menu; see https://www.w3.org/WAI/ARIA/apg/patterns/menubar/ . Alert/dialog might be most appropriate? https://www.w3.org/WAI/ARIA/apg/patterns/

src/util/TextSelectionManager.js
```diff
@@ -723,22 +748,239 @@ class BRSelectMenu extends LitElement {
...
+  handleDeleteHighlight() {
+    if (this.currentAnnotationNodes) {
+      const currentUUID = retrieveUUID(this.currentAnnotationNodes[0]);
+      const storage = loadHighlightsFromLocalStorage();
+      for (const idx in storage) {
+        if (storage[idx].uuid === currentUUID) {
+          storage.splice(idx, 1);
+          saveToLocalStorage(storage);
+          for (const ele of this.currentAnnotationNodes) {
+            const tempText = ele.textContent;
```
cdrini: This is a lot of logic in this web component that requires internal knowledge of how annotations are saved. It would be better if it didn't have that internal knowledge, and all that logic was centralized in one place. Let's create a new class (in plugin.annotations.js) called `AnnotationStorageService`. That should have methods for easy edit/save/delete. This will also set us up for when we switch all these methods from local storage to instead using API calls.

This component can then just take in the storage service as another parameter.

src/util/TextSelectionManager.js
```diff
@@ -723,22 +748,239 @@ class BRSelectMenu extends LitElement {
...
+@customElement('br-annotation-menu')
+class BRAnnotationMenu extends LitElement {
+  /** @type {import('../BookReader.js').default} */
+  br;
+
+  @property({type: String})
+  icon = '';
+
+  @property({type: String})
+  label = '';
```
cdrini: Are these used?

src/util/TextSelectionManager.js
```diff
@@ -723,22 +748,239 @@ class BRSelectMenu extends LitElement {
...
+              parent.textContent = tempText;
+            }
+          }
+          this.hide();
+        }
+      }
+    }
+  }
+
+  show(nodes) {
```
cdrini: jsdoc/types

src/util/TextSelectionManager.js
```diff
@@ -723,22 +748,239 @@ class BRSelectMenu extends LitElement {
...
+    if (this.br.plugins.translate?.userToggleTranslate) return;
```
cdrini: Is this relevant here? This can only be displayed via the select menu, which should already have that check embedded in it.

cdrini (commented):
Got through the first half ; will finish the rest later today!

src/plugins/plugin.annotations.js
```diff
@@ -0,0 +1,371 @@
...
+import { live } from "lit/directives/live.js";
+import { BookReaderTextFragment } from "../util/TextSelectionManager.js";
+import '@internetarchive/icon-share';
+// eslint-disable-next-line no-unused-vars
+import { OlPopover } from "../util/OlPopover.js";
+
+const BR_HIGHLIGHTS_LOCAL_STORAGE_KEY = "BRhighlightStorage";
+
+@customElement('br-annotation-modal')
+
```
cdrini, line 12:
```suggestion
```

src/plugins/plugin.annotations.js
```diff
@@ -0,0 +1,371 @@
...
+      <button
+      @click=${this.handleSaveAnnotation}
+      class="br-annotate-menu__option save"
+      >Save</button>
+    </div>
+    `;
+  }
+
+  renderColorOptions() {
+    const colors = {
```
cdrini, line 99: This appears to be duplicating the static constants on this class also below where have another similar object. These might be able to be DRY'd up somehow so that we're not duplicated the color names / hex values.

src/plugins/plugin.annotations.js
```diff
@@ -0,0 +1,371 @@
...
+  }
+
+  render() {
+    return html`
+      <ol-popover
+        aria-label="Annotation actions"
+        placement="bottom-start"
+        .anchor=${this.anchorEl}
+        .open=${live(this.open)}
+        ._position=${this.positionObj}
```
cdrini, line 157: This seems like something that shouldn't be private ; we can like make _position just position.

src/plugins/plugin.annotations.js
```diff
@@ -0,0 +1,371 @@
...
+      this.storageService.edit(currentUUID, 'highlightColor', e.target.value);
+    }
+    this.handleShowColor();
+  }
+
+  handleDeleteHighlight() {
+    if (this.currentAnnotationNodes) {
+      const currentUUID = retrieveUUID(this.currentAnnotationNodes[0]);
+      this.storageService.delete(currentUUID);
+      for (const ele of this.currentAnnotationNodes) {
```
cdrini, line 180: I wonder if these delete logic might be DRY'd? I think we have another spot where we can delete from that is outside the Annotation modal class. 

One potential way to DRY this is to have storageService emit an event for when something is deleted/edited, and then have the annotation plugin class be the one responsible for making sure changes to the data are reflected in the UI.

src/plugins/plugin.annotations.js
```diff
@@ -0,0 +1,371 @@
...
+    this.currentAnnotationNodes = null;
+    this.showColorOptions = false;
+    this.display = 'none';
+    this.open = false;
+    this.requestUpdate();
+    return;
+  }
+
+  updateTextArea(text) {
+    const inputEle = document.querySelector("#annotateTextArea");
```
cdrini, line 241:
```suggestion
    const inputEle = this.querySelector("#annotateTextArea");
```

src/plugins/plugin.annotations.js
```diff
@@ -0,0 +1,371 @@
...
+import { BookReaderTextFragment } from "../util/TextSelectionManager.js";
+import '@internetarchive/icon-share';
+// eslint-disable-next-line no-unused-vars
+import { OlPopover } from "../util/OlPopover.js";
+
+const BR_HIGHLIGHTS_LOCAL_STORAGE_KEY = "BRhighlightStorage";
+
+@customElement('br-annotation-modal')
+
+export class BRAnnotationModal extends LitElement {
```
cdrini, line 13: AnnotationModal
takes in:
    - AnnotationObject
    - StorageService (.get, .delete, .edit)
    - position

AnnotationPlugin
     - own storage service (creates/destroys it)
     - when BR renders a page, render associated annotations
     - it knows where the highlight is rendered, so it has all the info needed to compute the position

src/util/OlPopover.js
```diff
@@ -0,0 +1,865 @@
+import { LitElement, html, css, nothing } from 'lit';
```
cdrini, line 1: Add a comment here that this is from the open library repo so we can at some point potentially try to update it.

src/util/OlPopover.js
```diff
@@ -0,0 +1,865 @@
...
+     * Compute the final position of the popover panel, flipping and shifting
+     * as needed to keep it within the viewport.
+     */
+    _computePosition(panelW, panelH) {
+      const trigger = this._anchorEl;
+      if (!trigger) return;
+
+      const anchor = trigger.getBoundingClientRect();
+      if (anchor.x == 0 && anchor.y == 0 && anchor.width == 0) return;
+      const pageContainerBoundary = trigger.closest(".BRpagecontainer")?.getBoundingClientRect();
```
cdrini, line 510: It would be good for this class to be br-agnostic ; is there a way we can pass in the position info it's trying to compute.

src/util/scroll-lock.js
```diff
@@ -0,0 +1,57 @@
+/**
```
cdrini, line 1: Add a link to the OL file so we can potentially in theory at some point sync/updated.

src/util/TextSelectionManager.js
```diff
@@ -680,21 +719,33 @@ class BRSelectMenu extends LitElement {
...
-    // TODO
+  /**
+   * @param {MouseEvent} e
+   */
+  handleAddAnnotation (e) {
+    const anchorEl = /** @type {HTMLElement} */ (e.currentTarget);
+    if (!this.activeHighlightNodes) { // highlight selection if not already done
+      this.handleHighlightSave();
+    }
+    this.br.plugins.textSelection?.textSelectionManager.showAnnotationModal(this.activeHighlightNodes, anchorEl);
```
cdrini, line 740: I think we need to create the plugin annotations class matching the other plugins, so that we can attach the annotation-specific methods/data/logic to that alone.

That also keeps/fulfills the promise of BR plugins being optional. Letting people load the plugin files only if they support/want to support the associated functionality.

```suggestion
    this.br.plugins.annotations.showAnnotationModal(this.activeHighlightNodes, anchorEl);
```

src/util/TextSelectionManager.js
```diff
@@ -230,6 +246,27 @@ export class TextSelectionManager {
     this.selectMenu.hide();
   }
 
+  /**
+   * @param {HTMLElement[]} nodes
+   * @param {HTMLElement} [anchorEl] Element to anchor the annotation popover
+   *  to, e.g. the br-menu-option button that was clicked
+   */
+  showAnnotationModal(nodes, anchorEl) {
```
cdrini, line 254: These sorts of methods would make sense to move to a dedicated AnnotationPlugin

src/util/TextSelectionManager.js
```diff
@@ -1192,6 +1278,24 @@ function retrieveUUID(ele) {
   return null;
 }
 
+/**
+   * Saves all the highlights in an array to localStorage
+   * @param {BookReaderSavedHighlight[]} highlights
+   */
+function saveToLocalStorage(highlights) {
```
cdrini, line 1285: Should these still exist? Or should these now be done by the StorageService?

I think we want to treat highlights and annotations as the same thing. They're both "Annotations" ; one just has a comment attached to it, and one doesn't.

src/util/TextSelectionManager.js
```diff
@@ -1081,14 +1145,36 @@ export function renderHighlight(textLayer, textFragment, cssClassName = null) {
   // Don't mark if already marked
   if (startTextNode.parentElement.classList.contains("BRhighlight") ||
       endTextNode.parentElement.classList.contains("BRhighlight")) {
-    return;
+    const existingMarks = [];
+    const walker = document.createTreeWalker(
```
cdrini, line 1149: Is there anyway we can re-use some of the helper methods we have? Not sure.

cdrini:
UI feedback:

1. CSS tweaks for modal
2. chrome mobile issues

PR #1557, opened 2026-06-15
Fix sentry error from calling non-existent method

cdrini:
Restore method removed in #1492 that was apparently being used!

Fixes https://sentry.archive.org/organizations/ia-ux/issues/624258/?environment=production&project=6&query=is%3Aunresolved&referrer=issue-stream&sort=freq

image


cdrini:
Auto-merging; small error fix.

PR #1561, opened 2026-06-19
Update dependencies

cdrini:


cdrini:
Auto-merging; simple dependency updates. e2e passes and using the site generally works

PR #1562, opened 2026-06-26
Make Copy Link to Highlight links more reslient to url encoding

cdrini:
Sharing a "Copy link to highlight" link on eg facebook messenger resulted in the parameters being re-encoded:

Input link: `https://archive.org/details/macklife0000mack/page/246/mode/2up?text=246:to%20some%20extent%20%E2%80%98silly%E2%80%99.-,But%20I%20can,We%E2%80%99re%20all%20%E2%80%98them%E2%80%99.,-Lee%20Mack`

Link in facebooks messenger: `https://archive.org/details/macklife0000mack/page/246/mode/2up?text=246%3Ato%20some%20extent%20%E2%80%98silly%E2%80%99.-%2CBut%20I%20can%2CWe%E2%80%99re%20all%20%E2%80%98them%E2%80%99.%2C-Lee%20Mack`

Note the parameter becomes fully url encoded, with eg `:` being changed to `%3A` , and also `,` being replaced.

This broke the rather brittle custom url encoding the text parameter was using. Switched to a technique where the delimiter characters are removed from the source text, so the entire string can safely be encoded as a whole!

Test here: https://deploy-preview-1562--ia-bookreader.netlify.app/details/theworksofplato01platiala

### DIFF
--- a/src/BookReader.js
+++ b/src/BookReader.js
@@ -41,7 +41,6 @@ import { ModeThumb } from './BookReader/ModeThumb.js';
 import { ImageCache } from './BookReader/ImageCache.js';
 import { PageContainer } from './BookReader/PageContainer.js';
 import { NAMED_REDUCE_SETS } from './BookReader/ReduceSet.js';
-import {BookReaderTextFragment} from './util/TextSelectionManager.js';
 
 /**
  * BookReader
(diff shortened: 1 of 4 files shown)


schu96 (changes requested):
@cdrini I just tested using this link and it doesn't seem to be working from just a direct copy link to highlight paste into the web console. Might need to do some more testing on what is causing it to not match

https://deploy-preview-1562--ia-bookreader.netlify.app/details/theworksofplato01platiala/page/70/mode/2up?text=70:Yes%20%22%20he-,replied.%2043.%20%22And,each%20other%20and,-be%20subject%20to

Image

schu96:
Confirmed that the paragraph does highlight on the current master branch
image

schu96 (approved):
Looking good for the original reported issue

Image

FYI, I was able to find an edge case where the highlighted text is not unique but the prefix/suffix should be able to help make it less ambiguous

http://127.0.0.1:8000/BookReaderDemo/demo-internetarchive.html?ocaid=theworksofplato01platiala&text=74:we%20know%20what%20it%20is%20itself%3F%22-,%22%20Certainly%20%22%20he%20replied.#page/74/mode/2up

Image

cdrini:
Great catch! There was a bug when a delimiter character appeared as an entire separate word.

PR #1564, opened 2026-06-29
Change context menu to horizontal layout

schu96:
Added some CSS styling changes to change the alignment from vertical to horizontal layout for highlight menu. Includes a new `...` menu option that changes the options for potential features


cdrini (approved):
Lgtm! One question on whether we want to show the text by default on desktop, but I think it's good as-is.

PR #1568, opened 2026-07-03
Group renovate PRs by dependency type to reduce number of PRs

cdrini:
These PRs are starting to cause some strain on our netlify usage, and aren't useful in their current granularity.


cdrini:
Auto-merging ; small dependency cleanup/organization

PR #1572, opened 2026-07-22
Convert iframe plugin to class-based registration pattern

cdrini:


cdrini:
Auto-merging, small refactor.

PR #1573, opened 2026-07-23
WEBDEV-8789: Use caret ranges for pinned @internetarchive/* deps

jbuckner:
Switch every `@internetarchive/*` dependency to a caret range (they were exact-pinned as of "Update IA deps"). `bergamot-translator` already used a caret.

**Why:** exact pins force downstream consumers (offshoot, etc.) that share these packages to add `overrides` to dedupe, and can produce duplicate custom-element registrations when versions don't line up (see the recent `ia-activity-indicator` dupe fix). Carets let npm hoist a single compatible version.

Versions and the `update:ia-deps` script are unchanged; resolved versions in the lockfile are identical (only the range strings move).

cc @cdrini — this softens the exact-pin convention you set up in "Update IA deps"; flagging since it's a deliberate reversal of that for the consumer-dedup benefit.

[WEBDEV-8789](https://webarchive.jira.com/browse/WEBDEV-8789)

[WEBDEV-8789]: https://webarchive.jira.com/browse/WEBDEV-8789?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cdrini (approved):
Lgtm! Pinning the version numbers is a security precaution in light of the various npm compromises. By pinning the version, if a package gets compromised, we won't automatically be updated to that latest (compromised) version with `npm i`.

As we discussed, to make it easier to maintain/consistent with other projects, it make sense for the IA ones not to have that requirement. This will be a bit tricky to maintain in BR, since we have save-exact in our .npmrc here, which I think makes sense for third-party deps, but this might be ok for now.

I also updated the GHA workflow to avoid using `npm install` and use `npm ci` instead, which is more secure since it only reads from `package-lock.json` . This does come at a performance cost, since `npm ci` deletes the `node_modules` directory entirely, which seems to add about 10s to each post-install job (even with global node modules cache), but that's manageable. Might be worth switching to pnpm in the future!

PR #1575, opened 2026-07-24
Fix scrollLeft resetting while scrolling in 1up on Chrome/Firefox

cdrini:
Calling getBoundingClientRect() during text layer rendering was forcing a synchronous reflow of the whole document -- for some reason resetting the scroll. So now we create a temporary iframe where we call getBoundingClientRect, which does not!

QA: https://deploy-preview-1575--ia-bookreader.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala

To reproduce:

1. Open a book in mobile
2. Click on "fullscreen"
3. Zoom in and scroll left some amount
4. Scroll down past a few pages

Expected behaviour: Your scroll left position should stay the same  
Actual behaviour: It resets to the edge after a page or so

https://github.com/user-attachments/assets/28eb5889-4635-4e6a-8585-8b31e5e44297

### DIFF
--- a/src/plugins/plugin.text_selection.js
+++ b/src/plugins/plugin.text_selection.js
@@ -115,7 +115,7 @@ export class TextSelectionPlugin extends BookReaderPlugin {
   _configurePageContainer(pageContainer) {
     // Disable if thumb mode; it's too janky
     // .page can be null for "pre-cover" region
-    if (this.br.mode !== this.br.constModeThumb && pageContainer.page?.isViewable) {
+    if (this.options.enabled && this.br.mode !== this.br.constModeThumb && pageContainer.page?.isViewable) {
       this.createTextLayer(pageContainer);
     }
     return pageContainer;
--- a/src/plugins/plugin.text_selection.js
+++ b/src/plugins/plugin.text_selection.js
@@ -6,6 +6,7 @@ import { Cache } from '../util/cache.js';
 import { toISO6391 } from './tts/utils.js';
 import { BookReaderTextFragment, renderHighlight, TextSelectionManager } from '../util/TextSelectionManager.js';
 import { genMap, lookAroundWindow, zip } from '../util/generators.js';
+import textSelectionCss from '../css/_TextSelection.scss';
 /** @typedef {import('../util/strings.js').StringWithVars} StringWithVars */
 /** @typedef {import('../BookReader/PageContainer.js').PageContainer} PageContainer */
 
@@ -40,6 +41,13 @@ export class TextSelectionPlugin extends BookReaderPlugin {
   _jumpedToHighlight = false;
 
   /**
+   * Isolated document/layout used to performantly measure OCR text-layer
+   * elements.
+   * @type {Document}
+   */
+  _measurementDocument;
+
+  /**
    * @param {import('../BookReader.js').default} br
    */
   constructor(br) {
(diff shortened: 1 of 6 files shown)


src/plugins/plugin.text_selection.js
```diff
@@ -56,6 +64,19 @@ export class TextSelectionPlugin extends BookReaderPlugin {
   init() {
     if (!this.options.enabled) return;
```
nsharma123, line 65: `_measurementDocument` is only ever set when this early return is *not* taken, i.e. only when `options.enabled` is true. But `_configurePageContainer` → `createTextLayer` → `renderParagraph` → `determineRealRects(..., this._measurementDocument)` all run unconditionally from `BookReader.prototype._createPageContainer` (it iterates `this.plugins` and calls `_configurePageContainer` on each, with no `options.enabled` check).

So setting `plugins: { textSelection: { enabled: false } }` now throws instead of silently no-op'ing like it did pre-PR (previously `determineRealRects` used the global `document`, which always exists). Confirmed locally:
```js
const br = new BookReader({ plugins: { textSelection: { enabled: false } }, data: [...] });
br.init();
br._createPageContainer(0, {}); // TypeError: Cannot read properties of undefined (reading 'body')
```
Worth guarding the `determineRealRects` call sites (or falling back to `document` when `_measurementDocument` is unset) so disabling the plugin doesn't break page rendering.

src/plugins/plugin.text_selection.js
```diff
@@ -56,6 +64,19 @@ export class TextSelectionPlugin extends BookReaderPlugin {
...
 
+    // Setup measurement iframe for OCR
+    const measurementIframe = document.createElement('iframe');
+    measurementIframe.setAttribute('aria-hidden', 'true');
+    measurementIframe.tabIndex = -1;
+    measurementIframe.style.cssText = 'position:fixed; top:-99999px; left:-99999px; width:2000px; height:4000px; border:0; visibility:hidden;';
+    document.body.appendChild(measurementIframe);
```
nsharma123, line 72: This hidden iframe (plus the injected `<style>`) is appended to `document.body` here but nothing ever removes it — there's no `removeChild`/teardown for it in this plugin or in BookReader's lifecycle. For any page that ends up creating more than one `BookReader`/`TextSelectionPlugin` instance over its lifetime (re-mounted embeds, navigating between books without a full reload, etc.), this leaks one still-attached iframe per instance for as long as the page lives. Might be worth tearing it down on some existing lifecycle hook, or sharing a single iframe across instances, if that scenario is realistic for this codebase.

nsharma123 (commented):
Reviewed the iframe-based measurement fix. The core idea (measuring in a detached, non-scrolling iframe instead of the live document) looks solid and I confirmed it builds, lints, and passes the existing Jest suite. Left two inline comments: one is a confirmed regression (repro'd locally), the other is a resource-leak concern worth considering.

src/plugins/plugin.text_selection.js
cdrini: Great catch! Adding a guard to _configurePageContainer ; since presumably that doesn't need to create the text layer at all if it's disabled.

src/plugins/plugin.text_selection.js
cdrini: Good catch! Yeah bookreader assumes only one instance per page right now, so shouldn't realistically happen.

PR #1576, opened 2026-07-28
Temporarily restore collapsed "Copy Link to Highlight" button on desktop

cdrini:
Addendum to #1564

We will eventually need to remove this animation once we have more options in the context menu, but for now, leave-in since it is a nice experience for the current feature set.


cdrini:
Auto-merging, small addendum to unblock next BR release.

PR #1589, opened 2026-08-12
Prevent "scrollIntoView" from causing unexpected body scrolling

cdrini:
The browser native scrollIntoView method recursively moves up the tree an adjusts scrolling of all scroll containers. This creates an unexpected experience in bookreader, which should not cause e.g. the body to scroll. This PR re-implements that method so that we can control it and not cause scrolling past bookreader.


cdrini:
Auto-merging ; small refactor. Confirmed working correctly for ReadAloud, SearchInside, and CopyLinkToHighlight

Issue #1587, opened 2026-08-11
Double-tap to zoom to paragraph width


cdrini:
A common feature of web browsers is that double-tapping on a made-for-desktop website zooms so that the tapped element fills the screen. We can enable a similar behaviour in BookReader using the text layer to set the zoom to the width of the tapped paragraph, making for a consistent experience that makes reading on mobile easier.

PR #1591, opened 2026-08-26
Add double-tap-and-drag zoom gesture to 1up/2up modes

cdrini:
QA: https://deploy-preview-1591--ia-bookreader.netlify.app/bookreaderdemo/demo-internetarchive?ocaid=theworksofplato01platiala#page/n9/mode/2up

---

Double-tap-and-drag to zoom is a common interaction on mobile devices, allowing an easy one-handed alternative to pinch-zooming. First notably added to Google Maps, it has since become a standard throughout Android, and is now present in Chrome, Firefox, Drive PDF viewer, and now -- BookReader!

Here is how it works:

Double-tap-and-drag to zoom helper graphic from Google Maps

([src](https://support.google.com/maps/answer/6396990?hl=en&co=GENIE.Platform%3DAndroid#:~:text=Zoom%20in%20the%20map))

Note: plain double tapping currently does nothing ; this will be a future expansion in #1587 . This is a stepping stone to that feature.

Note: On iOS this doesn't seem to be as prevalent, but I do see it in Apple Maps -- although of course the direction was reversed -- dragging up zoomed in instead of our. I made BR reverse the direction for iOS as a result, but not sure... UPDATE: removed it from iOS ; it seems like iOS has another action for double-tap-drag ; text selection! So just disabled on iOS for now. UPDATE: Turned it on for iOS so folks can try it and we can discuss.

### DIFF
--- a/src/BookReader/Mode2UpLit.js
+++ b/src/BookReader/Mode2UpLit.js
@@ -678,11 +678,18 @@ export class Mode2UpLit extends LitElement {
 
     const $page = $(ev.target).closest('.BRpagecontainer');
     if (!$page.length) return;
-    if ($page.data('side') == 'L') {
-      this.br.left();
-    } else if ($page.data('side') == 'R') {
-      this.br.right();
-    }
+    const side = $page.data('side');
+    if (side != 'L' && side != 'R') return;
+
+    // If this tap turns out to be the first of a double-tap(-drag) zoom,
+    // don't flip the page out from under it.
+    this.smoothZoomer.runAfterTapResolved(() => {
+      if (side == 'L') {
+        this.br.left();
+      } else {
+        this.br.right();
+      }
+    });
   }
 }
 
--- a/src/BookReader/Mode2UpLit.js
+++ b/src/BookReader/Mode2UpLit.js
@@ -668,7 +668,7 @@ export class Mode2UpLit extends LitElement {
   /**
    * @param {MouseEvent} ev
    */
-  handlePageClick = (ev) => {
+  handlePageClick = async (ev) => {
     // right click
     if (ev.which == 3 && this.br.protected) {
       return false;
@@ -683,13 +683,13 @@ export class Mode2UpLit extends LitElement {
 
     // If this tap turns out to be the first of a double-tap(-drag) zoom,
     // don't flip the page out from under it.
-    this.smoothZoomer.runAfterTapResolved(() => {
+    if (await this.smoothZoomer.isSingleTap()) {
       if (side == 'L') {
         this.br.left();
       } else {
         this.br.right();
       }
-    });
+    }
   }
 }
 
--- a/src/BookReader/ModeSmoothZoom.js
+++ b/src/BookReader/ModeSmoothZoom.js
@@ -315,11 +315,6 @@ export class ModeSmoothZoom {
     }
     if (!this.activeTouch.isDoubleTapCandidate) return;
 
-    // iOS already uses double-tap-and-drag natively, for text selection --
-    // don't fight it for the zoom gesture. (The disambiguation above, e.g.
-    // not flipping the page on the first tap, still applies either way.)
-    if (isIOS()) return;
-
     // Block this from the first pixel of movement: touchAction should
     // already have stopped the browser from panning natively, but this is a
     // defensive backstop against browsers that don't fully honor that.
@@ -335,11 +330,16 @@ export class ModeSmoothZoom {
     }
 
     // Matches Google Maps/Chrome: drag down to zoom in, drag up to zoom out.
+    // iOS's own double-tap-drag zoom (e.g. Maps, Safari) runs the opposite
+    // way -- drag up to zoom in -- so flip it there to match platform norms.
+    // (Experimental: this competes with iOS's native double-tap-drag text
+    // selection gesture -- trying it out to see if that's worth the tradeoff.)
+    const dyForZoom = isIOS() ? -dy : dy;
     // Reuses pinch-zoom's own scale pipeline (_pinchMove / _drawPinchZoomFrame)
     // by modeling the drag as one finger moving away from a stationary one
     // starting DOUBLE_TAP_DRAG_REFERENCE_PX apart, and feeding the resulting
     // distance ratio in as if it were a pinch gesture's scale.
-    const virtualPinchDistance = Math.max(1, DOUBLE_TAP_DRAG_REFERENCE_PX + dy);
+    const virtualPinchDistance = Math.max(1, DOUBLE_TAP_DRAG_REFERENCE_PX + dyForZoom);
     this._pinchMove({
       scale: virtualPinchDistance / DOUBLE_TAP_DRAG_REFERENCE_PX,
       clientX: this.activeTouch.startX,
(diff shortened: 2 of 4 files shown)


src/BookReader/ModeSmoothZoom.js
```diff
@@ -211,6 +259,188 @@ export class ModeSmoothZoom {
...
+  }
+
+  /**
+   * @param {{ clientX: number, clientY: number, timeStamp: number }} e
+   */
+  _handleTapUp = async (e) => {
+    if (!this.activeTouch) return;
+
+    if (this.activeTouch.dragEngaged) {
+      await this._pinchEnd();
```
jbuckner, line 357: I think this can hang. `_drawPinchZoomFrame` only resolves `pinchMoveFramePromise` when `curScale != newScale`, so any frame that doesn't change the scale leaves it pending forever, and `_pinchEnd` awaits it.

Pinch rarely hits that, but this gesture hits it easily since the scale only depends on `dy`. Any horizontal-only move at the end of the drag does it:

```
down (50,50) t=0 ; up t=10 ; down (50,50) t=100
move (50,90) t=120   // engage, zoom in
move (80,90) t=140   // sideways only, same dy, so same scale
up   (80,90) t=160   // never completes
```

When I ran that, `_handleTapUp` never finished, so `touch-action` stayed `none`, `BRsmooth-zooming` and `willChange: transform` stayed on, and `attachCtrlZoom()` / `attachScrollListeners()` never ran. In 1up that drops the `scroll` -> `updateVisibleRegion` listener, so scrolling stops loading pages until another zoom gesture completes cleanly.

Think it's just a matter of resolving `pinchMoveFramePromiseRes` in the `curScale == newScale` case too.

src/BookReader/ModeSmoothZoom.js
```diff
@@ -211,6 +259,188 @@ export class ModeSmoothZoom {
...
+      // The double-tap window is measured from this tap's *down* (matching
+      // Chrome/Android), not from now -- so only wait out what's left of it.
+      const downTime = this.activeTouch.time;
+      this.pendingTap = { time: downTime, x: e.clientX, y: e.clientY };
+      const remainingWindowMs = Math.max(0, DOUBLE_TAP_TIME_MS - (e.timeStamp - downTime));
+
+      // Preemptively block native panning for a possible second tap, since
+      // touchAction is read by the browser at the *start* of that touch --
+      // reacting to it once that touch is already moving is too late.
+      this.mode.$container.style.touchAction = "none";
```
jbuckner, line 379: This locks the whole container for ~300ms after every tap, and I think it strands touches that turn out not to be double-taps.

If the next touch is more than 100px away it's not a candidate, so `_handleTapDown` calls `_restoreTouchAction()`, but as your comment here says, that's too late for a touch that already started. And `_handleTapMove` bails at `if (!this.activeTouch.isDoubleTapCandidate) return`, so nothing takes over either. So: tap, then flick-scroll somewhere else within 300ms, and nothing happens at all. Most noticeable in 1up where scrolling is the main thing you do.

Chrome's own version doesn't have this problem since it resolves the ambiguity natively. The Samsung Internet path below already handles this exact shape though (permanent `touch-action: none` plus `interact.draggable` -> `_dragMove`). Could we route non-candidate touches through `_dragMove` while the lock is on?

jbuckner (commented):
Really like this, and the state machine is well thought through. Notes inline.

The `_pinchEnd` one on `ModeSmoothZoom.js:357` looks like a real blocker, the `touch-action` lock on :379 I think strands scrolls, and the multi-touch guards on :268/:308 are dead on every browser we ship to. The rest are nits plus one question about plain double-tap.

Lint is clean and all 42 tests pass.

src/BookReader/ModeSmoothZoom.js
```diff
@@ -211,6 +259,188 @@ export class ModeSmoothZoom {
     this.mode.$container.scrollLeft -= e.dx;
   }
 
+  /**
+   * Tracks the start of every touch, to detect when a touch is the second
+   * tap of a double-tap (the start of a possible double-tap-and-drag zoom).
+   * @param {{ pointerType: string, clientX: number, clientY: number, timeStamp: number, originalEvent: Event }} e
+   */
+  _handleTapDown = (e) => {
+    const multiTouch = (e.originalEvent?.touches?.length ?? 1) > 1;
```
jbuckner, line 268: I don't think this ever fires. interact.js switches to Pointer Events whenever `window.PointerEvent` exists (`browser.supportsPointerEvent`, interact.js:403), which is Chrome/Edge/Firefox and Safari 13+. So `originalEvent` is a `PointerEvent` and has no `.touches` at all, and the `?? 1` always wins. Same for the copy in `_handleTapMove` on line 308.

I checked with a PointerEvent-shaped fake event: a "second finger" during an engaged drag doesn't cancel, `activeTouch` and `pinching` both stay set. The `'a second finger touching down cancels an in-progress double-tap-drag'` test only passes because it builds `touches: [{}, {}]`, which won't happen in a browser.

The down path gets rescued by the `this.pinching` check so it's not as bad as it looks, but the move path is unguarded. `TouchesMonitor` further down reads raw `touchstart`/`touchend`, which fire everywhere, so attaching it on all touch platforms instead of just iOS would make these guards real. (FWIW `e.interaction.pointers` won't work, the proxy only exposes `interactable`, `element`, `prepared`, `pointerIsDown`, `pointerWasMoved`.)

src/BookReader/Mode2UpLit.js
```diff
@@ -678,10 +678,17 @@ export class Mode2UpLit extends LitElement {
...
-    if ($page.data('side') == 'L') {
-      this.br.left();
-    } else if ($page.data('side') == 'R') {
-      this.br.right();
+    const side = $page.data('side');
+    if (side != 'L' && side != 'R') return;
+
+    // If this tap turns out to be the first of a double-tap(-drag) zoom,
+    // don't flip the page out from under it.
+    if (await this.smoothZoomer.isSingleTap()) {
```
jbuckner, line 686: Q: should the second tap be suppressed too? Tap 1's `isSingleTap()` resolves false so its flip is skipped, but tap 2's resolves true and flips. So a plain double-tap on a page still flips it once.

The description says plain double-tapping does nothing right now, and the comment in `_handleTapUp` says not to let the second tap seed a third, so I wasn't sure if that's intended. Probably matters for #1587.

src/BookReader/ModeSmoothZoom.js
```diff
@@ -82,13 +118,16 @@ export class ModeSmoothZoom {
...
+    this.baseTouchAction = this.mode.$container.style.touchAction;
 
     this.attached = true;
   }
 
   detach() {
     this.detachCtrlZoom();
+    clearTimeout(this.doubleTapWindowTimer);
+    this.doubleTapWindowTimer = null;
+    this._resolveSingleTap(false);
```
jbuckner, line 130: Small one: `detach()` doesn't restore `touch-action` or clear `pendingTap` / `activeTouch`. Tap once, then switch modes within 300ms, and the old container keeps `touch-action: none` until it gets re-attached.

A `_restoreTouchAction()` right after this line should cover it. The `_resolveSingleTap(false)` above already nulls the resolver, so it won't accidentally flip a page.

src/BookReader/ModeSmoothZoom.js
```diff
@@ -211,6 +259,188 @@ export class ModeSmoothZoom {
...
+      this._restoreTouchAction();
+    }
+
+    this.activeTouch = {
+      time: e.timeStamp,
+      startX: e.clientX,
+      startY: e.clientY,
+      isDoubleTapCandidate,
+      dragEngaged: false,
+      startScale: this.mode.scale,
```
jbuckner, line 298: `startScale` doesn't look like it's read anywhere. Leftover?

src/BookReader/ModeSmoothZoom.js
```diff
@@ -4,6 +4,24 @@ import { isIOS, isSamsungInternet } from '../util/browserSniffing.js';
 import { sleep } from './utils.js';
 /** @typedef {import('./utils/HTMLDimensionsCacher.js').HTMLDimensionsCacher} HTMLDimensionsCacher */
 
+/**
+ * Max ms between the end of the first tap and the start of the second tap to
+ * count as a double-tap. Matches Android's ViewConfiguration.DOUBLE_TAP_TIMEOUT
+ * and Flutter's kDoubleTapTimeout, both 300ms; iOS doesn't publish a number.
+ */
+const DOUBLE_TAP_TIME_MS = 300;
```
jbuckner, line 12: Nit: the doc says end-of-first-tap to start-of-second, but the code measures down to down (line 370 and the "measured from the first tap's down" test both say so).

